PR #3, opened 2015-04-16
Add structured logging support.

mr-salty:
Optional tags (timeNanos and severity) are now consumed when processed.
If we are left with a single tag called 'message' left, we continue to use
textPayload, otherwise we send the record as structPayload.


mr-salty:
Thanks - I did not notice that but since it is an existing issue I'd propose addressing it separately (see #4 which I just filed)

PR #5, opened 2015-04-21
Improve severity level support.

mr-salty:
- Ensure we only send valid severity enum values (or 'DEFAULT' if unknown).
- Translate "known" (mostly log4j) values into valid values.
- Translate single-character severities into valid values.
- Constrain numeric severities to valid values (multiples of 100 from 0-800).


filipjs:
Good idea with stripping whitespaces :)
LGTM

PR #6, opened 2015-04-30
Support setting custom GCE metadata from config

hakobera:
This pull request is to support setting custom GCE metadata from config.

Why do I need this feature is I want to use fluent-plugin-google-cloud from any servers not only GCE instances.

Google Cloud Logging is very awesome feature, but current fluent-plugin-google-cloud can only work on GCE instance, because plugin get instance metadata (project_id, zone, vm_id) using local API.

I'm very happy that you merge this pull request if you can.

- [x] Add GCE metadata config param `project_id`, `zone` and `vm_id`
- [x] Add dev dependency `"test-unit", "~> 3.0.2"` to fix fail test with `"fluentd" >= "0.12"`
- [x] Add test


mr-salty:
LGTM - thanks!

PR #7, opened 2015-05-04
Add support for Application Default Credentials.

mr-salty:
* Support Application Default Credentials and JSON keys.
* Application Default Credentials is now the default.
  Note that this still falls back to using the service account if no other
  credentials are found so there should be no change for existing users.
* auth_method is now deprecated; we emit a warning if it is configured.
  private_key is still supported for existing users.
* add a config option application_default_credentials_path, since setting
  environment variables for fluentd is inconvenient.
* The default path is $HOME/.config/gcloud/application_default_credentials.json
  ($HOME likely resolves to /root on GCE)
* Ensure the tests are actually exercising all auth methods (and using
  Application Default Credentials for most tests).
* Some misc cleanup in tests.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -21,20 +21,28 @@ module Fluent
     COMPUTE_SERVICE = 'compute.googleapis.com'
     DATAFLOW_SERVICE = 'dataflow.googleapis.com'
 
-    # Legal values:
-    # 'compute_engine_service_account' - Use the service account automatically
-    #   available on Google Compute Engine VMs. Note that this requires that
-    #   the logs.writeonly API scope is enabled on the VM, and scopes can
-    #   only be enabled at the time that a VM is created.
-    # 'private_key' - Use the service account credentials (email, private key
-    #   local file path, and file passphrase) provided below.
-    config_param :auth_method, :string,
-        :default => 'compute_engine_service_account'
+    # Name of the the Google cloud logging write scope.
+    LOGGING_SCOPE = 'https://www.googleapis.com/auth/logging.write'
 
-    # Parameters necessary to use the private_key auth_method.
+    # DEPRECATED: auth_method (and support for 'private_key') is deprecated in
+    # favor of Google Application Default Credentials as documented at:
+    # https://developers.google.com/identity/protocols/application-default-credentials
+    # 'private_key' is still accepted to support existing users; any other
+    # value is ignored.
+    config_param :auth_method, :string, :default => nil
+
+    # If set, this specifies the location of the credentials file, overriding
+    # the value of the GOOGLE_APPLICATION_CREDENTIALS environment variable.
+    config_param :application_default_credentials_path, :default => nil
+
+    # DEPRECATED: Parameters necessary to use the private_key auth_method.
     config_param :private_key_email, :string, :default => nil
     config_param :private_key_path, :string, :default => nil
     config_param :private_key_passphrase, :string, :default => 'notasecret'
+
+    # If fetch_gce_metadata is set to true, we obtain the project_id, zone,
+    # and vm_id from the GCE metadata service.  Otherwise, those parameters
+    # must be specified in the config file explicitly.
     config_param :fetch_gce_metadata, :bool, :default => true
     config_param :project_id, :string, :default => nil
     config_param :zone, :string, :default => nil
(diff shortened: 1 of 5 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -339,12 +345,15 @@ def init_api_client
...
+          @private_key_email, LOGGING_SCOPE, key)
         @client.authorization = jwt_asserter.to_authorization
         @client.authorization.expiry = 3600  # 3600s is the max allowed value
       else
-        @client.authorization = Google::APIClient::ComputeServiceAccount.new
+        if !@application_credentials_path.nil?
+          ENV['GOOGLE_APPLICATION_CREDENTIALS'] = @application_credentials_path
+        end
+        @client.authorization = Google::Auth.get_application_default(
+            LOGGING_SCOPE)
```
filipjs: I assume Google::Auth looks in ENV[x] for the credentials?

mr-salty: Correct - see https://developers.google.com/identity/protocols/application-default-credentials
Unfortunately this is the only mechanism the googleauth API allows for specifying a different file to use.

FWIW, I'm a bit on the fence about whether we should have this configuration option at all.  The user could also edit /etc/init.d/google-fluentd and add 'export GOOGLE_APPLICATION_CREDENTIALS=<xxx>' instead.  That's really no more or less difficult to do (or document) than putting it in the config file, and since it's the "canonical" way to do it perhaps we should do that rather than giving the user two ways to do the same thing?

mr-salty: I think I will do this (get rid of the option), actually.  Better to keep it simple and add it if we need it, since once it is in the wild we'd have to deprecate it to remove it.

mr-salty: done (removed).

test/plugin/test_out_google_cloud.rb
```diff
@@ -199,6 +186,38 @@ def test_gce_metadata_does_not_load_when_fetch_gce_metadata_is_false
...
+    setup_logging_stubs
+    ENV['GOOGLE_APPLICATION_CREDENTIALS'] = 'test/plugin/data/invalid_credentials.json'
+    d = create_driver()
+    d.emit({'message' => log_entry(0)})
+    exception_count = 0
+    begin
+      d.run
+    rescue RuntimeError => error
+      assert error.message.include? 'Unable to read the credential file'
+      exception_count += 1
```
filipjs: Is this counter needed?

test/plugin/test_out_google_cloud.rb
```diff
@@ -463,6 +482,35 @@ def stub_metadata_request(metadata_path, response_body)
                 :headers => {'Content-Length' => response_body.length})
   end
 
+  def setup_auth_stubs
+    # Used by 'googleauth' to test whether we're running on GCE.
+    # It only cares about the request succeeding with Metdata-Flavor: Google.
```
filipjs: Metdata -> Metadata

test/plugin/test_out_google_cloud.rb
```diff
@@ -463,6 +482,35 @@ def stub_metadata_request(metadata_path, response_body)
...
+    stub_request(:get, 'http://169.254.169.254').
```
filipjs: Where is this IP address coming from?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -339,12 +341,12 @@ def init_api_client
...
                                                  @private_key_passphrase)
         jwt_asserter = Google::APIClient::JWTAsserter.new(
-          @private_key_email, 'https://www.googleapis.com/auth/logging.write',
-          key)
+          @private_key_email, LOGGING_SCOPE, key)
         @client.authorization = jwt_asserter.to_authorization
         @client.authorization.expiry = 3600  # 3600s is the max allowed value
       else
-        @client.authorization = Google::APIClient::ComputeServiceAccount.new
+        @client.authorization = Google::Auth.get_application_default(
```
filipjs: What will happen if the user doesn't have default credentials setup?

test/plugin/test_out_google_cloud.rb
mr-salty: done.

test/plugin/test_out_google_cloud.rb
mr-salty: It is a defined well-known address for the metadata server: https://cloud.google.com/compute/docs/metadata

It's hardcoded that way in https://github.com/google/google-auth-library-ruby/blob/b5ff47fecb009f45262ee7ab87d294e4ccb1e7c1/lib/googleauth/compute_engine.rb#L56 whereas other libraries hardcode http://metadata/, but stub_request is only looking for an exact match.

Perhaps I should add a method so we register all these stubs on both places (actually, per the docs we should include http://metadata.google.internal/ as well), but let's do that in another CL.

test/plugin/test_out_google_cloud.rb
mr-salty: I think one or the other is sufficient, correct?  The one test that does 'assert false' is using it for brevity, but I think I'll change it for consistency.  All of those 'exception_count = 0' statements are useless, though, so I'll rearrange them too.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: test_one_log_with_invalid_json_credentials causes this to throw an exception (Runtime Exception "Unable to read the credential file ...").

I might be able to add one more test for the case where we're not on GCE and didn't specify any credentials, but I need the other PR to enable me to do that without major changes... I can add it after I commit both of these.

test/plugin/test_out_google_cloud.rb
filipjs: Would it be useful to add that comment in the code? (meaning that this IP == http://metadata/)

filipjs:
LGTM

PR #8, opened 2015-05-05
Fix a metadata exception on non-GCE platforms

mr-salty:
- Fix a bug where we tried to fetch GCE metadata in one case even when
  fetch_gce_metadata was false.
- Factor the GCE fake stub setup out of the tests so we can control when they
  are installed
- Add a test that sends a log with custom metadata to make sure we're
  exercising this code path (I verfied the test fails without the change
  to the plugin so it would have caught this bug).

### DIFF
--- a/test/plugin/test_out_google_cloud.rb
+++ b/test/plugin/test_out_google_cloud.rb
@@ -178,7 +178,6 @@ class GoogleCloudOutputTest < Test::Unit::TestCase
   end
 
   def test_gce_metadata_does_not_load_when_fetch_gce_metadata_is_false
-    setup_gce_metadata_stubs
     d = create_driver(CUSTOM_METADATA_CONFIG)
     d.run
     assert_equal CUSTOM_PROJECT_ID, d.instance.project_id
--- a/test/plugin/test_out_google_cloud.rb
+++ b/test/plugin/test_out_google_cloud.rb
@@ -14,6 +14,7 @@
 
 require 'helper'
 require 'json'
+require 'mocha/test_unit'
 require 'webmock/test_unit'
 
 class GoogleCloudOutputTest < Test::Unit::TestCase
@@ -185,6 +186,7 @@ class GoogleCloudOutputTest < Test::Unit::TestCase
   end
 
   def test_gce_metadata_does_not_load_when_fetch_gce_metadata_is_false
+    Fluent::GoogleCloudOutput.any_instance.expects(:fetch_metadata).never
     d = create_driver(CUSTOM_METADATA_CONFIG)
     d.run
     assert_equal CUSTOM_PROJECT_ID, d.instance.project_id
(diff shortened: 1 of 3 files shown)


mr-salty:
@hakobera - Please take a look, comments welcome.

test/plugin/test_out_google_cloud.rb
```diff
@@ -176,6 +178,7 @@ def test_managed_vm_metadata_loading
   end
 
   def test_gce_metadata_does_not_load_when_fetch_gce_metadata_is_false
```
filipjs: This is the same as "test_custom_metadata", because setting the stubs shouldn't matter if they are never used?
Which would also mean, that you can put the stubs back into the main Setup() body?

test/plugin/test_out_google_cloud.rb
```diff
@@ -224,7 +231,17 @@ def test_one_log_private_key
     verify_log_entries(1, COMPUTE_PARAMS)
   end
 
+  def test_one_log_custom_metadata
```
filipjs: If my comment from above is valid, then it applies also here.

test/plugin/test_out_google_cloud.rb
mr-salty: My intent for introducing setup_gce_metadata_stubs was to allow us to act as a GCE environment (in which case we respond to the metadata server calls) or not (in which case the test framework will raise an exception).  I don't care if all the URLs are actually requested in the former case (we could check, but I think it would make the test more brittle than necessary), but we definitely care that they are *not* requested if we're not running on GCE which is how this test missed the bug.

But,  this test does have a bug; we shoudl not call setup_gce_metadata_stubs on the following line (181) since this test should succeed in a non-GCE environment.  That would have also caught the problem (I just verified by removing line 181 and also reverting the change to the plugin) independent of the new tests I added.

hakobera:
@filipjs Thanks for fix the issue. You are right, we have to skip call metadata API to get `instance/attributes/`.

@mr-salty Sorry, it's my mistake when I refactor the code.

test/plugin/test_out_google_cloud.rb
mr-salty: ack - although this test is norsly a superset of the earlier test, I think they're still both valuable.

mr-salty:
@hakobera thanks for taking a look, and no problem - the way the tests were structured before completely obscured the issue since it mocked the GCE metadata calls even when we weren't supposed to be running on GCE.

hakobera:
@mr-salty Thanks. I think it's better to add test "not to call fetch_metadata" when `fetch_gce_metadata` is false. [mocha](https://github.com/freerange/mocha) gem allow us to do like this.

```ruby
  def test_gce_metadata_does_not_load_when_fetch_gce_metadata_is_false
    # Add this line to check not to call fetch_metadata
    Fluent::GoogleCloudOutput.any_instance.expects(:fetch_metadata).never
    
    d = create_driver(CUSTOM_METADATA_CONFIG)
    d.run
    assert_equal CUSTOM_PROJECT_ID, d.instance.project_id
    assert_equal CUSTOM_ZONE, d.instance.zone
    assert_equal CUSTOM_VM_ID, d.instance.vm_id
  end
```

mr-salty:
Done, thanks.

test/plugin/test_out_google_cloud.rb
filipjs: Oh OK, so we doing the opposite and are using those stubs to check assert_not_requested. 
Makes sense now :)

filipjs:
LGTM

PR #12, opened 2015-05-10
Improve nanosecond timestamp support.

mr-salty:
- timeNanos is deprecated because the 53 bit mantissa used for doubles
  does not have sufficient precision to represent the current time in
  nanoseconds past the unix epoch (which currently requires 61 bits).
- instead, we now support formats with split seconds and nanoseconds:
  - timestampSeconds, timestampNanos
  - timestamp { seconds, nanos }
  
  In both cases, the "seconds" part represents seconds since the unix epoch,
  and the "nano" part is the nanosecond component only (0..999999999).

  Managed VMs is using the latter, but it can only be ingested via
  json/msgpack/etc, while the former is suitable for use in an in_tail regex.

This should be considered an interim solution until
https://github.com/fluent/fluentd/issues/461 is resolved.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -167,7 +167,21 @@ module Fluent
           'entries' => [],
         }
         arr.each do |time, record|
-          if (record.has_key?('timeNanos'))
+          if (record.has_key?('timestamp') &&
+              record['timestamp'].has_key?('seconds') &&
+              record['timestamp'].has_key?('nanos'))
+            ts_secs = record['timestamp']['seconds']
+            ts_nanos = record['timestamp']['nanos']
+            record.delete('timestamp')
+          elsif (record.has_key?('timestampSeconds') &&
+                 record.has_key?('timestampNanos'))
+            ts_secs = record['timestampSeconds']
+            ts_nanos = record['timestampNanos']
+            record.delete('timestampSeconds')
+            record.delete('timestampNanos')
+          elsif (record.has_key?('timeNanos'))
+            # This is deprecated since the precision is insufficient.
+            # Use timestampSeconds/timestampNanos instead
             ts_secs = (record['timeNanos'] / 1000000000).to_i
             ts_nanos = record['timeNanos'] % 1000000000
             record.delete('timeNanos')
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -100,6 +100,7 @@ module Fluent
       init_api_client()
 
       @successful_call = false
+      @timenanos_warning = false
 
       if @use_metadata_service
         # Grab metadata about the Google Compute Engine instance that we're on.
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -167,7 +167,21 @@ def write(chunk)
           'entries' => [],
         }
         arr.each do |time, record|
-          if (record.has_key?('timeNanos'))
+          if (record.has_key?('timestamp') &&
+              record['timestamp'].has_key?('seconds') &&
+              record['timestamp'].has_key?('nanos'))
```
filipjs: If a user wants to only use seconds can he make a config that will automatically set "nanos" to zero?
If not, then maybe treat unset "nanos" as 0?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -167,7 +167,21 @@ def write(chunk)
...
+            record.delete('timestamp')
+          elsif (record.has_key?('timestampSeconds') &&
+                 record.has_key?('timestampNanos'))
+            ts_secs = record['timestampSeconds']
+            ts_nanos = record['timestampNanos']
+            record.delete('timestampSeconds')
+            record.delete('timestampNanos')
+          elsif (record.has_key?('timeNanos'))
+            # This is deprecated since the precision is insufficient.
+            # Use timestampSeconds/timestampNanos instead
```
filipjs: Add a deprecated warning message here?
Or is it going to generate to much spam? (is there log every nsec?)

lib/fluent/plugin/out_google_cloud.rb
mr-salty: I debated that but decided not to - since this is an interim step I really only want people to use this if they have a need, and whole seconds can already be handled by the native fluentd "time" using time_format %s.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: Every time will be a lot of spam, but I can log a message the first time we see timeNanos.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -167,7 +167,21 @@ def write(chunk)
...
         arr.each do |time, record|
-          if (record.has_key?('timeNanos'))
+          if (record.has_key?('timestamp') &&
+              record['timestamp'].has_key?('seconds') &&
+              record['timestamp'].has_key?('nanos'))
+            ts_secs = record['timestamp']['seconds']
+            ts_nanos = record['timestamp']['nanos']
+            record.delete('timestamp')
+          elsif (record.has_key?('timestampSeconds') &&
+                 record.has_key?('timestampNanos'))
```
mr-salty: One other thing I did debate was adding support for millis/micros since the user might have a log format that uses those and we don't have an opportunity to do math. I decided not to for simplicity but I could be convinced...

filipjs: I guess that's up to you, I don't know how often that can be useful.

lib/fluent/plugin/out_google_cloud.rb
filipjs: :+1:

filipjs:
LGTM

PR #13, opened 2015-05-14
Drop log entries which are not Hashes.

mr-salty:
- fluentd requires 'record' to be a hash, but since it does not enforce that
  there are ways for non-hash records to slip through.  although this will
  be fixed upstream, it doesn't hurt to be defensive like some other plugins
  do.  without this change we will throw an exception.

  See https://github.com/fluent/fluentd/issues/591


filipjs:
LGTM

PR #14, opened 2015-05-18
Replace uses of 'metadata' with IP 169.254.169.254

mr-salty:
The documentation doesn't give any guidance about whether to use the name
or IP address, but I've had customer reports (on Docker) of failing to
resolve the name, so using the IP address and eliminating the DNS
dependency (the name is *not* in /etc/hosts) seems preferable.

There's precedent for doing this in
https://github.com/google/google-auth-library-ruby

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -288,7 +288,7 @@ module Fluent
     end
 
     def fetch_metadata(metadata_path)
-      open('http://metadata/computeMetadata/v1/' + metadata_path,
+      open('http://169.254.169.254/computeMetadata/v1/' + metadata_path,
            {'Metadata-Flavor' => 'Google'}) do |f|
         f.read
       end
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -288,7 +288,7 @@ def log_write_failure(request, error)
     end
 
     def fetch_metadata(metadata_path)
-      open('http://metadata/computeMetadata/v1/' + metadata_path,
+      open('http://169.254.169.254/computeMetadata/v1/' + metadata_path,
```
filipjs: Would it be useful to leave a link to docs here?
https://cloud.google.com/compute/docs/metadata

filipjs:
LGTM

lib/fluent/plugin/out_google_cloud.rb
mr-salty: done

PR #16, opened 2015-06-05
support overwriting tag

### LINKED ISSUE #17: Support overwriting tag with 'tag' key in record
it would be wonderful if we can overwrite the tag value with the 'tag' value in message

hoozecn:
fix #17


hoozecn:
The issue can be solved using out_rewrite_tag_filter. 

check 

https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/17#issuecomment-111776214

PR #18, opened 2015-06-08
Allow overriding metadata service attributes.

mr-salty:
- Previously, the plugin could either obtain metadata from the metadata
  service or from the config file, depending on the setting of
  use_metadata_service.
- Now, we allow the user to override metadata in the config file even
  if use_metadata_service is true (the default).  Configured data has
  precedence over data obtained from the service.
- Fetching from the metadata service now happens in config() rather
  than start().  This shouldn't have any practical impact.
- I restructured the tests a bit and added tests for various illegal
  configs and overriding data.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -313,7 +313,7 @@ module Fluent
     def log_write_failure(request, error)
       dropped = request['entries'].length
       $log.warn "Dropping #{dropped} log message(s)",
-        :error_class=>error.class.to_s, :error=>error.to_s
+        :error_class => error.class.to_s, :error => error.to_s
     end
 
     # "enum" of Platform values
@@ -338,7 +338,7 @@ module Fluent
           end
         end
       rescue Exception => e
-        $log.debug "Failed to access metadata service: ", :error=>e
+        $log.debug "Failed to access metadata service: ", :error => e
       end
 
       $log.info 'Unable to determine platform'
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -24,6 +24,9 @@ module Fluent
     # Name of the the Google cloud logging write scope.
     LOGGING_SCOPE = 'https://www.googleapis.com/auth/logging.write'
 
+    # Address of the metadata service.
+    METADATA_SERVICE_ADDR = '169.254.169.254'
+
     # DEPRECATED: auth_method (and support for 'private_key') is deprecated in
     # favor of Google Application Default Credentials as documented at:
     # https://developers.google.com/identity/protocols/application-default-credentials
@@ -92,29 +95,25 @@ module Fluent
         end
       end
 
-      if @use_metadata_service
+      @platform = detect_platform
+
+      # set attributes from metadata (unless overriden by static config)
+      case @platform
+      when Platform::GCE
         if @project_id.nil?
-          begin
-            @project_id = fetch_metadata('project/project-id')
-          rescue
-            @project_id = nil
-          end
+          @project_id = fetch_gce_metadata('project/project-id')
         end
         if @zone.nil?
-          begin
-            fully_qualified_zone = fetch_metadata('instance/zone')
-            @zone = fully_qualified_zone.rpartition('/')[2]
-          rescue
-            @zone = nil
-          end
+          # this returns "projects/<number>/zones/<zone>"; we only want
+          # the part after the final slash.
+          fully_qualified_zone = fetch_gce_metadata('instance/zone')
+          @zone = fully_qualified_zone.rpartition('/')[2]
         end
         if @vm_id.nil?
-          begin
-            @vm_id = fetch_metadata('instance/id')
-          rescue
-            vm_id = nil
-          end
+          @vm_id = fetch_gce_metadata('instance/id')
         end
+      when Platform::OTHER
+        # Do nothing
       end
 
       # all metadata parameters must now be set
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -86,37 +92,55 @@ def configure(conf)
...
+          begin
+            @project_id = fetch_metadata('project/project-id')
+          rescue
+            @project_id = nil
+          end
+        end
+        if @zone.nil?
+          begin
+            fully_qualified_zone = fetch_metadata('instance/zone')
+            @zone = fully_qualified_zone.rpartition('/')[2]
```
filipjs: Might be worth adding an example of a zone, to see what the split does.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -86,37 +92,55 @@ def configure(conf)
...
 
-      unless @use_metadata_service
-        unless @project_id && @zone && @vm_id
-          raise Fluent::ConfigError,
-              ('Please specify "project_id", "zone" and "vm_id" if you set "use_metadata_service" to false')
+      if @use_metadata_service
+        if @project_id.nil?
+          begin
+            @project_id = fetch_metadata('project/project-id')
+          rescue
```
filipjs: This code structure will hide errors while fetching metadata.
Is there a way to easy debug that if we don't propagate the errors?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -86,37 +92,55 @@ def configure(conf)
...
       # If this is running on a Managed VM, grab the relevant App Engine
       # metadata as well.
       # TODO: Add config options for these to allow for running outside GCE?
-      attributes_string = @use_metadata_service ?
-          fetch_metadata('instance/attributes/') : ""
+      attributes_string = ""
+      if (@use_metadata_service)
+        begin
+          attributes_string = fetch_metadata('instance/attributes/')
+        rescue
```
filipjs: Same comment about fetching metadata here.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -144,6 +168,15 @@ def start
       end
     end
 
+    def start
+      super
+
+      init_api_client()
```
filipjs: This call sets up authorization. Isn't that needed for "fetch_metadata" to work?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -36,10 +36,16 @@ class GoogleCloudOutput < BufferedOutput
...
-    # must be specified in the config file explicitly.
+    # Specify project/instance metadata.
+    #
+    # project_id, zone, and vm_id are required to have valid values, which
+    # can be obtained from the metadata service or set explicitly.
+    # Otherwise, the plugin will fail to initialize.
+    #
+    # Whether to attempt to obtain metadata from the local metadata service.
+    # It is safe to specify 'true' even on platforms with no metadata service.
     config_param :use_metadata_service, :bool, :default => true
```
filipjs: I think it should default to false if any of the param below is overridden in the config.
User should supply all 3 params, or explicitly set to fetch the rest from the metadata service.

mr-salty: I can expand the comment, but the aim of this CL is basically to make this flag unnecessary by auto-detecting whether we have a metadata service or not (in the next CL we will also auto-detect which platform we are running on).  If there is no metadata service then we will try to fetch from it and fail as you noted below.

We explicitly do not want it to be "all or nothing" because on EC2 we cannot obtain project_id from the metadata service.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: This is intentional; the error will be reported below (Unable to obtain metadata parameters) since project_id, etc. becomes nil.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: ok, will do.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: this is the same basic idea, if there is no metadata service then we just proceed as if no attributes are set.  But, I think I will re-structure things a bit so instead of having 'rescue' here I will remember whether we have a metadata service or not, and only try to fetch if so.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: no, the metadata service is just a local (to the VM) http server, it doesn't require any auth.

lib/fluent/plugin/out_google_cloud.rb
filipjs: It will be reported, but my question was if the user will be able to easily distinguish between 
"oops I forgot to set project_id" and "fetch_metadata returned HTTP error".

lib/fluent/plugin/out_google_cloud.rb
filipjs: It won't be all or nothing. The specific scenario I'm referring to is as follows:
User wants to override e.g. zone, but he does some mistake/forgets to do it.
If the flag defaults to true, it will be fetched automatically and logs will have wrong data.

The alternative solution would be something like:
if project_ud or zone or vm_id then use_metadata_service, default => false
else use_metadata_service, default => true

However I don't know how likely is that scenario, you probably have a better idea if it's worth pursuing that route :)

lib/fluent/plugin/out_google_cloud.rb
mr-salty: let me restructure it as I described, the new scheme will be:

- idetermine the platform we're running on by consulting the metadata service (currently this will be either 'GCE' or 'NONE')
- get rid of the 'rescue' clauses for other fetch_metadata calls.
- inside fetch_metadata, if platform == NONE, I'll either return nil or throw an exception (let me think about that).

lib/fluent/plugin/out_google_cloud.rb
mr-salty: I don't like the idea of changing the value of one item based on whether other flags are present, that seems confusing.

If we do that, then imagine I'm running on GCE and decide I want to override zone.  Now I can't just say 'zone XXX', I have to also add 'use_metadata_service true' or it will break.

Ideally users will never have to set 'use_metadata_service' regardless of what platform they are running on.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: done.

mr-salty:
PTAL; hopefully adding the platform detection makes this a little clearer.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -287,10 +316,41 @@ def log_write_failure(request, error)
...
+
+      begin
+        open('http://' + METADATA_SERVICE_ADDR) do |f|
+          if (f.meta['metadata-flavor'] == 'Google')
+            $log.info 'Detected GCE platform'
+            return Platform::GCE
+          end
+        end
+      rescue Exception => e
+        $log.debug "Failed to access metadata service: ", :error=>e
```
filipjs: should there be spaces around "=>" ?

test/plugin/test_out_google_cloud.rb
```diff
@@ -187,6 +242,16 @@ def test_managed_vm_metadata_loading
...
+    assert_equal CUSTOM_ZONE, d.instance.zone
+    assert_equal CUSTOM_VM_ID, d.instance.vm_id
+    assert_equal false, d.instance.running_on_managed_vm
+  end
+
+  def test_metadata_overrides_on_gce
+    setup_gce_metadata_stubs
     d = create_driver(CUSTOM_METADATA_CONFIG)
     d.run
     assert_equal CUSTOM_PROJECT_ID, d.instance.project_id
```
filipjs: Can you add a comment what is the expected output and why.

test/plugin/test_out_google_cloud.rb
```diff
@@ -195,6 +260,16 @@ def test_gce_metadata_does_not_load_when_use_metadata_service_is_false
     assert_equal false, d.instance.running_on_managed_vm
   end
 
+  def test_metadata_partial_overrides_on_gce
+    setup_gce_metadata_stubs
+    d = create_driver(CONFIG_MISSING_METADATA_PROJECT_ID)
+    d.run
+    assert_equal PROJECT_ID, d.instance.project_id
```
filipjs: Same here. Otherwise it's hard to tell what's the actual difference between those two tests.

filipjs:
LGTM
Looks great now :)

lib/fluent/plugin/out_google_cloud.rb
mr-salty: fixed (I also fixed the one other place with no spaces, which was also a log message).  I have a TODO to find a linter to use and reformat things.

test/plugin/test_out_google_cloud.rb
mr-salty: done.  I was debating whether to get rid of the CONFIG constants and inline the configuration in the create_driver call instead, which may make it more obvious.  I think the constants made more sense when there were only two of them but now that I have a bunch I don't think they're helping readability... so perhaps I will do that in a followup.

test/plugin/test_out_google_cloud.rb
mr-salty: done

PR #19, opened 2015-06-10
Add support for EC2 platform and metadata.

mr-salty:
- Detect whether we're running on ec2 using the [metadata server](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html)
- Add EC2 service name and label values:
  - service = ec2.amazonaws.com
  - zone = *availabiltyZone from metadata* (with aws: prefix)
  - ec2.amazonaws.com/resource_type = "instance"
  - ec2.amazonaws.com/resource_id = *instanceId from metadata*
  - ec2.amazonaws.com/account_id = *accountId from metadata*

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -99,13 +99,11 @@ module Fluent
 
       # TODO: Send instance tags and/or hostname as labels as well?
       @common_labels = {}
-      @running_on_managed_vm = false
 
       # set attributes from metadata (unless overriden by static config)
       @platform = detect_platform
       case @platform
       when Platform::GCE
-      # Use "generic" COMPUTE_SERVICE as the default environment.
         if @project_id.nil?
           @project_id = fetch_gce_metadata('project/project-id')
         end
@@ -145,6 +143,9 @@ module Fluent
           ('Unable to obtain metadata parameters: ' + missing.join(' '))
       end
 
+      # Default this to false; it is only overwritten if we detect Managed VM.
+      @running_on_managed_vm = false
+
       # Set labels, etc. based on the config
       case @platform
       when Platform::GCE
@@ -166,12 +167,12 @@ module Fluent
             @gae_backend_version
         elsif (attributes.include?('job_id'))
           # Dataflow
-          @running_on_managed_vm = false
           @service_name = DATAFLOW_SERVICE
           @dataflow_job_id = fetch_gce_metadata('instance/attributes/job_id')
           common_labels["#{DATAFLOW_SERVICE}/job_id"] = @dataflow_job_id
         end
-        # dataflow only uses their own labels; otherwise include GCE labels.
+        # include GCE labels unless we're running on dataflow, which
+        # uses their own labels exclusively.
         if (@service_name != DATAFLOW_SERVICE)
           common_labels["#{COMPUTE_SERVICE}/resource_type"] = 'instance'
           common_labels["#{COMPUTE_SERVICE}/resource_id"] = @vm_id
@@ -181,6 +182,7 @@ module Fluent
         common_labels["#{EC2_SERVICE}/resource_type"] = 'instance'
         common_labels["#{EC2_SERVICE}/resource_id"] = @vm_id
       when Platform::OTHER
+        # Use COMPUTE_SERVICE as the default environment.
         @service_name = COMPUTE_SERVICE
         common_labels["#{COMPUTE_SERVICE}/resource_type"] = 'instance'
         common_labels["#{COMPUTE_SERVICE}/resource_id"] = @vm_id
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -16,10 +16,11 @@ module Fluent
   class GoogleCloudOutput < BufferedOutput
     Fluent::Plugin.register_output('google_cloud', self)
 
-    # Constants for Google service names.
+    # Constants for service names.
     APPENGINE_SERVICE = 'appengine.googleapis.com'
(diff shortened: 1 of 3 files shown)


mr-salty:
Hi, I still need to add the tests but I wanted to send the code changes to you for review.  I am using this plugin on my EC2 instance now.

mr-salty:
tests added - PTAL, thanks!

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -95,11 +97,15 @@ def configure(conf)
...
-      @platform = detect_platform
+      # TODO: Send instance tags and/or hostname as labels as well?
+      @common_labels = {}
+      @running_on_managed_vm = false
 
       # set attributes from metadata (unless overriden by static config)
+      @platform = detect_platform
       case @platform
       when Platform::GCE
+      # Use "generic" COMPUTE_SERVICE as the default environment.
```
filipjs: This comment is out of place

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -95,11 +97,15 @@ def configure(conf)
         end
       end
 
-      @platform = detect_platform
+      # TODO: Send instance tags and/or hostname as labels as well?
+      @common_labels = {}
+      @running_on_managed_vm = false
```
filipjs: I would leave this before the second "case" so it's closer where it's actually set.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -156,9 +171,17 @@ def configure(conf)
...
+        if (@service_name != DATAFLOW_SERVICE)
+          common_labels["#{COMPUTE_SERVICE}/resource_type"] = 'instance'
+          common_labels["#{COMPUTE_SERVICE}/resource_id"] = @vm_id
+        end
+      when Platform::EC2
+        @service_name = EC2_SERVICE
+        common_labels["#{EC2_SERVICE}/resource_type"] = 'instance'
+        common_labels["#{EC2_SERVICE}/resource_id"] = @vm_id
+      when Platform::OTHER
+        @service_name = COMPUTE_SERVICE
```
filipjs: Here is there the command about "use generic" should be :)

test/plugin/test_out_google_cloud.rb
```diff
@@ -274,6 +305,29 @@ def test_metadata_partial_overrides_on_gce
     assert_equal false, d.instance.running_on_managed_vm
   end
 
+  def test_ec2_metadata_loading
+    setup_ec2_metadata_stubs
+    d = create_driver(CONFIG_EC2_PROJECT_ID)
```
filipjs: Maybe add one more test with CUSTOM_EC2_CONFIG that overrides zone or vm_id?

test/plugin/test_out_google_cloud.rb
```diff
@@ -127,6 +146,18 @@ def setup
...
 
+  EC2_PARAMS = {
+    'service_name' => EC2_SERVICE_NAME,
+    'log_name' => 'test',
+    'project_id' => EC2_PROJECT_ID,
+    'zone' => EC2_PREFIXED_ZONE,
+    'labels' => {
+      "#{EC2_SERVICE_NAME}/resource_type" => 'instance',
+      "#{EC2_SERVICE_NAME}/resource_id" => EC2_VM_ID,
+      "#{EC2_SERVICE_NAME}/account_id" => EC2_ACCOUNT_ID
```
filipjs: Is it important to test that this will work when "account_id" in not in metadata returned by EC2?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -156,9 +171,17 @@ def configure(conf)
           @dataflow_job_id = fetch_gce_metadata('instance/attributes/job_id')
           common_labels["#{DATAFLOW_SERVICE}/job_id"] = @dataflow_job_id
         end
-      end
-
-      if (@service_name != DATAFLOW_SERVICE)
+        # dataflow only uses their own labels; otherwise include GCE labels.
```
filipjs: remove "only" from comment

filipjs:
LGTM

lib/fluent/plugin/out_google_cloud.rb
mr-salty: done

lib/fluent/plugin/out_google_cloud.rb
mr-salty: done

lib/fluent/plugin/out_google_cloud.rb
mr-salty: done

test/plugin/test_out_google_cloud.rb
mr-salty: done

lib/fluent/plugin/out_google_cloud.rb
mr-salty: I meant "exclusively" and rewrote the comment to make this more clear (in managed VMs case we send both compute and appengine labels, but dataflow sends only dataflow labels).

test/plugin/test_out_google_cloud.rb
mr-salty: That's an interesting question, I don't think it's ever expected that we won't have account_id set but it's possible there is some past or future version of the metadata service that does not.

I was debating whether the plugin should throw an exception in that case, but this label is not really required, so I think for now I'll leave it as-is.  I don't think there's much value in testing the "missing" case but I can add one if you feel strongly about it.

filipjs: No need, it's fine without it.

PR #20, opened 2015-06-25
Ensure timestamp is a Hash before using it as one.

mr-salty:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -222,8 +222,9 @@ def write(chunk)
           'entries' => [],
         }
         arr.each do |time, record|
```
filipjs: Where is "time" taken from if there is nothing that can used in the record?
Also I think the original report was that "timestamp" was a "seconds since epoch", and is that reasonable that it should also work? Or in that case we expect the user to use a combo of "timestampSeconds" and "timestampNanos" in the record?

filipjs:
LGTM

broniek:
thanks :)

lib/fluent/plugin/out_google_cloud.rb
mr-salty: This has definitely become a bit of a mess, with too many ways to accomplish the same thing, and as Przemek points out it also pollutes the namespace unexpectedly. And, to cap it off, it's not really documented ;)  The real solution is to address this natively in fluentd https://github.com/fluent/fluentd/issues/461 and then all of this becomes deprecated.  But, there hasn't been much progress there so I probably need to push on that again.

But, "time" is the officially supported (and documented) fluent key, and it represents seconds since epoch.    If it is not present the current time (at ingestion) is used.  So, I don't think there is a need to allow timestamp to represent seconds as well; I don't want to encourage people to use it unless they have a specific need for nanoseconds.  With this change, if they do use it inadvertently it will just pass through as a struct field.

PR #21, opened 2015-07-07
Add support for configuring user-defined labels.

mr-salty:
- Add a label_map configuration option

     label_map (specified as a JSON object) is an unordered set of fluent 
     field names whose values are sent as labels rather than as part of the
     struct payload.
    
     Each entry in the map is a {"field_name": "label_name"} pair.  When
     the "field_name" (as parsed by the input plugin) is encountered, a label
     with the corresponding "label_name" is added to the log entry.  The
     value of the field is used as the value of the label.
    
     The map gives the user additional flexibility in specifying label
     names, including the ability to use characters which would not be
     legal as part of fluent field names.
    
     Example:
```
       label_map {
         "field_name_1": "sent_label_name_1",
         "field_name_2": "some.prefix/sent_label_name_2"
       }
```

- Ancillary change: modify the test to use the 'v1' config format since that
  is what we are using in the released google-fluentd (and it handles parsing
  multi-line hash values correctly).

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -54,6 +54,18 @@ module Fluent
     config_param :zone, :string, :default => nil
     config_param :vm_id, :string, :default => nil
 
+    # List of parsed fields whose values are sent as labels instead of
+    # becoming part of the struct payload.
+    #
+    # This is specified as a JSON array where the 'name' is the field name
+    # parsed by fluentd and the 'value' is the name of the label to send.
+    #
+    # label_map {
+    #   "parsed_field_name_1": "sent_label_name_1",
+    #   "parsed_field_name_2": "sent_label_name_2"
+    # }
+    config_param :label_map, :hash, :default => nil
+
     # TODO: Add a log_name config option rather than just using the tag?
 
     # Expose attr_readers to make testing of metadata more direct than only
@@ -269,6 +281,22 @@ module Fluent
             record.delete('severity')
           else
             entry['metadata']['severity'] = 'DEFAULT'
+          end
+
+          # If a field is present in the label_map, send its value as a label
+          # (mapping the field name to label name as specified in the config)
+          # and do not send that field as part of the payload.
+          if !label_map.nil?
+            labels = {}
+            @label_map.each do |field, label|
+              if record.has_key?(field)
+                labels[label] = record[field]
+                record.delete(field)
+              end
+            end
+            if !labels.empty?
+              entry['metadata']['labels'] = labels
+            end
           end
 
           # use textPayload if the only remainaing key is 'message',
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -54,16 +54,24 @@ module Fluent
     config_param :zone, :string, :default => nil
     config_param :vm_id, :string, :default => nil
 
-    # List of parsed fields whose values are sent as labels instead of
-    # becoming part of the struct payload.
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -271,6 +283,22 @@ def write(chunk)
             entry['metadata']['severity'] = 'DEFAULT'
           end
 
+          # If a field is present in the label_map, send its value as a label
+          # (mapping the field name to label name as specified in the config)
+          # and do not send that field as part of the payload.
+          if !label_map.nil?
```
filipjs: Since the label map is constant, wouldn't it be better to put those values into "commonLabels" and just remove payload fields here?

mr-salty: The values are parsed out of each log entry so they could be different for each one.

It's possible the label/value pair could end up being the same for all entries in the batch, I could write a loop later to detect that although I'm not sure if it's worth the effort.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -54,6 +54,18 @@ class GoogleCloudOutput < BufferedOutput
     config_param :zone, :string, :default => nil
     config_param :vm_id, :string, :default => nil
 
+    # List of parsed fields whose values are sent as labels instead of
```
filipjs: Remove "parsed" from here?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -54,6 +54,18 @@ class GoogleCloudOutput < BufferedOutput
...
+    # becoming part of the struct payload.
+    #
+    # This is specified as a JSON array where the 'name' is the field name
```
filipjs: array => map

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -54,6 +54,18 @@ class GoogleCloudOutput < BufferedOutput
...
+    # parsed by fluentd and the 'value' is the name of the label to send.
```
filipjs: Maybe change 'name' to 'map key' and 'value' to 'map value' to make it more clear?
On my first reading I thought we are just applying static values as label values, that's why I made that first comment below.

lib/fluent/plugin/out_google_cloud.rb
filipjs: I misunderstood what this commit was supposed to do.
This part is fine :)

lib/fluent/plugin/out_google_cloud.rb
mr-salty: done.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: ack.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: I think it really should say "object", which is what I changed it to.

filipjs:
LGTM

lib/fluent/plugin/out_google_cloud.rb
mr-salty: The JSON documentation refers to these as name/value pairs...  but it is confusing, I re-wrote this comment and I think it's more understandable, but we'll have to have Sam write some documentation with more examples.

mr-salty: (I updated the description of this PR as well so it will be obvious to peope looking at the change history)

PR #22, opened 2015-07-15
Enable rubocop and fix style violations.

mr-salty:
This change is a functional no-op; all changes are to conform
to the style guide as enforced by rubocop.

Note that I (mostly) didn't attempt to fix class/method length or complexity warnings; these are (auto) disabled in the TODO file.  I did split a few tests that were testing multiple things into parts.


mr-salty:
I've had this as a TODO for a while and this seems to be the accepted linter and style guide.  Other google projects (API client, auth library) are already using it as well.

I don't agree with all of it, especially https://github.com/bbatsov/ruby-style-guide#no-explicit-return which I debated disabling, but in the end I decided to go with the standard as-is.  433 violations to start!

Hopefully you can see the general themes - there a lot of changes about quoting, indentation, and string continuation.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -133,34 +142,32 @@ def configure(conf)
           fully_qualified_zone = fetch_gce_metadata('instance/zone')
           @zone = fully_qualified_zone.rpartition('/')[2]
         end
-        if @vm_id.nil?
-          @vm_id = fetch_gce_metadata('instance/id')
-        end
+        @vm_id = fetch_gce_metadata('instance/id') if @vm_id.nil?
```
filipjs: This was the suggestion from "lint" ?
Looks much less readable.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -133,34 +142,32 @@ def configure(conf)
...
       when Platform::EC2
         metadata = fetch_ec2_metadata
-        if @zone.nil? && metadata.has_key?('availabilityZone')
+        if @zone.nil? && metadata.key?('availabilityZone')
```
filipjs: I don't know how to read this: http://ruby-doc.org/core-2.2.2/Hash.html
According to the descriptions there are 4 methods to check for key presence: "has_key", "key", "include", "member" but all examples show "has_key".
Does that mean this is the preferred method? We use "has_hey" here and "include" a little down.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -230,43 +237,41 @@ def write(chunk)
       # Group the entries since we have to make one call per tag.
       grouped_entries = {}
       chunk.msgpack_each do |tag, *arr|
-        if !grouped_entries.has_key?(tag)
-          grouped_entries[tag] = []
-        end
+        grouped_entries[tag] = [] unless grouped_entries.key?(tag)
```
filipjs: If looked better

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -388,55 +390,54 @@ module Platform
...
-      rescue Exception => e
-        $log.debug "Failed to access metadata service: ", :error => e
+      rescue StandardError => e
+        @log.debug 'Failed to access metadata service: ', error: e
       end
 
-      $log.info 'Unable to determine platform'
-      return Platform::OTHER
+      @log.info 'Unable to determine platform'
+      Platform::OTHER
```
filipjs: Why is "return" not needed? This is normal for ruby?

lib/fluent/plugin/out_google_cloud.rb
mr-salty: yes, all suggestions were from rubocop, based on this style guide: https://github.com/bbatsov/ruby-style-guide

like I said, I don't really agree with them, but the other projects I looked at do not make any alterations to the style guide (aside from one of them using 120 char lines) so I decided it was best to stick to the style guide strictly even where I don't agree.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: yeah, that is a general problem with ruby, having redundant ways to do something.  The style guide says has_key is deprecated in favor of key so I made that change throughout.

I looked at the other uses of include? and they're all on other data types (array, Set, or hash)

lib/fluent/plugin/out_google_cloud.rb
mr-salty: same general comment as before.. it's certainly unfamiliar to C++ types like us but appears to be idiomatic in Ruby...

lib/fluent/plugin/out_google_cloud.rb
mr-salty: Yeah, as I said in my other note, I was going to change one thing it would be this - I don't like this syntax at all.  This is something I know is common in perl and I guess ruby adopted it as well.

mr-salty:
@filipjs - are you ok with me checking this in?  I think all of your comments are about the style guide itself, which I don't think we should change even if we don't personally agree with it.

lib/fluent/plugin/out_google_cloud.rb
filipjs: I agree, maybe it is worth to skip this rule? Or it's not something can be done easily?

filipjs:
LGTM

lib/fluent/plugin/out_google_cloud.rb
mr-salty: It can be done (you can disable individual warnings in the config file) but it seems like a bit of a slippery slope to me and I'd rather conform to the guide as-written even if I don't agree.

PR #25, opened 2015-08-04
Ensure label values are strings.

mr-salty:
Label values are required to be strings, so when using label_map,
coerce the parsed value to be a string.

I confirmed the test I added failed before changing the plugin and
passes now.

Rubocop also made me make a couple changes (including replacing the
conditional with 'next unless ...')

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -302,10 +302,10 @@ module Fluent
           unless label_map.nil?
             labels = {}
             @label_map.each do |field, label|
-              if record.key?(field)
-                labels[label] = record[field]
-                record.delete(field)
-              end
+              next unless record.key?(field)
+              # label values are required to be strings.
+              labels[label] = record[field].to_s
+              record.delete(field)
             end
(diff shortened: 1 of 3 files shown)


mr-salty:
@apolito this should fix your issue

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -302,10 +302,10 @@ def write(chunk)
           unless label_map.nil?
             labels = {}
             @label_map.each do |field, label|
-              if record.key?(field)
-                labels[label] = record[field]
-                record.delete(field)
-              end
+              next unless record.key?(field)
+              # label values are required to be strings.
+              labels[label] = record[field].to_s
```
filipjs: Does "to_s" work on every type? 
What happens when this field is a record (hash?) itself?

filipjs:
LGTM

lib/fluent/plugin/out_google_cloud.rb
mr-salty: to_s is a method on Object so it should always exist.  I added a test case for hash as well, which doesn't seem that useful to me for a user to actually do, but it does work.

PR #26, opened 2015-08-07
Include the VM hostname as an additional label when running on GCE or EC2

a-robinson:
I verified that this works as expected on GCE. I didn't explicitly check on EC2, but Socket.gethostname should work everywhere, including on Windows.

We could also get this from the metadata service, but then some extra string parsing is needed to pull off some of the extra info and it didn't look like EC2's metadata doc included hostname.

Let me know if this shouldn't be using one of the domain-scoped labels, I don't know what the standards are for them. I'd be happy to change the label key to be hostname.

@mr-salty @goltermann

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -57,6 +57,7 @@ module Fluent
     config_param :project_id, :string, :default => nil
     config_param :zone, :string, :default => nil
     config_param :vm_id, :string, :default => nil
+    config_param :vm_name, :string, :default => nil
 
     # label_map (specified as a JSON object) is an unordered set of fluent
     # field names whose values are sent as labels rather than as part of the
@@ -101,6 +102,7 @@ module Fluent
       require 'googleauth'
       require 'json'
       require 'open-uri'
+      require 'socket'
 
       # use the global logger
       @log = $log # rubocop:disable Style/GlobalVars
@@ -130,6 +132,7 @@ module Fluent
       @common_labels = {}
 
       # set attributes from metadata (unless overriden by static config)
+      @vm_name = Socket.gethostname if @vm_name.nil?
       @platform = detect_platform
       case @platform
       when Platform::GCE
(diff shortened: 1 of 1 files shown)


goltermann:
LGTM on logic.  I'm not a Ruby guy, so unsure about style.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -150,6 +153,7 @@ def configure(conf)
         end
         if @vm_id.nil? && metadata.key?('instanceId')
           @vm_id = metadata['instanceId']
+        @vm_name = Socket.gethostname if @vm_name.nil?
```
mr-salty: I think we should do this regardless of platform, so maybe just move it outside of the 'case' altogether (and eliminate the other instance above)?

mr-salty:
FWIW, don't worry about the travis failure (other than the indentation problem, although that should go away if we move that outside the case).  This rubocop (linter, etc) stuff is a relatively recent addition and I'm still trying to learn how to live with it.

These changes push the branch/cyclometric complexity over the limit in the "TODO" file, but I think that's too harsh.  I'll fix that later.

The ones about spacing, I'm not sure why it's complaining about those when it's the same in HEAD.  Maybe rubocop updates underneath me and I should be pinning it to a specific version in the dependencies?

lib/fluent/plugin/out_google_cloud.rb
a-robinson: Done.

a-robinson:
Thanks for the review, Todd!

Given that we now always have a hostname, what are your thoughts on keeping the service-specific label keys vs just always attaching a generic "hostname" label?

mr-salty:
Hm, I think the intent was to give all auto-generated labels a prefix to avoid polluting the namespace for user labels.  I agree it's a bit awkward here, but let's stick with it for now.

a-robinson:
And sorry for the dumb question, but how does the gem packaging for this work? We're not in a huge rush since there probably won't be a kubernetes patch version cut for the next couple weeks, but how will I know when this is available?

mr-salty:
I can build and push a new gem pretty easily, but I removed the "gem update" from k8s because it's kind of unsafe (due to dependency issues), so I really need to build and publish a new google-fluentd with the updated gem.  I publish a github release when I release a new deb so you can watch that project.

PR #27, opened 2015-08-17
Gemspec description update

brint:
Fixed a spelling error s/loge/logs/ and removed the trailing
single quote from the end of the string since it's unnecessary.


mr-salty:
LGTM - thanks

PR #28, opened 2015-09-01
Add a secret parameter

cosmo0920:
`private_key_passphrase` contains sensitive information. 
This parameter should be concealed with secret parameter feature.


mr-salty:
Thanks - note that this authentication method is deprecated, please migrate to JSON keys when you are able.

PR #29, opened 2015-09-04
Fix label map check that prevents labels from being populated properly

a-robinson:
@mr-salty


mr-salty:
This is good for consistency but AFAICT since config_param both defines an instance variable and uses attr_accessor to define a getter and setter, this should have no practical effect (at least for gets, sets from within the class would require self.label_map=... or else it *would* [set a local variable](http://stackoverflow.com/questions/19617169/ruby-instance-methods-why-can-i-use-getter-without-self-but-setter-only-with?rq=1) (ugh)).

So, I don't have an explanation for any misbehavior you saw, if you have example code let me know.

Issue #15, opened 2015-05-26
Support kubernetes-specific service name and common labels


a-robinson:
@satnam6502 do you have cycles for this? I can do it if you don't.

@mr-salty

mr-salty:
FYI, the cloud logging API currently only accepts a known set of service names and labels so we will need to add support for this on the backend (and wait for the next weekly release) before doing anything on the plugin.  @andreweames

mr-salty:
OK, Cloud Logging API support for this is in place, so we can modify the plugin at will..

service: "container.googleapis.com";
primary label: "container.googleapis.com/cluster";
secondary label: "container.googleapis.com/namespace";

mr-salty:
@a-robinson - fixed?

a-robinson:
Yes -- I'll send another PR soon to fall back to just populating cluster name, namespace name, pod name, and container name in the case that the metadata plugin isn't being used, but this can be considered done.

PR #30, opened 2015-09-04
Add support for better labels on Kubernetes and Container Engine logs.

a-robinson:
Assumes the usage of [fluent-plugin-kubernetes_metadata_filter](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter), as well as an upcoming change of mine to that plugin to include namespace IDs. The exact field names we parse out of the record may have to change slightly as a result of that, but everything else should be stable.

There are a few issues worth discussing/clarifying/pointing out:
1. Do we want both the container.googleapis.com labels and the compute.googleapis.com labels?
2. Do all the labels belong under container.googleapis.com even though they won't be first-class fields in the log viewer? I assume yes, but my knowledge is out of date.
3. Namespace ID is going to be pretty awkward to add to the metadata filter plugin.
4. We could fall back to parsing the namespace, pod, and container name out of the log name in the case that the metadata plugin isn't being used. I've left this as a TODO for now.
5. This adds a lot of json parsing. It'll be worth keeping an eye out for higher cpu usage, but I think it makes sense to try this setup first before doing anything more natively in this plugin.

@mr-salty @brendandburns

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -20,6 +20,7 @@ module Fluent
     # Constants for service names.
     APPENGINE_SERVICE = 'appengine.googleapis.com'
     COMPUTE_SERVICE = 'compute.googleapis.com'
+    CONTAINER_SERVICE = 'container.googleapis.com'
     DATAFLOW_SERVICE = 'dataflow.googleapis.com'
     EC2_SERVICE = 'ec2.amazonaws.com'
 
@@ -105,6 +106,7 @@ module Fluent
       require 'json'
       require 'open-uri'
       require 'socket'
+      require 'yaml'
 
       # use the global logger
       @log = $log # rubocop:disable Style/GlobalVars
@@ -202,6 +204,14 @@ module Fluent
           @service_name = DATAFLOW_SERVICE
           @dataflow_job_id = fetch_gce_metadata('instance/attributes/job_id')
           common_labels["#{DATAFLOW_SERVICE}/job_id"] = @dataflow_job_id
+        elsif attributes.include?('kube-env')
+          # Kubernetes/Container Engine
+          @service_name = CONTAINER_SERVICE
+          common_labels["#{CONTAINER_SERVICE}/instance_id"] = @vm_id
+          @raw_kube_env = fetch_gce_metadata('instance/attributes/kube-env')
+          @kube_env = YAML.load(@raw_kube_env)
+          common_labels["#{CONTAINER_SERVICE}/cluster_name"] =
+            cluster_name_from_kube_env(@kube_env)
         end
         # include GCE labels unless we're running on dataflow, which
         # uses their own labels exclusively.
@@ -304,16 +314,22 @@ module Fluent
             entry['metadata']['severity'] = 'DEFAULT'
           end
 
+          # TODO(a-robinson): Consider falling back to parsing the namespace,
+          # pod, and container names out of the log name if we don't have this
+          # information, which depends on the kubernetes_metadata_filter
+          # filter plugin.
+          if @service_name == CONTAINER_SERVICE && record['kubernetes']
+            handle_container_metadata(record, entry)
+          end
+
           # If a field is present in the label_map, send its value as a label
           # (mapping the field name to label name as specified in the config)
           # and do not send that field as part of the payload.
-          unless @label_map.nil?
+          if @label_map
             labels = {}
+            labels = entry['metadata']['labels'] if entry['metadata']['labels']
             @label_map.each do |field, label|
-              next unless record.key?(field)
-              # label values are required to be strings.
-              labels[label] = record[field].to_s
-              record.delete(field)
+              field_to_label(record, field, labels, label)
             end
             entry['metadata']['labels'] = labels unless labels.empty?
           end
@@ -443,6 +459,15 @@ module Fluent
       end
     end
 
+    def cluster_name_from_kube_env(kube_env)
+      return kube_env['CLUSTER_NAME'] if kube_env.key?('CLUSTER_NAME')
+      instance_prefix = kube_env['INSTANCE_PREFIX']
+      gke_name_match = /^gke-(.+)-[0-9a-f]{8}$/.match(instance_prefix)
+      return gke_name_match.captures[0] if gke_name_match &&
+                                           !gke_name_match.captures.empty?
+      instance_prefix
+    end
+
     # Values permitted by the API for 'severity' (which is an enum).
     VALID_SEVERITIES = Set.new(
       %w(DEFAULT DEBUG INFO NOTICE WARNING ERROR CRITICAL ALERT EMERGENCY))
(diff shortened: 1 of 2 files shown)


a-robinson:
I'm working on tests in parallel, which I'll add to this PR soon, just throwing the core logic out here first.

a-robinson:
@jimmidyson in case he's interested

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -302,16 +312,41 @@ def write(chunk)
             entry['metadata']['severity'] = 'DEFAULT'
           end
 
+          if @service_name == CONTAINER_SERVICE
+            # TODO(a-robinson): Consider falling back to parsing the namespace,
+            # pod, and container names out of the log name if we don't have this
+            # information, which depends on the kubernetes_metadata_filter
+            # filter plugin.
+            labels = {}
+            field_to_label(record['kubernetes'], 'namespace_id', labels,
```
jimmidyson: How about turning this into an array each loop? Bit more readable IMO.

jimmidyson:
Looks quite intensive for every log event. Not in this PR but noticed it sends each log message off as it comes in. Would a batch/bulk approach be better?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -302,16 +312,41 @@ def write(chunk)
...
+                                  "#{CONTAINER_SERVICE}/namespace_id")
+            field_to_label(record['kubernetes'], 'namespace_name', labels,
+                                  "#{CONTAINER_SERVICE}/namespace_name")
+            field_to_label(record['kubernetes'], 'pod_id', labels,
```
jimmidyson: Why is pod id & namespace id required? I assume these are the respective uids? Is name but sufficient? Are names mutable now?

jimmidyson:
If you need the namespace & pod IDs passed through from the kubernetes metadata filter plugin then please raise an issue (& PR if you have time).

jimmidyson:
If you decide not to use the metadata filter plugin then you will have to duplicate the label retrieval logic, the watches for changes, the caching, etc. Another option is to add a mode config to the metadata filter plugin so that if GCE mode is enabled, metadata is added in a way to make it more consumable for this plugin.

lib/fluent/plugin/out_google_cloud.rb
a-robinson: Done, although it'll only be cleaner if the names coming out of the metadata plugin are the same as the names we want to send off to cloud logging, which'll depend on how you feel about them.

lib/fluent/plugin/out_google_cloud.rb
a-robinson: No, names aren't mutable, and I'm not a big fan of this either, but the Google cloud logging and monitoring services prefer indexing by IDs rather than names. The IDs can then be mapped to names at query time as appropriate.

The main benefit is that it allows for distinguishing between two different objects with the same name that existed at different times (i.e. create foo, delete foo, create foo).

a-robinson:
Agreed on the likely processing cost - I'll try generating a bunch of logs and see how much CPU gets used by this once I've added the additional fields to the metadata plugin. For what it's worth, it doesn't send off each entry separately, though -- it only processes their labels separately. See where it pushes each entry onto `write_log_entries_request['entries']` at the end of the loop, and then sends off the write_log_entries_request at the end?

a-robinson:
And yes, I'll send a PR to the metadata plugin, probably guarded by a config parameter so that the extra namespace ID work is only done when it's being run along with this.

a-robinson:
Tests have been added.

Also, for reference, this fixes #15.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -500,6 +525,38 @@ def parse_severity(severity_str)
...
+        field_to_label(record['kubernetes'], field, labels,
+                       "#{CONTAINER_SERVICE}/#{field}")
+        field_to_label(record['kubernetes'], field, labels,
+                       "#{CONTAINER_SERVICE}/#{field}")
+        field_to_label(record['kubernetes'], field, labels,
+                       "#{CONTAINER_SERVICE}/#{field}")
+        field_to_label(record['kubernetes'], field, labels,
+                       "#{CONTAINER_SERVICE}/#{field}")
+      end
+      # Append label/ to all user-defined labels' keys.
```
mr-salty: s/Append/Prepend/

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -500,6 +525,38 @@ def parse_severity(severity_str)
...
+      if record.key?('labels')
+        record['kubernetes']['labels'].each do |key, value|
+          labels["label/#{key}"] = value
+        end
+      end
+      record.delete('kubernetes')
+      record.delete('docker')
```
mr-salty: Add a comment explaining why we're deleting this without consuming its contents.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -302,16 +312,22 @@ def write(chunk)
...
+            handle_container_metadata(record, entry)
+          end
+
           # If a field is present in the label_map, send its value as a label
           # (mapping the field name to label name as specified in the config)
           # and do not send that field as part of the payload.
-          unless label_map.nil?
+          if @label_map
             labels = {}
+            labels = entry['metadata']['labels'] if entry['metadata']['labels']
```
mr-salty: you could replace these 2 lines with ?:

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -302,16 +312,22 @@ def write(chunk)
...
             @label_map.each do |field, label|
-              next unless record.key?(field)
-              # label values are required to be strings.
-              labels[label] = record[field].to_s
-              record.delete(field)
+              field_to_label(record, field, labels, label)
             end
             entry['metadata']['labels'] = labels unless labels.empty?
```
mr-salty: Although, I'm starting to wonder if it would be cleaner to get rid of storing labels in a local variable (here and in handle_container_metadata) in favor of:

adding 'labels' => {} where we initialize the entry
adding something below that does delete entry['metadata']['labels'] if entry['metadata']['labels'].empty?

I'll leave it up to you.

lib/fluent/plugin/out_google_cloud.rb
a-robinson: The reason I didn't do this initially is that it's still two lines, but I've changed it so that at least it's only one statement even if it's still two lines

lib/fluent/plugin/out_google_cloud.rb
a-robinson: Done

lib/fluent/plugin/out_google_cloud.rb
a-robinson: Thanks, fixed

lib/fluent/plugin/out_google_cloud.rb
a-robinson: I'll leave as is for now so you can pick your preferred style separately, if that's alright.

mr-salty:
LG, pending the CI build passing (looks like it's stuck?  I'll see what I can do to kick it if it doesn't pass soon, but 'bundle exec rake' in your client does the same thing)

a-robinson:
Ah thanks, I hadn't actually checked out the Rakefile to see the rubocop option. It complains about the multiline ternary operator, so I guess I'll switch that back.

a-robinson:
Now the rubocop runs cleanly, along with the tests.

mr-salty:
Thanks.  Yeah, following the style guide is trying at times, but I decided against trying to tweak it (other than relaxing some of their length/complexity rules which seem far too restrictive).

Looks like travis was having some issues, I kicked the build but if it's passing locally and you're ready to submit I can merge it, LMK.  https://www.traviscistatus.com/

a-robinson:
Yes -- it passes locally and I'm ready to have it merged.

mr-salty:
SG, I'll do a gem release tomorrow.

a-robinson:
Thanks! It may make more sense to wait until I can test it end-to-end with the updated kubernetes-metadata plugin, though, just in case anything else comes up from that. I'll have a PR out for that today.

mr-salty:
ok, that SGTM

PR #31, opened 2015-09-10
Add support for attaching container metadata based on the tag.

a-robinson:
This will be useful in the short term, because the kubernetes metadata
filter plugin isn't actually broadly usable until another kubernetes
feature (the daemon controller) gets in that will allow the metadata
plugin to authenticate to the kubernetes API.

This breaks rubocop due to the long write() function. It probably needs broader refactoring than what belongs in this commit, but I'm happy to fix it up here if you'd like.

@mr-salty @jimmidyson @brendandburns

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -21,7 +21,7 @@ Metrics/CyclomaticComplexity:
 # Offense count: 5
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
-  Max: 100
+  Max: 150
 
 # Offense count: 3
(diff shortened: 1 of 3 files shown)


mr-salty:
Argh, rubocop... can you edit .rubocop_todo.yml and change the max method length to 150 or something?

mr-salty:
I mean, that method definitely could use some refactoring, but rubocop is driving me nuts.  I'll take care of it one of these days.  The default is 10!

a-robinson:
Sure, done.

mr-salty:
LGTM

PR #34, opened 2015-09-11
Add option to allow for forcing the service of logs to be GCE, ignoring

a-robinson:
the instance metadata.

The plan for Kubernetes is to have 2 out_google_cloud stanzas in our
configuration, one which handles node daemon logs and sets this to true,
and one which handles the container logs and doesn't set this.

@mr-salty @vishh @swalkowski

Cross reference https://github.com/kubernetes/kubernetes/issues/13033

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -62,6 +62,14 @@ module Fluent
     config_param :vm_id, :string, :default => nil
     config_param :vm_name, :string, :default => nil
 
+    # Whether to try to detect of the VM is owned by a "subservice" such as App
+    # Engine of Kubernetes, rather than just associating the logs with the
+    # compute service of the platform. This currently only has any effect when
+    # running on GCE.
+    # The initial motivation for this is to separate out Kubernetes node
+    # component (Docker, Kubelet, etc.) logs from container logs.
+    config_param :detect_subservice, :bool, :default => true
+
     # label_map (specified as a JSON object) is an unordered set of fluent
     # field names whose values are sent as labels rather than as part of the
     # struct payload.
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -189,8 +195,10 @@ def configure(conf)
         # Check for specialized GCE environments (Managed VM or Dataflow).
         # TODO: Add config options for these to allow for running outside GCE?
         attributes = fetch_gce_metadata('instance/attributes/').split
-        if attributes.include?('gae_backend_name') &&
-           attributes.include?('gae_backend_version')
+        if @force_compute_service
```
mr-salty: We don't need to fetch 'attributes' at all if this is true, so maybe omit the whole thing, i.e.

```
unless @force_compute_service
  attributes = ...
  if attributes ...
end
```

I'm not crazy about the name 'force_compute_service', although I can't come up with a better name.  It seems like we're really disregarding the more-specific service, but I'm not sure how to capture that succinctly.  I'm ok leaving it as is if we can't come up with a better name.

a-robinson: Yeah, I'm not the biggest fan either, but it seemed the best framing of it I could come up with. It's better than a string field that lets you pick the service name because we may not be able to populate the other fields for that service name (or even know what it is). It's more general than a parameter specifically for disabling the container service and could potentially be useful in other similar cases. Also, naming a parameter that would do this but only for the container service would seem to be even harder. ignore_kube_env could perhaps work, but kube-env is more of an implementation detail than something most people know anything about.

Do you have any ideas?

a-robinson: And as for fetching attributes, I've switched it around.

mr-salty: Not really...  coming up with names is definitely the hardest part.  Definitely not 'kube' since it also affects MVM, dataflow, etc.  Maybe "detect_gce_subservice" defaulting to true?  Or, I'm fine leaving it as-is.

a-robinson: detect_gce_subservice sounds good to me. It might be slightly pigeonholing the config options if you later want to support subservices on non-GCE platforms (as the TODO indicates). Maybe just detect_subservice or even detect_service? detect_service would match the fact that the platforms are sort of separate from the services.

mr-salty: detect_subservice SGTM, I think that makes sense.  Normally we set the service name based on the platform, but when this is true we try to determine if we're really running as part of some other service.

a-robinson: Great, done.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -62,6 +62,13 @@ class GoogleCloudOutput < BufferedOutput
     config_param :vm_id, :string, :default => nil
     config_param :vm_name, :string, :default => nil
 
+    # Associate logs with the compute service of the platform rather than
+    # attempting to detect if the VM is owned by a subservice such as App Engine
+    # or Kubernetes. This currently only has any effect when running on GCE.
+    # The initial motivation for this is to separate out Kubernetes node
+    # component (Docker, Kubelet, etc.) logs from container logs.
+    config_param :detect_subservice, :bool, :default => false
```
mr-salty: default => true

I guess this also means I should bug you to add a test :)

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -186,34 +193,37 @@ def configure(conf)
...
-          common_labels["#{DATAFLOW_SERVICE}/job_id"] = @dataflow_job_id
-        elsif attributes.include?('kube-env')
-          # Kubernetes/Container Engine
-          @service_name = CONTAINER_SERVICE
-          common_labels["#{CONTAINER_SERVICE}/instance_id"] = @vm_id
-          @raw_kube_env = fetch_gce_metadata('instance/attributes/kube-env')
-          @kube_env = YAML.load(@raw_kube_env)
-          common_labels["#{CONTAINER_SERVICE}/cluster_name"] =
-            cluster_name_from_kube_env(@kube_env)
+        unless @detect_subservice
```
mr-salty: and this should be 'if'

lib/fluent/plugin/out_google_cloud.rb
a-robinson: Haha well technically combined with still using `unless` below the default was working as intended :)

a-robinson:
I wasn't sure if you were waiting for a test or just done for the day, so I've gone ahead and added a test. Let me know when you have a new google-fluentd image built with all this in it. Thanks!

mr-salty:
Thanks for adding the test, I will build it tomorrow.

PR #36, opened 2015-09-22
Fix the kubernetes tag regexp to allow pod names with periods in them.

a-robinson:
Also make the regexp configurable, since the default won't work for
configurations in which there's more than one component of the tag
before the part we care about.

@mr-salty @swalkowski

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -70,6 +70,12 @@ module Fluent
     # component (Docker, Kubelet, etc.) logs from container logs.
     config_param :detect_subservice, :bool, :default => true
 
+    # The regular expression to use on Kubernetes logs to extract some basic
+    # information about the log source. The regex must contain capture groups
+    # for pod_name, namespace_name, and container_name.
+    config_param :kubernetes_tag_regexp, :string, :default =>
+      '\.(?<pod_name>[^_]+)_(?<namespace_name>[^_]+)_(?<container_name>.+)$'
+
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -70,6 +70,12 @@ class GoogleCloudOutput < BufferedOutput
     # component (Docker, Kubelet, etc.) logs from container logs.
     config_param :detect_subservice, :bool, :default => true
 
+    # The regular expression to use on Kubernetes logs to extract some basic
+    # information about the log source. The regex must contain capture groups
+    # for pod_name, namespace_name, and container_name.
+    config_param :kubernetes_tag_regexp, :string, :default =>
+      '\.(?<pod_name>[^_]+)_(?<namespace_name>[^_]+)_(?<container_name>.+)$'
```
jimmidyson: Why not include `var.log.containers.` prefix in there as default? It's the default for kubernetes file path AFAIK so should be a good default IMO.

a-robinson: It's not the default being used in our fluentd-gcp config, since there we're using the record reformer to strip that stuff out of the tag:
https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-gcp/fluentd-gcp-image/google-fluentd.conf

jimmidyson: Aha fair enough.

mr-salty:
LGTM

PR #37, opened 2015-09-22
Fix misordered expectation and value in assert statement.

a-robinson:
This makes the error message label the expectation and actual value
correctly rather than mixing them up.

Just a minor thing I bumped into while verifying #36.

@mr-salty


mr-salty:
LGTM

PR #38, opened 2015-09-28
Add support for Cloud Functions logs

swalkowski:
This way logs coming from functions are processed in a custom way, by extracting certain labels (function name, execution id) and fields (timestamp, severity) from tag name and log text, and marking the entries with Cloud Functions service name.

@mr-salty @rsokolowski

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -19,6 +19,7 @@ module Fluent
 
     # Constants for service names.
     APPENGINE_SERVICE = 'appengine.googleapis.com'
+    CLOUDFUNCTIONS_SERVICE = 'cloudfunctions.googleapis.com'
     COMPUTE_SERVICE = 'compute.googleapis.com'
     CONTAINER_SERVICE = 'container.googleapis.com'
     DATAFLOW_SERVICE = 'dataflow.googleapis.com'
@@ -154,6 +155,14 @@ module Fluent
         @compiled_kubernetes_tag_regexp = Regexp.new(@kubernetes_tag_regexp)
       end
 
+      @cloudfunctions_tag_regexp =
+        /\.(?<function_name>.+)-[^-]+_default_worker$/
+      @cloudfunctions_log_regexp = /^
+        (?:\[(?<severity>.)\])?
+        \[(?<timestamp>.{24})\]
+        (?:\[(?<execution_id>[^\]]+)\])?
+        [ ](?<text>.*)$/x
+
       # set attributes from metadata (unless overriden by static config)
       @vm_name = Socket.gethostname if @vm_name.nil?
       @platform = detect_platform
@@ -199,12 +208,16 @@ module Fluent
       # Default this to false; it is only overwritten if we detect Managed VM.
       @running_on_managed_vm = false
 
+      # Default this to false; it is only overwritten if we detect Cloud
+      # Functions.
+      @running_cloudfunctions = false
+
       # Set labels, etc. based on the config
       case @platform
       when Platform::GCE
         @service_name = COMPUTE_SERVICE
         if @detect_subservice
-          # Check for specialized GCE environments (Managed VM or Dataflow).
+          # Check for specialized GCE environments.
           # TODO: Add config options for these to allow for running outside GCE?
           attributes = fetch_gce_metadata('instance/attributes/').split
           # Do nothing, just don't populate other service's labels.
@@ -233,6 +246,7 @@ module Fluent
             @kube_env = YAML.load(@raw_kube_env)
             common_labels["#{CONTAINER_SERVICE}/cluster_name"] =
               cluster_name_from_kube_env(@kube_env)
+            detect_cloudfunctions(attributes, common_labels)
           end
         end
         common_labels["#{COMPUTE_SERVICE}/resource_type"] = 'instance'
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -294,8 +308,29 @@ def write(chunk)
...
+            # Service name is set to Cloud Functions only for logs actually
+            # coming from a function.
+            @service_name = CLOUDFUNCTIONS_SERVICE
+            labels = write_log_entries_request['commonLabels']
+            labels["#{CLOUDFUNCTIONS_SERVICE}/function_name"] =
+              match_data['function_name']
+          else
+            # Other logs are considered as coming from the Container Engine
+            # service.
+            @service_name = CONTAINER_SERVICE
```
mr-salty: service_name should already equal CONTAINER_SERVICE here, correct?  Should we just leave it unmodified?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -492,6 +526,14 @@ def fetch_ec2_metadata
       end
     end
 
+    def detect_cloudfunctions(attributes, common_labels)
+      return unless attributes.include?('gcf_region')
+      # Cloud Functions detected
+      @running_cloudfunctions = true
+      @gcf_region = fetch_gce_metadata('instance/attributes/gcf_region')
+      common_labels["#{CLOUDFUNCTIONS_SERVICE}/region"] = @gcf_region
```
mr-salty: This gets included on entries that fail the 'if matchdata' test above on line 315 - seems like it should not be?

lib/fluent/plugin/out_google_cloud.rb
mr-salty: also see the comment below about the 'region' label.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -294,8 +308,29 @@ def write(chunk)
...
+              match_data['function_name']
+          else
+            # Other logs are considered as coming from the Container Engine
+            # service.
+            @service_name = CONTAINER_SERVICE
+          end
+        end
         arr.each do |time, record|
           next unless record.is_a?(Hash)
+          if @service_name == CLOUDFUNCTIONS_SERVICE && record.key?('log')
```
mr-salty: is the format of 'log' set in stone or could you emit it as json instead?  I think we could eliminate some of the special-case code if so?

test/plugin/test_out_google_cloud.rb
```diff
@@ -839,6 +877,32 @@ def test_multiple_container_logs_metadata_from_tag
...
+      # The test driver doesn't clear its buffer of entries after running, so
+      # do it manually here.
+      d.instance_variable_get('@entries').clear
+      @logs_sent = []
+      n.times { |i| d.emit(cloudfunctions_log_entry(i)) }
+      d.run
+      verify_log_entries(n, CLOUDFUNCTIONS_PARAMS)
+    end
+  end
+
```
mr-salty: Add tests for the cases where running_cloudfunctions is true but the tag or log do not match the regex

mr-salty:
PTAL - looks good in general, but see my comment about making 'log' JSON which would definitely simplify things if it is possible.

lib/fluent/plugin/out_google_cloud.rb
swalkowski: This is run in a loop (line 294) so service_name could be set to CLOUDFUNCTIONS_SERVICE in previous iteration while we want it to be CONTAINER_SERVICE in this iteration. The general problem here is that I want the service_name to depend on the value of tag so it can change with every loop iteration. I could potentially introduce a local variable like actual_service_name, set it inside the loop and use it when creating a log entry - but I am not sure if this wouldn't be more confusing after all.

lib/fluent/plugin/out_google_cloud.rb
swalkowski: I think either way would be okay but it is indeed cleaner not to include this label if the service_name is not CLOUDFUNCTIONS_SERVICE eventually. So I am moving setting this label to the 'if match_data' block.

lib/fluent/plugin/out_google_cloud.rb
swalkowski: The format of 'log' will be rather static although it is possible that it will evolve over time. I could potentially include a stringified JSON as the 'log' string value and parse it in the plugin, if this is what you mean? I would still need to check if this is what we want, though.

Emitting a true JSON object as the 'log' value may not be possible as the overall composition of the emitted records is rather not controllable.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: I see.  Yeah, this is probably the best option for now.  I think I'm overdue for a big refactoring since we've grafted on a lot of functionality and it's getting unwieldy.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: right now, your logs look something like this, right:

 { "stream": "stdout", "log": "[D][2015-09-25T12:34:56.789Z][123456789-0]test log entry 0" }

I assume in the real world they're written as JSON and parsed by fluentd ('format json'), or similar?

if we made this { "stream": "stdout", "log": { "severity": "D", "timestamp": "2015-09-25T12:34:56.789Z", "execution_id": "123456789-0", "text": "test log entry 0" } }

Then fluentd would parse it for you and we wouldn't need the regex.

Even better would be if we could drop "log" and move all those fields to the top level; then it would handle the severity and timestamp (you'd have to rename it to 'time', and there's no nanosecond support yet but it is coming) automatically, and we could use label_map in the config file to map execution_id to a label, and the bulk of the custom code here wouldn't be necessary.

swalkowski: As it is not really possible to emit true JSON, we can only have a string as the "log" field value. In this situation, I would keep the logic based on parsing this string through a regexp.

test/plugin/test_out_google_cloud.rb
swalkowski: Tests added.

swalkowski:
Thanks for reviewing, I have addressed the comments in the second commit. PTAL.

mr-salty:
LGTM

PR #39, opened 2015-10-08
Use the Kubernetes container name as the log name rather than the tag.

a-robinson:
This prevents too many different log streams from being created for pods
in a replication controller, and allows for much better aggregation.

It could be confusing if people don't name their containers well, but in our experience most people tend to just name their container the same as the pod that it's in, and don't use meaningless names for them.

If it's alright, I'm going to be shortly following this up by also including whether it's stdout or stderr in the log name as well, making for names like "redis.stdout" and "redis.stderr" coming from a redis container. This'll also make it such that the payload can be just the log message, with no unnecessary json boilerplate. That'll take a little more surgery, though, since each batch of logs for the tag might have to be split into two different streams, which isn't supported at the moment.

@mr-salty

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -417,7 +417,8 @@ module Fluent
         # Don't send an empty request if we rejected all the entries.
         next if write_log_entries_request['entries'].empty?
 
-        log_name = CGI.escape(log_name(tag))
+        log_name = CGI.escape(
+          log_name(tag, write_log_entries_request['commonLabels']))
         url = 'https://logging.googleapis.com/v1beta3/projects/' \
           "#{@project_id}/logs/#{log_name}/entries:write"
         begin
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -664,13 +665,21 @@ def set_payload(record, entry)
...
+        return "#{APPENGINE_SERVICE}/#{tag}"
+      elsif @service_name == CONTAINER_SERVICE
+        # For Kubernetes logs, use just the container name as the log name
+        # if we have it.
+        container_name_key = "#{CONTAINER_SERVICE}/container_name"
+        if commonLabels && commonLabels.key?(container_name_key)
+          return commonLabels[container_name_key]
+        end
       end
+      tag
```
mr-salty: ha, i was trying to comment on this but i see you already fixed it.  One of the things I disagree with the most, but...

a-robinson: sorry :) I forgot to run rubocop before pushing

mr-salty:
LGTM

mr-salty:
(I wish there was a "merge this once the checks pass" button, are you aware of anything like that?)

a-robinson:
Nope, I'm not. We ended up [building our own submit queue / merge-bot](https://github.com/kubernetes/contrib/tree/master/submit-queue).

PR #40, opened 2015-10-12
Make kubernetes logs map the stream into a label rather than the payload

a-robinson:
In the process, switch over from using structPayload to textPayload.

Picking this over the alternatives of splitting the logs into two separate streams (containername.stdout and containername.stderr) or by shoehorning this into the severity field. Not choosing separate streams because it makes it tougher to aggregate them and aggregating them seems like the more common case. Not choosing to use the severity field because it doesn't really fit the model as well -- what if we later want to actually parse severity out of container logs, or if docker starts allowing more stream names than just stdout and stderr?

I'm happy to use a different label than container.googleapis.com/stream if you have other suggestions. container.googleapis.com/stream_name, perhaps?

@mr-salty for review
@bokowski @fabioy @roberthbailey @vishh as people I discussed the choice with in case they have additional thoughts.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -382,11 +382,16 @@ module Fluent
 
           set_severity(record, entry)
 
-          # If the record has been annotated by the kubernetes_metadata_filter
-          # plugin, then use that metadata. Otherwise, rely on commonLabels
-          # populated at the grouped_entries level from the group's tag.
-          if @service_name == CONTAINER_SERVICE && record['kubernetes']
-            handle_container_metadata(record, entry)
+          if @service_name == CONTAINER_SERVICE
+            # Move the stdout/stderr annotation from the record into a label
+            field_to_label(record, 'stream', entry['metadata']['labels'],
+                           "#{CONTAINER_SERVICE}/stream")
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -656,7 +661,8 @@ def set_payload(record, entry)
       # available, or if the only remainaing key is 'message'.
       if @service_name == CLOUDFUNCTIONS_SERVICE && @cloudfunctions_log_match
         entry['textPayload'] = @cloudfunctions_log_match['text']
-      elsif @service_name == CLOUDFUNCTIONS_SERVICE && record.key?('log')
+      elsif (@service_name == CLOUDFUNCTIONS_SERVICE ||
+             @service_name == CONTAINER_SERVICE) && record.key?('log')
```
mr-salty: might you want to do this only if 'log' is the sole remaining field (like 'message' below) to allow for the possibility of using struct if there are additional fields?

a-robinson: I actually think the control that doing it this way gives us to manually curate what makes it through is better.

I don't want to start spitting out json payloads at users in the log viewer just because the metadata filter plugin starts adding an additional field to the record that we miss when making a new fluentd container image.

mr-salty:
LGTM

PR #41, opened 2015-10-16
Parse container logs into a structured log if their body is json.

a-robinson:
Treat them exactly the same as json logs from other sources if they are
json.

Avoid trying to parse every entry as json. If the first entry in a batch
isn't json, then don't try to parse the rest as json, since it's just a
waste of CPU. JSON parsing is unfortunately not cheap. We might consider
switching over to a faster parser (like https://github.com/ohler55/oj).

@mr-salty

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -326,12 +326,54 @@ module Fluent
             @service_name = CONTAINER_SERVICE
           end
         end
+        is_container_json = nil
         arr.each do |time, record|
           next unless record.is_a?(Hash)
+
+          entry = {
+            'metadata' => {
+              'serviceName' => @service_name,
+              'projectId' => @project_id,
+              'zone' => @zone,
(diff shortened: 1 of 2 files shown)


mr-salty:
Thanks - LG in general, can you fix the rubocop gripes?

I wonder if we should provide a config option to force disable this, although I'm not sure if it's more trouble than it's worth... I'll leave it up to you.

a-robinson:
Ah, sorry! One day I'll get better about checking rubocop... I checked it after the plugin changes but not after the test changes.

I can't say I think this is better, but rubocop likes it.

a-robinson:
As for the config option, I don't think it's likely to come up, so I'll skip it unless you think otherwise.

mr-salty:
SG on the config option...  and I feel that way about a lot of what rubocop complains about, but feel obligated to obey it.

FWIW, 'rake' by itself will run tests + rubocop

I'll get this merged and built along with the other changes I'm waiting for before you're back.

PR #44, opened 2015-11-05
Obtain the project_id from credentials if needed.

mr-salty:
If the project_id has not been set in the config file and cannot be
obtained via the metadata service, try to get it from the credentials.

Add a test for this, and change to using constants for the paths to the
credentials files.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -195,6 +195,14 @@ module Fluent
         fail Fluent::ConfigError, 'Unknown platform ' + @platform
       end
 
+      # If we still don't have a project ID, try to obtain it from the
+      # credentials.
+      if @project_id.nil?
+        @project_id = CredentialsInfo.project_id
+        @log.info 'Set Project ID from credentials: ', @project_id unless
+          @project_id.nil?
+      end
+
       # all metadata parameters must now be set
       unless @project_id && @zone && @vm_id
         missing = []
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -195,6 +195,14 @@ def configure(conf)
         fail Fluent::ConfigError, 'Unknown platform ' + @platform
       end
 
+      # If we still don't have a project ID, try to obtain it from the
+      # credentials.
+      if @project_id.nil?
+        @project_id = CredentialsInfo.project_id
```
filipjs: Where is "CredentialsInfo" coming from?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -564,6 +572,37 @@ def fetch_ec2_metadata
...
 
+    # TODO: This functionality should eventually be available in another
+    # library, but implement it ourselves for now.
+    module CredentialsInfo
+      # Determine the project ID from the credentials, if possible.
+      # Returns the project ID (as a string) on success, or nil on failure.
+      def self.project_id
+        return nil if @auth_method == 'private_key'
+        creds = Google::Auth.get_application_default(LOGGING_SCOPE)
+        if creds.issuer
```
filipjs: "issuer"
Is that expected or should be "isuser"?

filipjs:
LGTM

lib/fluent/plugin/out_google_cloud.rb
mr-salty: No, it's "issuer" - http://www.rubydoc.info/github/google/signet/Signet/OAuth2/Client#issuer-instance_method

PR #45, opened 2015-11-05
Log a message containing the URL of the log viewer.

mr-salty:
This is currently done by the install script, but it is broken for
non-GCE platforms and is better done here, so it will also work for
people who don't use the install script.


filipjs:
LGTM

PR #46, opened 2015-11-11
Parse project_id from IAM account credentials.

mr-salty:
This new credentials format is currently rolling out to all users; the
email address format is account@project-id.iam.gserviceaccount.com.
The existing format is still supported.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -54,6 +54,10 @@ module Fluent
     # can be obtained from the metadata service or set explicitly.
     # Otherwise, the plugin will fail to initialize.
     #
+    # Note that while 'project id' properly refers to the alphanumeric name
+    # of the project, the logging service will also accept the project number,
+    # so either one is acceptable in this context.
+    #
     # Whether to attempt to obtain metadata from the local metadata service.
     # It is safe to specify 'true' even on platforms with no metadata service.
     config_param :use_metadata_service, :bool, :default => true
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -596,15 +596,20 @@ def self.project_id
...
+      #
+      # Recognizes IAM format (account@project-name.iam.gserviceaccount.com)
+      # as well as the legacy format with a project number at the front of the
+      # string, terminated by a dash (-) which is not part of the ID, i.e.:
+      # 270694816269-1l1r2hb813leuppurdeik0apglbs80sv.apps.googleusercontent.com
       def self.extract_project_id(str)
-        @project_regexp = /^(?<project_id>\d+)-/
-        match_data = @project_regexp.match(str)
-        match_data ? match_data['project_id'] : nil
+        [/^.*@(?<project_id>.+)\.iam\.gserviceaccount\.com/,
```
filipjs: For IAM, is this project ID or name? regexp suggests it's name, while variable name that it's ID
Can you also update the comment above to make it clear?

filipjs:
LGTM

lib/fluent/plugin/out_google_cloud.rb
mr-salty: yeah, it's a bit confusing, I added some explanatory text here as well as earlier in the file.  The logging API will accept either the name (which is what "id" properly means) or number, so we don't really care too much which one we get here.

PR #49, opened 2015-12-17
Support encoded names of cloud functions and set service name properly

swalkowski:
Adding support for encoded function names extracted from a tag. Updating the Cloud Functions tag regexp to reflect the expected format.

Moving the block which sets current service name when running Cloud Functions before the name is read.

@mr-salty


mr-salty:
LGTM

PR #52, opened 2016-01-12
Add a severity parsing mapping from "F" to "CRITICAL".

a-robinson:
It should really map to FATAL, but that doesn't seem to be an option in
the API. For what it's worth, fatal is the highest log level in:
https://github.com/google/glog
https://golang.org/pkg/log/

@mr-salty


mr-salty:
This was already done in #47 and released in v0.4.13

a-robinson:
This is what I get for working off an old version of the code on an airplane. Thanks!

mr-salty:
NP... good to see rubocop catch something useful - I was perplexed at first because I forgot about the earlier change and the "duplicate key" wasn't obvious from the diff.

PR #57, opened 2016-03-13
Extract httpRequest object from structPayload

n0nick:
When structPayload contains a httpRequest object with valid fields [1],
extract these values into a new root httpRequest object so Google Cloud
Logging would pick them up as HTTP request attributes.

Hopefully, this would show up in Log Viewer similar to how HTTP requests
are logged for App Engine projects.

[1] https://cloud.google.com/logging/docs/api/ref/rest/v1beta3/projects.logs.entries/write#HttpRequest

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -31,6 +31,11 @@ module Fluent
     # Address of the metadata service.
     METADATA_SERVICE_ADDR = '169.254.169.254'
 
+    # Fields allowed in httpRequest object
+    HTTP_REQUEST_FIELDS = %w(requestMethod requestUrl requestSize status
+                             responseSize userAgent remoteIp referer
+                             cacheHit validatedWithOriginServer)
+
     # Disable this warning to conform to fluentd config_param conventions.
     # rubocop:disable Style/HashSyntax
 
@@ -434,6 +439,8 @@ module Fluent
 
           set_severity(record, entry)
 
+          set_http_request(record, entry)
+
           # If a field is present in the label_map, send its value as a label
           # (mapping the field name to label name as specified in the config)
           # and do not send that field as part of the payload.
@@ -654,6 +661,21 @@ module Fluent
       else
         entry['metadata']['severity'] = 'DEFAULT'
       end
+    end
+
+    def set_http_request(record, entry)
+      return unless record['httpRequest'].is_a?(Hash)
+
+      entry['httpRequest'] = {}
+
+      HTTP_REQUEST_FIELDS.each do |field|
+        if record['httpRequest'].key?(field)
+          entry['httpRequest'][field] = record['httpRequest'][field]
+          record['httpRequest'].delete(field)
+        end
+      end
+
+      record.delete('httpRequest') if record['httpRequest'].empty?
     end
 
     # Values permitted by the API for 'severity' (which is an enum).
--- a/test/plugin/test_out_google_cloud.rb
+++ b/test/plugin/test_out_google_cloud.rb
@@ -299,6 +299,19 @@ class GoogleCloudOutputTest < Test::Unit::TestCase
     }
   }
 
+  HTTP_REQUEST_MESSAGE = {
+    'requestMethod' => 'POST',
+    'requestUrl' => 'http://example/',
+    'requestSize' => 210,
+    'status' => 200,
+    'responseSize' => 65,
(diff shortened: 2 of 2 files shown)


n0nick:
The [failing job](https://travis-ci.org/GoogleCloudPlatform/fluent-plugin-google-cloud/jobs/115652436) is on Ruby 1.9.3, due to the `jwt` gem version 1.5.3 using the `%i` literal which was only introduced in Ruby 2.0.

It seems that `jwt` is [required by `googleauth`](https://github.com/google/google-auth-library-ruby/blob/master/googleauth.gemspec#L30) which specifies `~> 1.4`, so 1.5.3 is a valid choice.

IMO this is not related to my PR.

mr-salty:
gem dependencies are a constant source of fun :)  I agree it's unrelated; I'll look into it.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -656,6 +663,21 @@ def set_severity(record, entry)
...
+      entry['httpRequest'] = {}
+
+      HTTP_REQUEST_FIELDS.each do |field|
+        if record['httpRequest'].key?(field)
+          entry['httpRequest'][field] = record['httpRequest'][field]
+          record['httpRequest'].delete(field)
+        end
+      end
+
+      record.delete('httpRequest') if record['httpRequest'].empty?
```
mr-salty: I wonder if we should unambiguously delete the httpRequest even if it has "unknown" fields.  I don't have a really strong feeling either way, any opinion as a user of it?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -656,6 +663,21 @@ def set_severity(record, entry)
       end
     end
 
+    def set_http_request(record, entry)
+      return unless record.key?('httpRequest')
```
mr-salty: you should also return if !record['httpRequest'].is_a?(Hash) or it could throw an exception.  Please add a test for this as well.

test/plugin/test_out_google_cloud.rb
```diff
@@ -1048,6 +1061,18 @@ def test_multiple_cloudfunctions_logs_tag_not_matched
...
+    setup_logging_stubs
+    d = create_driver(APPLICATION_DEFAULT_CONFIG)
+    d.emit('httpRequest' => HTTP_REQUEST_MESSAGE)
+    d.run
+    verify_log_entries(1, COMPUTE_PARAMS, 'httpRequest') do |entry|
+      assert_equal HTTP_REQUEST_MESSAGE, entry['httpRequest'], entry
+      assert_equal nil, entry['structPayload']['httpRequest'], entry
+    end
+  end
+
```
mr-salty: add a test for the case I mentioned above (httpRequest not a hash) as well as httpRequest with "extra" fields (make sure they make it into the struct payload, or not, depending on what we decide there).

lib/fluent/plugin/out_google_cloud.rb
n0nick: Good point. Will push an update in a moment.

n0nick:
FYI I just ran this code on a local fluent agent -- it works (`httpRequest` is moved to root), but I didn't get the log entry view experience that I wanted (AppEngine-like request lines with method/path extracted etc).

Kind of disappointing :/

test/plugin/test_out_google_cloud.rb
```diff
@@ -1048,6 +1061,42 @@ def test_multiple_cloudfunctions_logs_tag_not_matched
...
+  def test_http_request_when_not_hash
+    setup_gce_metadata_stubs
+    setup_logging_stubs
+    d = create_driver(APPLICATION_DEFAULT_CONFIG)
+    d.emit('httpRequest' => 'a_string')
+    d.run
+    verify_log_entries(1, COMPUTE_PARAMS, 'structPayload') do |entry|
+      assert_equal 'a_string', entry['structPayload']['httpRequest'], entry
```
mr-salty: can we assert_equal nil, entry['httpRequest'], entry here as well?

mr-salty:
Sorry, I didn't realize you had pushed more changes.  One small test comment, LG otherwise.

test/plugin/test_out_google_cloud.rb
n0nick: :+1:  done

mr-salty:
LGTM

Issue #55, opened 2016-02-09
update to google-api-client 0.9


mr-salty:
https://rubygems.org/gems/google-api-client 0.9 has been released and this plugin requires source changes to work with the new version (see #54).

This becomes an issue if the user installs a gem which requires 0.9 (for instance, https://rubygems.org/gems/fluent-plugin-bigquery) since rubygems resolves the dependency by installing 0.9 causing our plugin to fail.

mr-salty:
See https://github.com/google/google-api-ruby-client/blob/master/MIGRATING.md

PR #60, opened 2016-03-28
Issue 55

laramiel:
This builds on my prior pull requests.

In addition, it update the google cloud logging plugin to use the google-api > 0.9, which resolves Issue #55.

### DIFF
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -25,6 +25,8 @@ eos
   # workaround for jwt 1.5.3 breaking ruby 1.9 support (included by googleauth)
   # see https://github.com/jwt/ruby-jwt/issues/132
   gem.add_runtime_dependency 'jwt', '< 1.5.3'
+  # workaround for mime_times 3.+ breaking ruby 1.9 support (google-api-client)
+  gem.add_runtime_dependency 'mime-types', '< 3.0'
 
   gem.add_development_dependency 'mocha', '~> 1.1'
   gem.add_development_dependency 'rake', '~> 10.3'
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -19,7 +19,7 @@ eos
   gem.require_paths = ['lib']
 
   gem.add_runtime_dependency 'fluentd', '~> 0.10'
-  gem.add_runtime_dependency 'google-api-client', '>= 0.8.6', '< 0.9'
+  gem.add_runtime_dependency 'google-api-client', '> 0.9'
   gem.add_runtime_dependency 'googleauth', '~> 0.4'
   gem.add_runtime_dependency 'json', '~> 1.8'
   # workaround for jwt 1.5.3 breaking ruby 1.9 support (included by googleauth)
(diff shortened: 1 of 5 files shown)


laramiel:
Updated the mime-types gemspec.

Note: This change removes the private_key auth method (see my remove_pk_auth branch), since google-api > 0.9 is not compatible with the old authentication method.

Perhaps making a release 0.4.16 or something would be a useful change before merging this?

laramiel:
Well, that fails on google-api-client 0.9 and above due to the use of __dir__.

https://github.com/google/google-api-ruby-client/issues/387

mr-salty:
Heh, we just can't win.  Well, maybe it's time to show 1.9 the door - I don't have a good sense for how many people still use it.

mr-salty:
Also, on the "cutting a new release", that's certainly a good idea but I can release at other than HEAD so it doesn't have to block this CL.

I also meant to say, google-fluentd itself ships with ruby 2.1 (as does td-agent) so 1.9 is not a concern there...   it's only if someone does their own ruby install and installs fluentd "from scratch" that 1.9 even becomes an issue.  And, 1.9 has been officially EOL for over a year: https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/

laramiel:
Then time to remove 1.9.x from Travis?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -117,6 +97,13 @@ class GoogleCloudOutput < BufferedOutput
     #   }
     config_param :label_map, :hash, :default => nil
 
+    # DEPRECATED: The following parameters, if present in the config
+    # indicate that the plugin configuration must be updated.
+    config_param :auth_method, :string, :default => nil
+    config_param :private_key_email, :string, :default => nil
+    config_param :private_key_path, :string, :default => nil
+    config_param :private_key_passphrase, :string, :default => nil
```
mr-salty: keep ```:secret => true``` here which prevents its value from being logged.

test/plugin/test_out_google_cloud.rb
```diff
@@ -117,9 +118,9 @@ def setup
 
   # rubocop:disable Metrics/LineLength
   PRIVATE_KEY_CONFIG = %(
```
mr-salty: on line 1215 there's a http stub with a comment "Used for 'private_key' auth." which I presume we can remove (sorry, github won't let me comment on the actual line).

mr-salty:
I removed support for ruby 1.9 in  https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/61 - PTAL and merge that.  LGTM otherwise - thanks!

fluent-plugin-google-cloud.gemspec
```diff
@@ -14,7 +14,7 @@ eos
   gem.authors       = ['Todd Derr', 'Alex Robinson']
   gem.email         = ['salty@google.com']
 
-  gem.required_ruby_version = '~> 2.0'
```
laramiel: I think that this works as well as the new one.

mr-salty: oops - I didn't notice you had that in there.  I looked at a few other gems which used the Requirement::new syntax (despite using the shorter syntax later in the gemspec), but I wasn't able to determine the justification for doing it.

Either way is fine with me but I'd rather use >= 2.0 than ~>

laramiel: Sure. I have no problem with that.

mr-salty:
LGTM - thanks again!

PR #62, opened 2016-04-11
Decompose out_google_cloud.rb to pull out attribute extraction & client

laramiel:
This pull request decomps out_google_cloud.rb to pull out some of the
attribute detection and authorization into a separate file, called google_util.rb.

I use this in my next change.


.rubocop_todo.yml
```diff
@@ -1,28 +1,34 @@
...
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
-# Offense count: 2
+# Offense count: 33
 Metrics/AbcSize:
-  Max: 120
+  Max: 100
```
mr-salty: FWIW, I had included some padding in some of these because it was driving me nuts that every time someone sent a PR we had to update the rubocop config.  I also pinned rubocop to a specific version to prevent spurious failures.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -97,6 +88,22 @@ class GoogleCloudOutput < BufferedOutput
...
+    # are provided by GCE, or through the the 'gcloud auth' command.
+    # This requires no parameters, though it may read credentials
+    # located by the environment variable GOOGLE_APPLICATION_CREDENTIALS.
+    #
+    # Method 2: Use Service account keys.
+    #
+    # service_credentials_path is a path to a service account key file,
+    # in json format, which can be acquired from the Google Cloud Developers
+    # Console. When set, the other fields are unnecessary.
+    config_param :service_credentials_path, :string, :default => nil
```
mr-salty: When I did the ADC support I convinced myself it was better not to have this in the fluentd config.  Ideally the user will put the credentials in one of the well-known directories, and failing that they can use the environment variable.  Providing another way to do it seems like an unnecessary complication.

laramiel: Yeah, I went down that path, and ran into a problem.

Specifically, the credentials needed for my next piece are not a part of default application credentials. So users must install their own.

So, I can either reuse the default application credentials, or I can provide a config for them.

Overriding default application credentials seems a bit obscure and error-prone. Ok, so users can download credentials and store them in a separate path. Then I have to provide a method to change the appropriate environment variables. That is essentially equivalent to the config, however it is awkward to accomplish when starting a service. Documentation and explanation for this case would become quite complex.

Suppose, also, that the user wants to run my new config and logging. Now they may need to provide two different sets of credentials. Yet there is only a single mechanism to provide credentials. :/

I take issue with the design of google default application credentials library; essentially it requires all parameters passed as env variables, and it does not provide good mechanisms to get the credentials in other ways. But that is neithere here nor there.

.rubocop_todo.yml
laramiel: Ok, sure.

jkohen:
@qingling128 can you please triage this PR?

qingling128:
This is quite outdated. We've refactored the code multiple times. I'm gonna close this one given the inactivity. Let us know if it's still needed.

PR #63, opened 2016-04-11
Update to rubocop 0.35

mr-salty:
* Fix newly introduced warnings (extra parens, variable case)
* also remove a renundant required_ruby_version that crept in.

### DIFF
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -13,13 +13,11 @@ eos
   gem.version       = '0.4.16'
   gem.authors       = ['Todd Derr', 'Alex Robinson']
   gem.email         = ['salty@google.com']
-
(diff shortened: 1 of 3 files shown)


mr-salty:
@laramiel PTAL.  For some reason I started running into https://github.com/bbatsov/rubocop/issues/2218 on my local machine (which I hadn't seen before), so I upgraded to 0.35 to fix it.  I originally went down the rathole of upgrading to the latest (0.39) but then I saw your PR which would make a gigantic mess so I reverted that.

laramiel:
LGTM.

laramiel:
Yeah, I see why 0.39 is not yet useful. For example, it now prefers `raise` to `fail`, etc.
I'll update to 0.39 once my next change is done.

mr-salty:
yeah, the biggest one is it wants 'freeze' on all constants, and we have a lot.

PR #65, opened 2016-04-21
Detect JSON by checking each line, in all ingested logs

bokowski:
Instead of just checking the first line of each batch, check every line
for structured content in JSON format, for all ingested logs and not
just for logs coming from containers. This makes the logic easier to
understand and explain to users, and more uniform across the different
logs sources. The implementation now checks whether the start of the
line is valid JSON using simple string operations. Overall, assuming
that most batches of log lines are not JSON-formatted and have a
relatively low number of lines in the batch, this change should result
in less CPU compared to the old implementation which calls JSON.parse
and uses rescue to recover.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -333,7 +333,6 @@ module Fluent
             end
           end
         end
-        is_container_json = nil
         arr.each do |time, record|
           next unless record.is_a?(Hash)
 
@@ -359,25 +358,20 @@ module Fluent
             if record.key?('kubernetes')
               handle_container_metadata(record, entry)
             end
-            # If the log from the user container is json, we want to export it
-            # as a structured log. Now that we've pulled out all the
-            # container-specific metadata from the record, we can replace the
-            # record with the json that the user logged.
-            # To save CPU in the common case of unstructured logs, only check if
-            # the contents are parsable as json for the first entry of each
-            # batch.
-            if is_container_json.nil? && record.key?('log')
-              record_json = parse_json_or_nil(record['log'])
-              if record_json.nil?
-                is_container_json = false
-              else
-                record = record_json
-                is_container_json = true
-              end
-            elsif is_container_json && record.key?('log')
-              record_json = parse_json_or_nil(record['log'])
-              record = record_json unless record_json.nil?
-            end
+          end
+
+          # If the log is json, we want to export it as a structured log unless
+          # there is additional metadata that would be lost.
+          is_json = false
+          if record.length == 1 && record.key?('log')
+            record_json = parse_json_or_nil(record['log'])
+          end
+          if record.length == 1 && record.key?('message')
+            record_json = parse_json_or_nil(record['message'])
+          end
+          unless record_json.nil?
+            record = record_json
+            is_json = true
           end
 
           set_timestamp(record, entry, time)
@@ -400,7 +394,7 @@ module Fluent
               @cloudfunctions_log_match['execution_id']
           end
 
-          set_payload(record, entry, is_container_json)
+          set_payload(record, entry, is_json)
           entry.metadata.labels = nil if entry.metadata.labels.empty?
 
           entries.push(entry)
@@ -462,11 +456,23 @@ module Fluent
       # Only here to please rubocop...
       return nil if input.nil?
 
-      begin
-        return JSON.parse(input)
-      rescue JSON::ParserError
-        return nil
-      end
+      input.each_codepoint do |c|
+        if c == 123
+          # left curly bracket (U+007B)
+          begin
+            return JSON.parse(input)
+          rescue JSON::ParserError
+            return nil
+          end
+        else
+          # Break (and return nil) unless the current character is whitespace,
+          # in which case we continue to look for a left curly bracket.
+          # Whitespace as per the JSON spec are: tabulation (U+0009),
+          # line feed (U+000A), carriage return (U+000D), and space (U+0020).
+          break unless c == 9 || c == 10 || c == 13 || c == 32
+        end # case
+      end # do
+      return nil
     end
 
     # "enum" of Platform values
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -359,25 +358,20 @@ def write(chunk)
...
-            elsif is_container_json && record.key?('log')
-              record_json = parse_json_or_nil(record['log'])
-              record = record_json unless record_json.nil?
-            end
+          end
+
+          # If the log is json, we want to export it as a structured log unless
+          # there is additional metadata that would be lost.
+          is_json = false
+          if record.length == 1 && record.key?('log')
```
a-robinson: What's the length check doing?

a-robinson: Have you tested this on real logs? I think record typically has more than just a "log" field at this point, which would just always skip this code.

If it didn't, then the set_timestamp function called right below this would always just be a no-op.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -745,18 +751,20 @@ def field_to_label(record, field, labels, label)
...
 
-    def set_payload(record, entry, is_container_json)
+    def set_payload(record, entry, is_json)
       # Use textPayload if
-      # 1. This is a Cloud Functions log that matched the expected regexp
-      # 2. This is a Cloud Functions log and the 'log' key is available
-      # 3. This is an unstructured Container log and the 'log' key is available
-      # 4. The only remaining key is 'message'
-      if @service_name == CLOUDFUNCTIONS_SERVICE && @cloudfunctions_log_match
+      # 1. The log was not valid JSON
```
a-robinson: I think you mean for the comment to be:

# Use text payload if the log was not valid JSON and:
# 1. [...]
etc.

Since that's effectively being AND'ed with all the other optional conditions.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -462,11 +456,23 @@ def parse_json_or_nil(input)
...
-      begin
-        return JSON.parse(input)
-      rescue JSON::ParserError
-        return nil
-      end
+      input.each_codepoint do |c|
+        if c == 123
+          # left curly bracket (U+007B)
+          begin
+            return JSON.parse(input)
```
a-robinson: You certainly don't have to do it right now, but I wonder whether we should switch this to the JSON parser that fluentd recently switched to for improved performance:
https://github.com/fluent/fluentd/pull/748

@mr-salty

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -462,11 +456,23 @@ def parse_json_or_nil(input)
...
+          rescue JSON::ParserError
+            return nil
+          end
+        else
+          # Break (and return nil) unless the current character is whitespace,
+          # in which case we continue to look for a left curly bracket.
+          # Whitespace as per the JSON spec are: tabulation (U+0009),
+          # line feed (U+000A), carriage return (U+000D), and space (U+0020).
+          break unless c == 9 || c == 10 || c == 13 || c == 32
```
a-robinson: I assume ruby doesn't have a library method for checking whether a codepoint is whitespace?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -745,18 +751,20 @@ def field_to_label(record, field, labels, label)
...
-      # 4. The only remaining key is 'message'
-      if @service_name == CLOUDFUNCTIONS_SERVICE && @cloudfunctions_log_match
+      # 1. The log was not valid JSON
+      # 2. This is a Cloud Functions log that matched the expected regexp
+      # 3. This is a Cloud Functions log and the 'log' key is available
+      # 4. This is an unstructured Container log and the 'log' key is available
+      # 5. The only remaining key is 'message'
+      if is_json
+        entry.struct_payload = record
+      elsif @service_name == CLOUDFUNCTIONS_SERVICE && @cloudfunctions_log_match
```
a-robinson: Should we really be overriding cloud functions' specific regex matching behavior?

lib/fluent/plugin/out_google_cloud.rb
mr-salty: Yeah, probably not a bad idea to do it... sometime - can you create an issue for it?

lib/fluent/plugin/out_google_cloud.rb
a-robinson: For example, Kubernetes logs will ALWAYS also have a "stream" field, so this condition will never trigger for k8s logs if I'm understanding it correctly.

a-robinson: Ah, nevermind about that last comment, I see we're handling the stream field above this.

bokowski: My understanding was that in the normal case, additional fields (like "stream") will have been handled and removed already. The purpose of the length check is to prevent clobbering logs that already come in in some kind of structured format, and only handling cases where the remaining data is one line of text. I've noticed that individual text lines from log files sometimes come in a field "log" and sometimes "message" which is why I am using two checks here.
Any suggestions for how I can add a comment to make this clearer?

lib/fluent/plugin/out_google_cloud.rb
bokowski: Thanks!

lib/fluent/plugin/out_google_cloud.rb
bokowski: This has been created now: #67

lib/fluent/plugin/out_google_cloud.rb
bokowski: I couldn't find anything other than the regular expression /\s/ which I wanted to avoid because of a fear (perhaps irrational?) of this being less performant than the explicit check. The check also nicely corresponds with what's in the JSON spec.

a-robinson: Great, thanks for checking!

lib/fluent/plugin/out_google_cloud.rb
a-robinson: Again, can we avoid overriding the behavior here?

cc @swalkowski, who wrote this logic.

lib/fluent/plugin/out_google_cloud.rb
a-robinson: So I believe that this simply won't work for k8s logs as currently written because real k8s logs also have a "time" field in them that'll still be around at this point. Try rebasing on top of #68 and see if the tests still pass.

lib/fluent/plugin/out_google_cloud.rb
bokowski: Sorry for not responding yesterday (I wish GitHub didn't publish individual comments right away). This is a good catch, let me update the code.

lib/fluent/plugin/out_google_cloud.rb
bokowski: Thanks for pointing this out. Have you had a look at set_timestamp(record, entry, time)? It contains several heuristics for parsing timestamps, shouldn't k8s timestamps be handled (and consumed) there?

bokowski: I've created a new pull request for parsing the "time" field.

a-robinson:
@bokowski - I believe this is waiting on you to update the code, as you mentioned in your comment, but let me know if I'm mistaken and this has been blocked on me.

a-robinson:
@igorpeshansky - if you're alright with merging this soon, I'm happy to copy it to a PR of my own with fixes to the remaining comments applied on top

a-robinson:
@bokowski - any plans to resolve the last couple comments on here, or should I copy this over to a PR of my own?

lib/fluent/plugin/out_google_cloud.rb
steren: It is #69

bokowski:
I *think* once the build problems are resolved, this should work. I have not tested how the timestamp processing from PR #69 interacts with this change but I hope that there isn't anything major that needs to be done to this PR after #69 is merged.

bokowski:
I'd be happy for others to take over, btw.

steren:
It looks like the tests are not passing, for real this time. As mentioned elsewhere, it's probably preferable for somebody else than Boris to take over at this point.

a-robinson:
This can be closed now the #77 is in

igorpeshansky:
Merged via #77.

PR #70, opened 2016-04-27
Use the stream coming from k8s/Docker to determine severity.

bokowski:
Currently, logs from Docker/k8s end up with default severity.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -647,6 +647,16 @@ module Fluent
       elsif record.key?('severity')
         entry.metadata.severity = parse_severity(record['severity'])
         record.delete('severity')
+      elsif @service_name == CONTAINER_SERVICE && \
+            entry.metadata.labels.key?("#{CONTAINER_SERVICE}/stream")
+        stream = entry.metadata.labels["#{CONTAINER_SERVICE}/stream"]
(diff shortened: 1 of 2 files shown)


a-robinson:
@igorpeshansky - I'm not convinced this is a good change. I'd prefer to chat about this with @bokowski before moving forward.

bokowski:
@a-robinson What are your concerns? Do you disagree that log entries should have a severity of info/error depending on whether they come from stdout/stderr, or did you want to discuss the particular way I ended up implementing this?

a-robinson:
Eh, sure, I'll go along with it. My concern is about whether the stdout/stderr distinction is semantically close enough to an info/error severity distinction to apply it to all logs. Given the lack of log viewer support for showing different fields for different service's logs, though, this will probably be valuable enough for long enough to be worth the potential confusion to some users. I've at least seen a couple people express mild interest in this. 

But could you include a test before we merge this, though?

Also, for future reference, I'd recommend using a separate branch for independent changes :) The Kubernetes repo has [a doc on the basic git workflow](https://github.com/kubernetes/kubernetes/blob/master/docs/devel/development.md#workflow), or you should be able to find tons of tutorials with a quick search.

steren:
Boris having changed team, we should not count more on his availability. @a-robinson would it be something you can help?

igorpeshansky:
Merged via #76.

PR #69, opened 2016-04-27
Parse timestamps that k8s puts in a field 'time'

bokowski:
Current behavior is to use the ingestion time; the 'time' value is not used.

PR #77, opened 2016-06-14
Detect JSON by checking each line, in all ingested logs

a-robinson:
This takes @bokowski's #65, resolves the merge conflict with the other PRs that have gone in recently, and fixes the broken tests from #65 by removing the time field from the record while checking for json. That's needed because we don't attempt to parse json if the record itself has multiple fields in it. The time is re-added after the json parsing if needed, as implemented by #69.

@igorpeshansky - I believe this is good to go, with just the one line change on top of #65.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -375,8 +375,10 @@ module Fluent
             end
           end
 
-          # Save the timestamp if available
+          # Save the timestamp if available, then clear it out to allow for
+          # determining whether we should parse the log or message field.
           timestamp = record.key?('time') ? record['time'] : nil
+          record.delete('time')
           # If the log is json, we want to export it as a structured log unless
           # there is additional metadata that would be lost.
           is_json = false
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -348,7 +348,6 @@ module Fluent
             end
           end
         end
-        is_container_json = nil
         arr.each do |time, record|
           next unless record.is_a?(Hash)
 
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -374,31 +373,28 @@ def write(chunk)
...
+
+          # Save the timestamp if available, then clear it out to allow for
+          # determining whether we should parse the log or message field.
+          timestamp = record.key?('time') ? record['time'] : nil
+          record.delete('time')
+          # If the log is json, we want to export it as a structured log unless
+          # there is additional metadata that would be lost.
+          is_json = false
+          if record.length == 1 && record.key?('log')
+            record_json = parse_json_or_nil(record['log'])
```
igorpeshansky: @a-robinson @rbuskens @mr-salty I'm a bit concerned about this now being done unconditionally, outside of the CONTAINER_SERVICE branch. We already have complaints about output plugin performance, and this is likely to exacerbate that problem... Thoughts?

bokowski: I would expect the performance to be better for the normal car (log steams
not containing JSON), because of the simple string prefix check for an
opening curly brace (instead of relying on catching exceptions), and not
worse for those log streams containing only JSON. The only expected
slowdown would be in cases where log lines frequently start with a curly
brace but do not contain valid JSON.

igorpeshansky: The point is that this code, efficient as it may be, is now called unconditionally, while before it was only called when running in a container. I don't know how much effect this will have on the overall performance.

a-robinson: It's cheap as long as the log body isn't json. And if the body is json, then there's real value to be had in having it parsed into a structured payload vs exported as a text payload.

igorpeshansky: Ok, I'll buy that.

igorpeshansky:
LGTM :shipit:
Is this ready to merge?

a-robinson:
I believe so, yes.

lib/fluent/plugin/out_google_cloud.rb
mr-salty: So, this means that any message that looks like json will be parsed as json?  I'm not sure that's actually desirable.  The user already has the ability to specify "format json" in the input config file if that's what they wanted, whereas this happens by default with no way to disable it.

igorpeshansky: @a-robinson @rbuskens Hmm, good point. I didn't think of that. Should we back this out and move it back behind the container guard?

bokowski: My hope was to handle JSON-shaped text not just in the container case. Specifically, in the case of App Engine Flex, because the user code runs in a docker container, logs are text-shaped and currently there isn't any way for users to modify the fluentd configuration to specify the "format json" option.

mr-salty: I don't think our output plugin should arbitrarily parse JSON; it's ok to do it in special cases but I think it's surprising and potentially undesirable behavior in the general case.

FWIW, at some point along the line I did start to wonder if we were doing the right thing by baking all of this type of functionality into our output plugin, or whether we should be using [filter plugins](http://docs.fluentd.org/articles/filter-plugin-overview) or something else. Then we can compose and configure the proper chain of filters for a given environment, while making the output plugin a more generic "just send what you get to cloud logging".  I never had the time to sit down and think it through, but it might be worth considering.

bokowski: I disagree: I believe encouraging/nudging users towards using structured
logs data is a good thing, and would see parsing JSON as a key way to do
that. It's a benign automatism because it doesn't lead to data loss, and
because it is easy to understand what happens. Not parsing JSON means that
you ask users to perform an action (changing configuration), which makes it
less likely that logs data becomes structured data.

igorpeshansky: Ok, we've decided for expediency's sake to move this back under the guard (#78). We can always move it back out later.

PR #82, opened 2016-08-23
Migrate to Logging API v2beta1

mr-salty:
* Migrate to Logging API v2beta1
  * MonitoredResource now specifies the service and keys, rather than using labels
  * Update all "hardcoded" services to use the new schema
  * Add some special-case code for Dataflow and ML which use labels/label_map config
  * Update tests; add tests for dataflow and ML


mr-salty:
I think that takes care of all of the open issues, unless you want to pull anything from grpc into master and merge it here first.

PR #83, opened 2016-09-08
Stackdriver product rename

tmatsuo:
Renaming some APIs to Stackdriver brand


igorpeshansky:
These look fine, but the main question is whether the plugin itself should be renamed to `fluent-plugin-stackdriver`...

tmatsuo:
Yeah it makes sense to change the name of the plugin (and potentially the name of the repo).
I'd appreciate it if you could take care of plugin naming change outside of this PR.

PR #90, opened 2016-10-21
Enable JSON parsing for App Engine

### LINKED ISSUE #89: Enable JSON parsing for App Engine

meltsufin:
Fixes #89.

I didn't update the tests. I'm trying to get feedback if this is an acceptable change.

@igorpeshansky PTAL

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -423,6 +423,9 @@ module Fluent
             @cloudfunctions_log_match =
               @cloudfunctions_log_regexp.match(record['log'])
           end
+
+          set_timestamp(resource, record, entry, time)
+
           if resource.type == CONTAINER_RESOURCE_TYPE ||
              resource.type == APPENGINE_RESOURCE_TYPE
             # Move the stdout/stderr annotation from the record into a label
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -440,6 +441,10 @@ def write(chunk)
             # determining whether we should parse the log or message field.
             timestamp = record.key?('time') ? record['time'] : nil
             record.delete('time')
+            if timestamp.nil?
+              timestamp = record.key?('timestamp') ? record['timestamp'] : nil
+              record.delete('timestamp')
```
igorpeshansky: Careful, this will break the code that restores the `'time'` field in line 462 (which you'd probably want to do with the `'timestamp'` field as well). I'd suggest having two separate variables for these, `time_value` and `timestamp_value`.

meltsufin:
@igorpeshansky  I found that a different way to make this work was to just call `set_timestamp` earlier.
The issue here is that the [from_docker](https://github.com/GoogleCloudPlatform/appengine-sidecars-docker/blob/master/fluentd_logger/out_from_docker.rb) plugin already does some transformation of the docker log, yet in this plugin some of the work is duplicated. Ultimately, I think it should just be handled in one place. 

Looping in @dlorenc.

igorpeshansky:
Looks like calling `set_timestamp` earlier will break in cases where the parsed record has its own timestamp (because your call to `set_timestamp` will use the old record, not the one parsed from the JSON). It's possible that this is the intended behavior, but the potential interactions are quite hairy. Without extensive comments that document the desired semantics, I'm very reluctant to change the order of these calls.

dlorenc:
cc @andrewsg

meltsufin:
@igorpeshansky Agreed, we need a clear definition of what are the supported input formats for this plugin. Otherwise, we don't know what we might be breaking.

PR #92, opened 2016-11-04
Fix UTF-8 encoding issue with gRPC support

qingling128:
* b/32474926 gRPC plugin needs to handle invalid UTF-8 chracters

### DIFF
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -10,7 +10,7 @@ eos
   gem.homepage      = \
     'https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud'
   gem.license       = 'Apache-2.0'
-  gem.version       = '0.5.3.grpc.alpha.1'
+  gem.version       = '0.5.3.grpc.alpha.3'
   gem.authors       = ['Todd Derr', 'Alex Robinson']
   gem.email         = ['salty@google.com']
   gem.required_ruby_version = Gem::Requirement.new('>= 2.0')
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -122,6 +122,16 @@ module Fluent
     # Cloud Logging API.
     config_param :use_grpc, :bool, :default => false
 
+    # Whether to allow non-UTF-8 characters in user logs. If set to true, any
+    # non-UTF-8 character would be replaced by the string specified by
+    # 'non_utf8_replacement_string'. If set to false, any non-UTF-8 character
+    # would trigger the plugin to error out.
+    config_param :coerce_to_utf8, :bool, :default => true
+
+    # If 'coerce_to_utf8' is set to true, any non-UTF-8 character would be
+    # replaced by the string specified here.
+    config_param :non_utf8_replacement_string, :string, :default => ' '
+
     # DEPRECATED: The following parameters, if present in the config
     # indicate that the plugin configuration must be updated.
     config_param :auth_method, :string, :default => nil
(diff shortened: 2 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1415,5 +1416,14 @@ def api_client
       end
       @client
     end
+
+    # Encode by UTF-8 with invalid characters replaced by spaces.
+    def coerce_to_utf8_encode(input)
```
qingling128: Looking at how the old JSON approach handled invalid UTF-8 characters at the code Igor pointed out: https://cs.corp.google.com/piper///depot/google3/net/proto2/util/converter/internal/json_stream_parser.cc?l=152&rcl=133771901, it's throwing away invalid characters and replacing them with spaces. So here I'm trying to mock the behavior. Note that in that code base, we are also using a boolean flag to control whether to allow non UTF-8 encoded input (aka, whether to replace them or error out). Do we need a similar flag here?

igorpeshansky: Sounds like a good idea to me, in case users prefer to crash rather than ingest invalid logs. But, just like the API, we should make this behaviour the default.

fluent-plugin-google-cloud.gemspec
```diff
@@ -10,7 +10,7 @@ eos
   gem.homepage      = \
     'https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud'
   gem.license       = 'Apache-2.0'
-  gem.version       = '0.5.3.grpc.alpha.1'
+  gem.version       = '0.5.3.grpc.alpha.3'
```
qingling128: Seems it was left as `.1`, so I updated it.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -122,6 +122,11 @@ class GoogleCloudOutput < BufferedOutput
     # Cloud Logging API.
     config_param :use_grpc, :bool, :default => false
 
+    # Whether to allow non UTF-8 characters. If set to true, any non UTF-8
+    # character would be replaced by a space. If set to false, any non UTF-8
+    # character would trigger the plugin to error out.
+    config_param :allow_non_utf8, :bool, :default => true
```
qingling128: Done

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1415,5 +1421,26 @@ def api_client
...
+          undef: :replace,
+          replace: ' ')
+      else
+        begin
+          input.encode('utf-8')
+        rescue EncodingError
+          @log.error 'Encountered encoding issues potentially due to non ' \
+                     'UTF-8 characters. To allow non UTF-8 characters and ' \
+                     'replace them with spaces, please set "allow_non_utf8" ' \
+		     'to true.'
```
qingling128: Oops. I changed the wording last minute, and did not run the test again. Lesson learnt.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -122,6 +122,16 @@ class GoogleCloudOutput < BufferedOutput
     # Cloud Logging API.
     config_param :use_grpc, :bool, :default => false
 
+    # Whether to allow non-UTF-8 characters in user logs. If set to true, any
+    # non-UTF-8 character would be replaced by the string specified by
+    # 'non_utf8_replacement'. If set to false, any non-UTF-8 character would
```
rbuskens: 'non_utf8_replacement' --> 'non_utf8_replacement_string'?

qingling128: Changed for all 4 occurences. And I just built and tested the plugin on a VM again.

rbuskens:
LGTM.

dhrupadb:
LGTM

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: FWIW, this is non-Rubyesque. It's preferred to leave the expected type (if any) to the documentation, rather than embed it into the name. That way the functionality can be extended later. Let's please fix this before customers start depending on this behavior.

igorpeshansky:
One quick follow-up on the procedure: can we please agree to push individual commits during the review, and squash them on the final merge? Makes it much easier to see the incremental changes.

PR #73, opened 2016-06-10
Pin fluentd at 0.12 until we figure out how to make 0.14 work.

igorpeshansky:

PR #94, opened 2016-11-08
Update Fluentd version to latest

dhrupadb:
Removed Fluentd version pin. Fixed test which now supports nanosecond timestamp log message verification. Updated function call to remove variable name declaration for use_v1_config (no longer seems to be supported). This is effectively a revert of #73.

### DIFF
--- a/test/plugin/test_out_google_cloud.rb
+++ b/test/plugin/test_out_google_cloud.rb
@@ -324,7 +324,7 @@ class GoogleCloudOutputTest < Test::Unit::TestCase
 
   def create_driver(conf = APPLICATION_DEFAULT_CONFIG, tag = 'test')
     Fluent::Test::BufferedOutputTestDriver.new(
-      Fluent::GoogleCloudOutput, tag).configure(conf, use_v1_config: true)
+      Fluent::GoogleCloudOutput, tag).configure(conf, true)
   end
 
   def test_configure_service_account_application_default
@@ -638,9 +638,7 @@ class GoogleCloudOutputTest < Test::Unit::TestCase
     [Time.at(123_456.789), Time.at(0), Time.now].each do |ts|
       # Test the "native" fluentd timestamp as well as our nanosecond tags.
       d.emit({ 'message' => log_entry(emit_index) }, ts.to_f)
-      # The native timestamp currently only supports second granularity
-      # (fluentd issue #461), so strip nanoseconds from the expected value.
-      expected_ts.push(Time.at(ts.tv_sec))
+      expected_ts.push(ts)
       emit_index += 1
       d.emit('message' => log_entry(emit_index),
              'timeNanos' => ts.tv_sec * 1_000_000_000 + ts.tv_nsec)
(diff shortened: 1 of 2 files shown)


dhrupadb:
Seems to be a fluentd issue where the newer version requires a higher version of serverengine (another Treasure Data package). Ruby now has version 2.3 as well. Should we continue to support 2.0.0? Ruby has officially stopped maintaining it as of December last year.

test/plugin/test_out_google_cloud.rb
```diff
@@ -742,9 +742,8 @@ def test_timestamps
     [Time.at(123_456.789), Time.at(0), Time.now].each do |ts|
       # Test the "native" fluentd timestamp as well as our nanosecond tags.
       d.emit({ 'message' => log_entry(emit_index) }, ts.to_f)
-      # The native timestamp currently only supports second granularity
-      # (fluentd issue #461), so strip nanoseconds from the expected value.
-      expected_ts.push(Time.at(ts.tv_sec))
+
```
igorpeshansky: No need for this blank line.

test/plugin/test_out_google_cloud.rb
```diff
@@ -428,7 +428,7 @@ def setup
 
   def create_driver(conf = APPLICATION_DEFAULT_CONFIG, tag = 'test')
     Fluent::Test::BufferedOutputTestDriver.new(
-      Fluent::GoogleCloudOutput, tag).configure(conf, use_v1_config: true)
```
igorpeshansky: This named parameter syntax ought to work. Can we please investigate why it needed to be changed?

test/plugin/test_out_google_cloud.rb
```diff
@@ -742,9 +742,8 @@ def test_timestamps
     [Time.at(123_456.789), Time.at(0), Time.now].each do |ts|
       # Test the "native" fluentd timestamp as well as our nanosecond tags.
       d.emit({ 'message' => log_entry(emit_index) }, ts.to_f)
-      # The native timestamp currently only supports second granularity
-      # (fluentd issue #461), so strip nanoseconds from the expected value.
-      expected_ts.push(Time.at(ts.tv_sec))
+
+      expected_ts.push(Time.at(ts))
```
igorpeshansky: This is just `expected_ts.push(ts)`. Once you do this, it shouldn't be flaky, but please double-check.

test/plugin/test_out_google_cloud.rb
dhrupadb: This is where the .configre function is defined in Fluentd : https://github.com/fluent/fluentd/blob/335d3b2b709ad368da6655636e8ab401b116ae4b/lib/fluent/test/base.rb#L51. Explicitly specifying the variable throws a RuboCop error. The error is `Useless assignment to variable`

test/plugin/test_out_google_cloud.rb
dhrupadb: The test is still flaky. There's about a 500 odd nanosecond difference between the two timestamps. I'd imagine this is expected as even between the point when the log is generated and the timestamp for logged message is picked up there would be a tiny delay?

test/plugin/test_out_google_cloud.rb
igorpeshansky: There are two issues here. One is that the parameter is actually named `use_v1`, and the other is that Ruby, unlike Python, does not support named invocation syntax for non-named parameters. So what you did is exactly correct.

test/plugin/test_out_google_cloud.rb
igorpeshansky: As discussed offline, this is actually a rounding error in the `to_f` call (we're missing 9 bits of precision, which amounts to ~512 nanoseconds). Making this something over 512 should work.

test/plugin/test_out_google_cloud.rb
```diff
@@ -764,8 +763,10 @@ def test_timestamps
     verify_log_entries(emit_index, COMPUTE_PARAMS) do |entry|
       assert_equal expected_ts[verify_index].tv_sec,
                    entry['timestamp']['seconds'], entry
-      assert_equal expected_ts[verify_index].tv_nsec,
-                   entry['timestamp']['nanos'], entry
+      # Fluentd v0.14 onwards supports nanosecond timestamp values.
+      # Added in 1 ms delta to avoid flaky tests.
```
igorpeshansky: I would add the above explanation here and reference [the Wikipedia article](https://en.wikipedia.org/wiki/Double-precision_floating-point_format).

dhrupadb:
PTAL

igorpeshansky:
LGTM :shipit:

PR #100, opened 2016-12-14
fix referer field handling

qingling128:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1160,26 +1160,27 @@ def set_http_request_grpc(record, entry)
       input = record['httpRequest']
       output = Google::Logging::Type::HttpRequest.new
       output.request_method = input.delete('requestMethod') if
-        input.key?('requestMethod')
+        input.key?('requestMethod') && !input['requestMethod'].nil?
```
igorpeshansky: Without extra parameters on construction, the default return value of a Hash lookup is `nil`. So this is equivalent to: `if !input['request_method'].nil?` (i.e., the `key?` call is redundant). Also everywhere below.

test/plugin/base_test.rb
```diff
@@ -1025,17 +1025,26 @@ def test_http_request_partial_from_record
...
-    verify_log_entries(1, COMPUTE_PARAMS, 'httpRequest') do |entry|
-      assert_equal http_request_message_without_referer, entry['httpRequest'],
-                   entry
-      assert_nil get_fields(entry['structPayload'])['httpRequest'], entry
+    {
+      http_request_message_with_referer_nil => \
+        http_request_message_expected_struct_payload,
+      http_request_message_with_referer_absent => 'null'
+    }.each do |message, expected|
+      @logs_sent = []
```
igorpeshansky: Should we just turn these into two separate tests? You could even move the absent referer function into the grpc test, as that distinction only matters there.

test/plugin/test_out_google_cloud.rb
```diff
@@ -240,10 +244,16 @@ def http_request_message
     HTTP_REQUEST_MESSAGE
   end
 
-  # A wrapper around the constant HTTP_REQUEST_MESSAGE_WITHOUT_REFERER, so the
-  # definition can be skipped in the shared module and defined here.
-  def http_request_message_without_referer
-    HTTP_REQUEST_MESSAGE_WITHOUT_REFERER
+  # A wrapper around the constant HTTP_REQUEST_MESSAGE_EXPECTED_WITHOUT_REFERER,
+  # so the definition can be skipped in the shared module and defined here.
+  def http_request_message_expected_without_referer
```
igorpeshansky: I think we've reached an unwarranted level of complexity here. Let's do this: keep only `http_request_message` as a virtual function (with the corresponding constant), add a non-virtual `http_request_message_with_nil_referer` that just returns `http_request_message.merge('referer' => nil)`, and move the special handling of the absent referer into the grpc test. In fact, you could even move the `HTTP_REQUEST_MESSAGE` constant into the base class, and override `http_request_message` in this file to rename the key.

test/plugin/base_test.rb
```diff
@@ -1025,17 +1025,26 @@ def test_http_request_partial_from_record
...
+      setup_logging_stubs do
+        d = create_driver
+        d.emit('httpRequest' => message)
+        d.run
+      end
+      verify_log_entries(1, COMPUTE_PARAMS, 'httpRequest') do |entry|
+        assert_equal http_request_message_expected_without_referer,
+                     entry['httpRequest'], entry
+        assert_equal expected,
+                     get_fields(entry['structPayload'])['httpRequest'].to_json,
```
igorpeshansky: Sorry, are we sure this is the right behavior? Should we really be leaving a struct with a `null` referer in this case? Or should we be deleting the referer and ending up with a `nil` struct?

test/plugin/base_test.rb
```diff
@@ -1275,10 +1296,17 @@ def http_request_message
...
+  # so the definition can be skipped in the shared module and defined in the
+  # test classes later.
+  def http_request_message_expected_without_referer
+    _undefined
+  end
+
+  # A wrapper around the constant HTTP_REQUEST_MESSAGE_EXPECTED_STRUCT_PAYLOAD,
+  # so the definition can be skipped in the shared module and defined in the
+  # test classes later.
+  def http_request_message_expected_struct_payload
```
igorpeshansky: Instead of this, why don't we define `assert_null_struct` as a virtual method? But see the other comment about whether this is the right behavior in the first place.

test/plugin/base_test.rb
qingling128: Right now when we get a payload like:

```
{"httpRequest"=>{"referer"=>nil}}
```

The constructed proto is:

```
<Google::Protobuf::Struct: fields: {"httpRequest"=><Google::Protobuf::Value: null_value: :NULL_VALUE, number_value: 0.0, string_value: "", bool_value: false, struct_value: <Google::Protobuf::Struct: fields: {"referer"=><Google::Protobuf::Value: null_value: :NULL_VALUE, number_value: 0.0, string_value: "", bool_value: false, struct_value: nil, list_value: nil>}>, list_value: nil>}>
```

I found it weird at first sight too. Another option is to construct something like:
```
<Google::Protobuf::Struct: fields: {"httpRequest"=><Google::Protobuf::Value: null_value: :NULL_VALUE, number_value: 0.0, string_value: "", bool_value: false, struct_value: nil, list_value: nil>}>
```

But that means the following payloads would be equivalent:

```
{"httpRequest"=>{"referer"=>nil}}
{"httpRequest"=>nil}
```

Which is probably fine, I guess? But I did not see the necessity to add extra logic to behave it this way. Is there?

lib/fluent/plugin/out_google_cloud.rb
qingling128: Good to know. This has been changed in all field checks.

test/plugin/base_test.rb
qingling128: Yeah, it's too complicated this way. The test has been split into two.

test/plugin/base_test.rb
qingling128: Not very sure of the right behavior here yet. Will submit another patch if we figure it should behavior differently.

test/plugin/base_test.rb
qingling128: This method has been removed during the split of the test.

test/plugin/test_out_google_cloud.rb
qingling128: Refactoring has been done to implement the default `http_request_message` method in `base_test`, and let the grpc test override them. And a non-virtual `http_request_message_with_nil_referer` has been added.

test/plugin/base_test.rb
```diff
@@ -307,6 +307,19 @@ def setup
...
+    'requestMethod' => 'POST',
+    'requestUrl' => 'http://example/',
+    'requestSize' => 210,
+    'status' => 200,
+    'responseSize' => 65,
+    'userAgent' => 'USER AGENT 1.0',
+    'remoteIp' => '55.55.55.55',
+    'referer' => 'http://referer/',
+    'cacheHit' => false,
+    'validatedWithOriginServer' => true
```
igorpeshansky: The current official name is `'cacheValidatedWithOriginServer'`, isn't it? Let's keep that name in the base class, and rename in the v1 tests, instead of having to accommodate this in the gRPC test.

To be clear, here's what I'm suggesting:
```
module BaseTest
  HTTP_REQUEST_MESSAGE = {
    'requestMethod' => 'POST',
    'requestUrl' => 'http://example/',
    'requestSize' => 210,
    'status' => 200,
    'responseSize' => 65,
    'userAgent' => 'USER AGENT 1.0',
    'remoteIp' => '55.55.55.55',
    'referer' => 'http://referer/',
    'cacheHit' => false,
    'cacheValidatedWithOriginServer' => true
  }
  def http_request_message
    HTTP_REQUEST_MESSAGE
  end
end
class GoogleCloudOutputTest < Test::Unit::TestCase
  include BaseTest
  def http_request_message
    super.merge('validatedWithOriginServer' => super['cacheValidatedWithOriginServer'])
      .reject do |k, _| k == 'cacheValidatedWithOriginServer' end
  end
end
```
(You could even define a `rename_key` helper function to do the above:
```
def rename_key(hash, old_key, new_key)
  hash.merge(new_key => hash[old_key]).reject do |k, _| k == old_key end
end
```
and then the above override of `http_request_message` becomes:
```
def http_request_message
  rename_key(super, 'validatedWithOriginServer', 'cacheValidatedWithOriginServer')
end
```
).

test/plugin/base_test.rb
igorpeshansky: I still see one test in `test_out_google_cloud_grpc.py`. Did you decide against splitting it?
Also, this implementation in `BaseTest` is invalid -- do you even need it, now that you provide one in both tests anyway?

qingling128: My bad. I forgot to delete the one in the base test. It was definitely invalid. The test did not error out because both tests override it. I've just deleted it.

test/plugin/base_test.rb
qingling128: Ah, yeah, consistency is indeed a big concern. To ensure backward compatibility, it's better to make the non-grpc path behave the same way. I've made changes for this. There are some exceptions I need to make to make things consistent, which feels a bit hacky though...

test/plugin/base_test.rb

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -72,6 +72,26 @@ def test_server_error
     end
   end
 
+  def test_http_request_from_record_with_referer_nil_or_absent
```
igorpeshansky: Let's actually split this into two tests, `test_http_request_from_record_with_referer_nil` and `test_http_request_from_record_with_referer_absent`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1368,11 +1041,16 @@ def list_from_ruby(arr)
...
+    # {"httpRequest"=>{"referer"=>nil}} instead of layers of empty structs.
+    def struct_from_ruby(hash, reduce = true)
       ret = Google::Protobuf::Struct.new
       hash.each do |k, v|
-        ret.fields[k] ||= value_from_ruby(v)
+        ret.fields[k] ||= value_from_ruby(v) \
+          if !v.nil? && !value_from_ruby(v).nil?
       end
+
+      ret = nil if reduce && ret.fields.count { |_, v| !v.nil? } == 0
```
igorpeshansky: Same here, let's change this to `return nil` from `ret = nil`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1348,7 +1016,12 @@ def value_from_ruby(value)
       when Google::Protobuf::Struct
         ret.struct_value = value
       when Hash
-        ret.struct_value = struct_from_ruby(value)
+        struct = struct_from_ruby(value)
```
qingling128: Changed the param to be required instead of optional.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1348,7 +1016,12 @@ def value_from_ruby(value)
...
+        if struct.nil?
+          ret = nil
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1368,11 +1041,16 @@ def list_from_ruby(arr)
       ret
     end
 
-    def struct_from_ruby(hash)
+    # If reduce is true, nil will be returned for hash like
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1368,11 +1041,16 @@ def list_from_ruby(arr)
...
+    # {"httpRequest"=>{"referer"=>nil}} instead of layers of empty structs.
+    def struct_from_ruby(hash, reduce = true)
```

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1389,13 +1067,23 @@ def set_payload_grpc(record, entry, is_json)
       elsif @service_name == CLOUDFUNCTIONS_SERVICE && record.key?('log')
         entry.text_payload = convert_to_utf8(record['log'])
       elsif is_json
-        entry.struct_payload = struct_from_ruby(record)
+        # In the non-grpc path, "struct_payload={}" is returned when there is no
```
qingling128: The comment has been changed. And the argument has been changed to be required.

test/plugin/test_out_google_cloud.rb
```diff
@@ -70,6 +70,20 @@ def test_server_error
     assert_equal 1, exception_count
   end
 
+  def test_http_request_from_record_with_referer_nil_or_absent
```
qingling128: Nice catch. Changed.

test/plugin/test_out_google_cloud.rb
```diff
@@ -175,25 +189,16 @@ def test_parse_severity
 
   private
 
+  def rename_key(hash, old_key, new_key)
+    hash.merge(new_key => hash[old_key]).reject { |k, _| k == old_key }
+  end
+
   # The non-grpc path has a unique field 'validatedWithOriginServer', while
   # the grpc path has a unique field 'cacheValidatedWithOriginServer'.
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1368,11 +1041,16 @@ def list_from_ruby(arr)
...
 
-    def struct_from_ruby(hash)
+    # If reduce is true, nil will be returned for hash like
+    # {"httpRequest"=>{"referer"=>nil}} instead of layers of empty structs.
+    def struct_from_ruby(hash, reduce = true)
       ret = Google::Protobuf::Struct.new
       hash.each do |k, v|
-        ret.fields[k] ||= value_from_ruby(v)
+        ret.fields[k] ||= value_from_ruby(v) \
+          if !v.nil? && !value_from_ruby(v).nil?
```
qingling128: This has gone away after the recent change.

qingling128: Did some investigation on how the REST path handles null fields. Turned out my previous implementation did break the existing functionality with null field handling. The previous fix has been reverted, and a new fix that directly applies to the `set_http_request_grpc` method has been included. And some new tests covering the null field handling behavior have been added.

qingling128: Basically, the reason we had layered empty structs in 'httpRequest' is because:
When user specify 'referer=nil', in a `google.protobuf.Struct` it's presented as `{"nullValue"=>"NULL_VALUE"}`. And in the previous implementation, `referer` is not being deleted from `httpRequest` even though it's technically nil:

```
  output.referer = input.delete('referer') unless
    input['referer'].nil?
```

As a result, when we check if `input` is empty, we did not delete `httpRequest` properly:

```
record.delete('httpRequest') if input.empty?
```

test/plugin/base_test.rb
```diff
@@ -1302,6 +1302,16 @@ def get_number(_field)
     _undefined
   end
 
+  # Get the value of a null field.
+  def get_null(_field)
```
igorpeshansky: You shouldn't need to extract the null value. Instead, why not construct the value to be compared with and return that as `null_value`? So, here it would be just `nil`, and in the gRPC path it'd be the hash `{'nullValue'=>'NULL_VALUE'}`. Would that work?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -72,6 +72,36 @@ def test_server_error
     end
   end
 
+  def test_http_request_from_record_with_referer_nil
+    setup_gce_metadata_stubs
+    @logs_sent = []
```
igorpeshansky: Why do you need to reset `@logs_sent` here, but not elsewhere?

test/plugin/base_test.rb

test/plugin/test_out_google_cloud_grpc.rb
qingling128: This is some residue from the test before splitting. Removed.

igorpeshansky:
LGTM :shipit:

PR #107, opened 2017-01-06
Fix malformat hash key handling.

qingling128:


test/plugin/base_test.rb
```diff
@@ -544,6 +544,33 @@ def test_struct_payload_log
     end
   end
 
+  def test_struct_payload_malformat_log
```
igorpeshansky: s/malformat/malformatted/. Also in the PR title.

test/plugin/base_test.rb
```diff
@@ -544,6 +544,33 @@ def test_struct_payload_log
...
+    setup_gce_metadata_stubs
+    message = 'test message'
+    setup_logging_stubs do
+      d = create_driver
+      d.emit(
+        'int_as_key' => { 1 => message },
```
igorpeshansky: I'd just use `'int_key'`, `'array_key'`, `'hash_key'`, `'nil_key'`. Also, you forgot symbol (e.g., `{ :symbol => message }`).

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -131,6 +131,26 @@ def test_severities
     end
   end
 
+  def test_struct_payload_non_utf8_log
+    setup_gce_metadata_stubs
+    setup_logging_stubs do
+      d = create_driver
+      d.emit('msg' => log_entry(0),
+             'normal_key' => 'test' + non_utf8_character + 'some non utf8',
```
igorpeshansky: This is the test we should have added earlier but didn't, right?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -131,6 +131,26 @@ def test_severities
...
+             'non_utf8' + non_utf8_character + 'key' => 5000,
+             'some_null_field' => nil)
```
igorpeshansky: Do we also want to test the case where a non-utf8 character appears nested in a structured payload?

qingling128: Added.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: Yes. Should have added this.

test/plugin/base_test.rb
qingling128: Key names are changed and a test case for symbol is added. Seems `{ symbol: message }` is preferred in Ruby 1.9 though. (got Rubocop errors for `{ :symbol => message }`

test/plugin/base_test.rb
qingling128: Changed in this test. Will change the PR title when merging the PR.

igorpeshansky:
LGTM :shipit:

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1044,7 +1044,7 @@ def list_from_ruby(arr)
     def struct_from_ruby(hash)
       ret = Google::Protobuf::Struct.new
       hash.each do |k, v|
-        ret.fields[k] ||= value_from_ruby(v)
+        ret.fields[convert_to_utf8(k.to_s)] ||= value_from_ruby(v)
```
igorpeshansky: Are we sure the `to_s` method will do the same thing as the REST path (which probably represents these as a JSON string instead)?

igorpeshansky:
Looks like I was too late to add that last comment... Let's make a note that we should at least check it, but proceed with the release. Definitely doesn't block the `0.5.4.grpc.alpha.1` build.

PR #110, opened 2017-01-12
Migrate to Logging API v2.

qingling128:
* Migrate to logging API v2beta1 (https://cloud.google.com/logging/docs/api/v2/migration-to-v2).
* Migrate top-level labels to MonitoredResource labels.
* Add subservice-specific label transformations and tests.


piosz:
@Crassirostris could you please take a look?

Crassirostris:
@piosz Yes, absolutely

@qingling128 Where I can try out this change? Do you release alpha version of the gem?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1049,52 +1139,76 @@ def struct_from_ruby(hash)
...
+        labels_to_extract = %w(region job_name job_id step_id)
+      elsif resource_type == ML_RESOURCE_TYPE
+        label_prefix = ML_SERVICE
+        labels_to_extract = %w(job_id task_name)
+      else
+        return extracted_labels
+      end
+
+      labels_to_extract.each do |label|
+        extracted_labels[label] = labels.delete("#{label_prefix}/#{label}") if
```
igorpeshansky: How will the gRPC code react to `nil` values in these labels (e.g., if the original label was `nil`)? Or will it only be `nil` if the key is not present (covered by your conditional)?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1147,3 +1261,18 @@ def convert_to_utf8(input)
     end
   end
 end
+
+module Google
```
igorpeshansky: This was only needed in the original code because we called `resource.dup`. Unless we go back to that, we shouldn't need this patch.

qingling128:
@Crassirostris We normally release the alpha version after the pull request is approved and merged. I just built a gem from this branch and attached it to the original ticket.

Crassirostris:
@qingling128 Thanks for clarifying and building a gem!

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -435,73 +527,67 @@ def write(chunk)
...
-            entry.metadata.timestamp = {
-              seconds: ts_secs,
-              nanos: ts_nanos
-            }
-
-            entry.metadata.severity =
-              compute_severity(record, entry)
-
+            entry = Google::Apis::LoggingV2beta1::LogEntry.new(
+              labels: entry_common_labels,
```
qingling128: Hmm... Added this bit back.

lib/fluent/plugin/out_google_cloud.rb
qingling128: `labels` would never be `nil` in our case, since we have:

```
  case @platform
      when Platform::GCE
        common_labels["#{COMPUTE_SERVICE}/resource_name"] = @vm_name
      when Platform::EC2
        common_labels["#{EC2_SERVICE}/resource_name"] = @vm_name
      when Platform::OTHER
        common_labels["#{COMPUTE_SERVICE}/resource_name"] = @vm_name
      end
```

But it does not hurt to add an extra check to return `{}` if `labels` happens to be `nil` or not a hash. Just added.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Removed this patch.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -344,74 +486,24 @@ def write(chunk)
...
-              labels["#{CONTAINER_SERVICE}/#{field}"] = match_data[field]
-            end
-          end
-        end
+        group_resource_type, group_resource_labels, group_common_labels =
+          compute_group_type_and_labels(tag)
+
+        # freeze the per-request state.  Any further changes must be made
+        # on a per-entry basis.
+        group_resource_type.freeze
```
qingling128: Moved this to `compute_group_type_and_labels`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,6 +377,105 @@ def format(tag, time, record)
       [tag, time, record].to_msgpack
     end
 
+    def compute_group_type_and_labels(tag)
+      # Note that we assume that labels added to common_labels below are not
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -282,41 +302,64 @@ def configure(conf)
...
+        extracted_labels.nil? || extracted_labels.empty?
+
+      # The resource and labels are now set up; ensure they can't be modified
+      # without first duping them.
+      @resource_type.freeze
+      @resource_labels.freeze
+      @common_labels.freeze
 
       # Log an informational message containing the Logs viewer URL
+      # TODO: once the log viewer supports v2 services, use resource_type and
```
qingling128: Changed to the V2 format.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -162,7 +172,7 @@ class GoogleCloudOutput < BufferedOutput
     attr_reader :running_on_managed_vm
     attr_reader :gae_backend_name
     attr_reader :gae_backend_version
-    attr_reader :service_name
+    attr_reader :resource_type
```
qingling128: I split the resource object into type and labels because the `MonitoredResource` object is different between the grpc path and non-grpc path. If `type` and `labels` are stored simply as `string` and `hash`, we can construct the two different types of `MonitoredResource` objects at the end of the method to avoid extra `if @use_grpc ... else ...` blocks along the way.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -46,7 +46,17 @@ class GoogleCloudOutput < BufferedOutput
     CLOUDFUNCTIONS_SERVICE = 'cloudfunctions.googleapis.com'
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -18,10 +18,10 @@
...
 require 'yaml'
 require 'google/apis'
-require 'google/apis/logging_v1beta3'
-require 'google/logging/v1/logging_pb'
-require 'google/logging/v1/logging_services_pb'
-require 'google/logging/v1/log_entry_pb'
+require 'google/logging/v2/logging_pb'
+require 'google/logging/v2/logging_services_pb'
+require 'google/logging/v2/log_entry_pb'
+require 'google/apis/logging_v2beta1'
```
qingling128: Nah, no special reason. I just changed it back.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: I meant individual label values, i.e., can each `label` be `nil`?

lib/fluent/plugin/out_google_cloud.rb
qingling128: `Google::Protobuf::Map` does not support `merge!`. Looking at the implementation, `merge_into_self` behaves like `merge!`, but is restricted to be used internal only:

```
// Used only internally -- shared by #merge and #initialize.
VALUE Map_merge_into_self(VALUE _self, VALUE hashmap) {
```

One option is to store the non-grpc version of `MonitoredResource` with `type` and `labels` along the way, and convert it to the grpc version in the end. Would we prefer that?

lib/fluent/plugin/out_google_cloud.rb
qingling128: Moved this to the `LogEntry` level (The original implementation only set it there).

lib/fluent/plugin/out_google_cloud.rb
qingling128: Oh. `label` would never be `nil`. Either `labels_to_extract` would be an array of strings (if `resource_type` is dataflow or ml), or the method returns early, not reaching this point.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -282,41 +300,57 @@ def configure(conf)
...
+      # without first duping them.
+      @resource_type.freeze
+      @resource_labels.freeze
+      @common_labels.freeze
 
       # Log an informational message containing the Logs viewer URL
       @log.info 'Logs viewer address: ',
-                'https://console.developers.google.com/project/', @project_id,
-                '/logs?service=', @service_name, '&key1=instance&key2=', @vm_id
+                'https://pantheon.corp.google.com/logs/viewer?project=',
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -553,21 +631,19 @@ def write(chunk)
...
-                common_labels: labels,
+              Google::Apis::LoggingV2beta1::WriteLogEntriesRequest.new(
+                log_name: log_name,
+                resource: Google::Apis::LoggingV2beta1::MonitoredResource.new(
+                  type: group_resource_type,
+                  labels: group_resource_labels
+                ),
+                labels: group_common_labels,
                 entries: entries)
+            write_request.labels = nil if write_request.labels.empty?
```
qingling128: I traced back to the original commit that added this logic and copied the comment from there:
"# Remove the labels if we didn't populate it with anything.".

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -435,73 +519,69 @@ def write(chunk)
...
+                type: entry_resource_type,
+                labels: entry_resource_labels
+              ),
+              severity: severity,
+              timestamp: {
+                seconds: ts_secs,
+                nanos: ts_nanos
+              }
+            )
+            # Remove the labels if we didn't populate it with anything.
```
igorpeshansky: s/it/them/?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -435,73 +519,69 @@ def write(chunk)
...
+            entry.resource.labels = nil if entry.resource.labels.empty?
```
igorpeshansky: Might be better to do this before the `entry` is created, as
```
entry_resource_labels = nil if entry_resource_labels.empty?
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -282,41 +300,57 @@ def configure(conf)
...
+      # without first duping them.
+      @resource_type.freeze
+      @resource_labels.freeze
+      @common_labels.freeze
 
       # Log an informational message containing the Logs viewer URL
       @log.info 'Logs viewer address: ',
-                'https://console.developers.google.com/project/', @project_id,
-                '/logs?service=', @service_name, '&key1=instance&key2=', @vm_id
+                'https://console.cloud.google.com/logs/viewer?project=',
```
igorpeshansky: Should we use interpolation, e.g.,
```
      @log.info 'Logs viewer address: ',
                'https://console.cloud.google.com/logs/viewer',
                "?project=#{@project_id}&resource=#{@resource_type}/",
                "instance_id/#{@vm_id}"
```
?

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: We could also do this:
```
module Google
  module Protobuf
    # Add Map::merge!.
    class Map
      def merge!(hash)
        self.merge_into_self(hash)
      end
    end
  end
end
```

qingling128: The `merge_into_self` method is never registered in the C implementation:
https://github.com/google/protobuf/blob/master/ruby/ext/google/protobuf_c/map.c#L808
Thus we can't easily patch the method at Ruby level I'm afraid.

And if I use `merge` instead of `merge!`, it would hurt the performance, which is not what we want either, right?

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Ah, I thought `merge_into_self` was a Ruby method. We can do this instead, then:
```
module Google
  module Protobuf
    # Add Map::merge!.
    class Map
      def merge!(hash)
        hash.each { |k, v| self[k] = v }
        self
      end
    end
  end
end
```

qingling128: Ran into some weird issue when implementing this. A bug has been filed with Protobuf component to have it fixed.

Meanwhile, I've modified the implementation to store the non-grpc version of `MonitoredResource` and convert it to the grpc version in the end. A `TODO` has been added to construct the grpc version of `MonitoredResource` from the beginning after the protobuf bug is fixed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -18,10 +18,10 @@
 require 'time'
 require 'yaml'
 require 'google/apis'
-require 'google/apis/logging_v1beta3'
-require 'google/logging/v1/logging_pb'
-require 'google/logging/v1/logging_services_pb'
-require 'google/logging/v1/log_entry_pb'
+require 'google/apis/logging_v2beta1'
```
summit: Can you use logging_v2 instead?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -435,73 +518,65 @@ def write(chunk)
...
           end
 
           entries.push(entry)
         end
         # Don't send an empty request if we rejected all the entries.
         next if entries.empty?
 
-        log_name = log_name(tag, labels)
+        log_name = "projects/#{@project_id}/logs/#{log_name(
+          tag, group_resource.type, group_resource.labels)}"
```
summit: Is the log_id URL encoded?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -958,50 +1027,58 @@ def decode_cloudfunctions_function_name(function_name)
...
       record.delete('docker')
+      [resource_labels, common_labels]
     end
 
-    def field_to_label(record, field, labels, label)
-      return unless record.key?(field)
-      labels[label] = record[field].to_s
-      record.delete(field)
+    def field_to_label(record, field, label)
+      labels = {}
```
igorpeshansky: How about just:
```
def field_to_label(record, field, label)
  record.key?(field) ? {label => record.delete(field).to_s} : {}
```
?

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Weird -- somehow I still see it... Is github playing tricks on me?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1049,52 +1126,78 @@ def struct_from_ruby(hash)
...
         entry.text_payload = convert_to_utf8(record['message'])
       else
-        entry.struct_payload = struct_from_ruby(record)
+        entry.json_payload = struct_from_ruby(record)
       end
     end
 
-    def log_name(tag, common_labels)
-      if @service_name == CLOUDFUNCTIONS_SERVICE
+    def log_name(tag, resource_type, resource_labels)
```
igorpeshansky: Does it make sense to pass the resource in here as well?

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Looks like there isn't one -- only `logging_v1beta3` or `logging_v2beta1`: https://github.com/google/google-api-ruby-client/tree/master/generated/google/apis.

test/plugin/base_test.rb
```diff
@@ -526,16 +631,16 @@ def test_one_log_ec2
     verify_log_entries(1, EC2_PARAMS)
   end
 
-  def test_struct_payload_log
+  def test_json_payload_log
```
igorpeshansky: I wonder if we should just change s/struct_payload/structured_payload/ instead of switching to "json_payload"... WDYT?

test/plugin/base_test.rb
```diff
@@ -1230,42 +1318,66 @@ def cloudfunctions_log_entry_text_not_matched(i)
...
           end
         end
 
-        assert_equal params[:zone], entry['metadata']['zone']
-        assert_equal params[:service_name], entry['metadata']['serviceName']
-        check_labels entry, batch['commonLabels'], params[:labels]
+        # per-entry resource or log_name overrides the corresponding field
+        # from the request.  Labels are merged, with the per-entry label
+        # taking precedence in case of overlap.
+        resource = \
```
igorpeshansky: How about:
```
resource = entry['resource'] or request['resource']
log_name = entry['logName'] or request['logName']
```
?

test/plugin/base_test.rb
```diff
@@ -1312,8 +1424,8 @@ def assert_equal_with_default(_field, _expected_value, _default_value, _entry)
     _undefined
   end
 
-  # Get the fields of the struct payload.
-  def get_fields(_struct_payload)
+  # Get the fields of the json payload.
+  def get_fields(_json_payload)
```
igorpeshansky: I'd just call it `_payload`...

test/plugin/test_out_google_cloud.rb
```diff
@@ -243,9 +231,9 @@ def assert_equal_with_default(field, expected_value, _default_value, entry)
     end
   end
 
-  # Get the fields of the struct payload.
-  def get_fields(struct_payload)
-    struct_payload
+  # Get the fields of the json payload.
+  def get_fields(json_payload)
```
igorpeshansky: I'd just call it `payload`... Also in the other test.

test/plugin/base_test.rb
```diff
@@ -1052,7 +1145,7 @@ def test_http_request_from_record
     end
     verify_log_entries(1, COMPUTE_PARAMS, 'httpRequest') do |entry|
       assert_equal http_request_message, entry['httpRequest'], entry
```
igorpeshansky: Do we still need to override `http_request_message`? If not, let's use the constant directly instead of this method.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Oh, this is intended. I added back in a later commit, since we are using `resource.dup` now. Without this patch, the resource is not duplicating labels (the new and the old resources point to the same labels...)

```
$ cat inner_map_not_dup.rb
require 'google/apis/logging_v2beta1'

old_resource = Google::Apis::LoggingV2beta1::MonitoredResource.new(
  labels: { 'aaa' => 'bbb' })
new_resource = old_resource.dup
new_resource.labels['aaa'] = 'ccc'
puts "new resource:"
puts new_resource.labels
puts "old resource:"
puts old_resource.labels

$ ruby inner_map_not_dup.rb
new resource:
{"aaa"=>"ccc"}
old resource:
{"aaa"=>"ccc"}
```

On a second thought, this should be a bug, right?

lib/fluent/plugin/out_google_cloud.rb
qingling128: For the grpc path, `log_name` would be converted to be utf-8 encoded. For the non-grpc path, it won't.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -190,6 +200,10 @@ def configure(conf)
       end
 
       # TODO: Send instance tags as labels as well?
+      # TODO(lingshi) Construct Google::Api::MonitoredResource when @use_grpc is
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,6 +370,110 @@ def format(tag, time, record)
       [tag, time, record].to_msgpack
     end
 
+    def compute_group_resource_and_labels(tag)
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,6 +370,110 @@ def format(tag, time, record)
...
+          group_resource.labels['container_name'] =
+            match_data['container_name']
+          %w(namespace_name pod_name).each do |field|
+            group_common_labels["#{CONTAINER_SERVICE}/#{field}"] =
+              match_data[field]
+          end
+        end
+      end
+
+      # freeze the per-request state.  Any further changes must be made
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,6 +370,110 @@ def format(tag, time, record)
...
+      # on a per-entry basis.
+      group_resource.freeze
+      group_resource.labels.freeze
+      group_common_labels.freeze
+
+      [group_resource, group_common_labels]
+    end
+
+    def compute_entry_resource_and_labels(group_resource, group_common_labels,
```
qingling128: Added a comment and modified this to return just the extra labels and merge in the caller.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

test/plugin/base_test.rb
qingling128: Yeah, I found the `json_payload` name a bit too specific, too, especially when some of the tests are testing no json input. Just changed the test names as suggested.

test/plugin/base_test.rb

test/plugin/base_test.rb
qingling128: Interesting. Changed. Seems `||` and `or` have different precedence, so I used `||` instead.

```
$ cat test.rb
a = nil
b = "meow"
c = a || b
d = a or b
e = (a or b)
puts c.inspect
puts d.inspect
puts e.inspect

$ ruby test.rb
"meow"
nil
"meow"
```

test/plugin/test_out_google_cloud.rb
qingling128: Changed here and also in the grpc test file.

lib/fluent/plugin/out_google_cloud.rb
qingling128: nvm, just checked the Ruby doc, `dup` is intended for a shallow copy, so is `clone`. We would have to override it this way.

http://ruby-doc.org/core-2.4.0/Object.html#method-i-dup

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -190,8 +200,12 @@ def configure(conf)
       end
 
       # TODO: Send instance tags as labels as well?
+      # TODO: Construct Google::Api::MonitoredResource when @use_grpc is
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,6 +370,113 @@ def format(tag, time, record)
...
+      # per-entry basis.
+      group_resource.freeze
+      group_resource.labels.freeze
+      group_common_labels.freeze
+
+      [group_resource, group_common_labels]
+    end
+
+    # Extract entry resource and common labels that should be applied to
+    # individual entries from the group resource and group common labels.
```
qingling128: Nice catch! Yeah, it turned out the group common labels were not being used, so I removed them. Just removed it from the comment as well.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,6 +370,113 @@ def format(tag, time, record)
...
+          common_labels.merge!(extracted_common_labels)
+        end
+      end
+
+      # If a field is present in the label_map, send its value as a label
+      # (mapping the field name to label name as specified in the config)
+      # and do not send that field as part of the payload.
+      if @label_map
+        @label_map.each do |field, label|
+          common_labels.merge!(
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,6 +370,113 @@ def format(tag, time, record)
...
+            field_to_label(record, field, label))
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -344,74 +487,20 @@ def write(chunk)
...
-            # Move the stdout/stderr annotation from the record into a label.
-            field_to_label(record, 'stream', entry.metadata.labels,
-                           "#{CONTAINER_SERVICE}/stream")
-            # If the record has been annotated by the kubernetes_metadata_filter
-            # plugin, then use that metadata. Otherwise, rely on commonLabels
-            # populated at the grouped_entries level from the group's tag.
-            if record.key?('kubernetes')
-              handle_container_metadata(record, entry)
-            end
+          entry_resource = group_resource.dup
```

lib/fluent/plugin/out_google_cloud.rb
qingling128: No worries.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1063,6 +1060,19 @@ def field_to_label(record, field, label)
       record.key?(field) ? { label => record.delete(field).to_s } : {}
     end
 
+    # For every original_label => new_label pair in the label_map, delete the
+    # original_label from the record if it exists, and extract the value to form
+    # a new map with the new_label as the new key.
+    def fields_to_label(record, label_map)
+      extracted_labels = {}
+      return extracted_labels if label_map.nil? || !label_map.is_a?(Hash)
```
igorpeshansky: This whole function could be written as:
```
def fields_to_labels(record, label_map)
  return {} if label_map.nil? || !label_map.is_a?(Hash)
  label_map.inject({}) do |extracted_labels, (original_label, new_label)|
    extracted_labels[new_label] = record.delete(original_label).to_s if \
      record.key?(original_label);
    extracted_labels
  end
end
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -958,50 +1030,69 @@ def decode_cloudfunctions_function_name(function_name)
...
     end
 
-    def field_to_label(record, field, labels, label)
-      return unless record.key?(field)
-      labels[label] = record[field].to_s
-      record.delete(field)
+    # For every original_label => new_label pair in the label_map, delete the
+    # original_label from the record if it exists, and extract the value to form
+    # a new map with the new_label as the new key.
+    def fields_to_label(record, label_map)
```
igorpeshansky: `fields_to_labels`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -958,50 +1030,69 @@ def decode_cloudfunctions_function_name(function_name)
...
+      record.key?(field) ? { label => record.delete(field).to_s } : {}
     end
 
-    def field_to_label(record, field, labels, label)
-      return unless record.key?(field)
-      labels[label] = record[field].to_s
-      record.delete(field)
+    # For every original_label => new_label pair in the label_map, delete the
+    # original_label from the record if it exists, and extract the value to form
+    # a new map with the new_label as the new key.
```
igorpeshansky: A few too many `new`s in this sentence...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,6 +371,109 @@ def format(tag, time, record)
...
+            extract_container_metadata(record)
+          resource_labels.merge!(extracted_resource_labels)
+          common_labels.merge!(extracted_common_labels)
+        end
+      end
+
+      # If a field is present in the label_map, send its value as a label
+      # (mapping the field name to label name as specified in the config)
+      # and do not send that field as part of the payload.
+      common_labels.merge!(fields_to_label(record, @label_map)) if @label_map
```
igorpeshansky: The `if @label_map` test is superfluous (`fields_to_label` returns an empty hash in that case).

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,6 +371,109 @@ def format(tag, time, record)
...
+      if group_resource.type == CLOUDFUNCTIONS_RESOURCE_TYPE &&
+         record.key?('log')
+        @cloudfunctions_log_match =
+          @cloudfunctions_log_regexp.match(record['log'])
+      end
+
+      if group_resource.type == CONTAINER_RESOURCE_TYPE
+        # Move the stdout/stderr annotation from the record into a label
+        common_labels.merge!(
+          field_to_label(record, 'stream', "#{CONTAINER_SERVICE}/stream"))
```
igorpeshansky: Isn't this the same as `fields_to_labels(record, {'stream' => "#{CONTAINER_SERVICE}/stream"})`? Also in all other uses of `field_to_label`...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -958,50 +1030,69 @@ def decode_cloudfunctions_function_name(function_name)
...
+    def extract_container_metadata(record)
+      resource_labels = {}
+      common_labels = {}
+      %w(namespace_id pod_id container_name).each do |field|
+        resource_labels.merge!(
+          field_to_label(record['kubernetes'], field, field))
+      end
+      %w(namespace_name pod_name).each do |field|
+        common_labels.merge!(
+          field_to_label(record['kubernetes'], field,
```
igorpeshansky: If you change `field_to_label` to `fields_to_labels` as I suggested above, you could also merge these two loops into one call.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Interesting. Rubocop recommended `each_with_object` over `inject` for this case, so I used `each_with_object` instead.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Right. I've just refactored these occurrences and deleted `field_to_labels`.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Right. Removed this check.

lib/fluent/plugin/out_google_cloud.rb
qingling128: But one of them is for `resource_labels`, while the other one is for `common_labels`. They can't be merged to one call for this reason.

lib/fluent/plugin/out_google_cloud.rb
qingling128: LOL. I removed a few.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -958,50 +1030,64 @@ def decode_cloudfunctions_function_name(function_name)
...
+    def extract_container_metadata(record)
+      resource_labels = {}
+      common_labels = {}
+      %w(namespace_id pod_id container_name).each do |field|
+        resource_labels.merge!(
+          fields_to_labels(record['kubernetes'], field => field))
+      end
+      %w(namespace_name pod_name).each do |field|
+        common_labels.merge!(
+          fields_to_labels(record['kubernetes'], field =>
```
igorpeshansky: Let's break after the comma instead.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -958,50 +1030,64 @@ def decode_cloudfunctions_function_name(function_name)
...
-    def field_to_label(record, field, labels, label)
-      return unless record.key?(field)
-      labels[label] = record[field].to_s
-      record.delete(field)
+    # For every original_label => new_label pair in the label_map, delete the
+    # original_label from the record if it exists, and extract the value to form
+    # a map with the new_label as the key.
+    def fields_to_labels(record, label_map)
+      return {} if label_map.nil? || !label_map.is_a?(Hash)
+      label_map.each_with_object({}) \
```
igorpeshansky: Why not `inject({})`? It's shorter. Also, I find the order a bit more natural but it's personal preference. WDYT?

igorpeshansky: Oh, didn't read your rubocop comment... Sigh, ok, let's stay with that.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Oh, whoops, missed that.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Cool.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -263,12 +278,19 @@ def configure(conf)
...
+        @resource.type = COMPUTE_RESOURCE_TYPE
+        # TODO: introduce a new MonitoredResource-centric configuration and
+        # deprecate subservice-name; for now, translate known uses.
         if @subservice_name
-          @service_name = @subservice_name
+          # TODO: what should we do if we encounter an unknown value?
+          if @subservice_name == DATAFLOW_SERVICE
+            @resource.type = DATAFLOW_RESOURCE_TYPE
+          elsif @subservice_name == ML_SERVICE
+            @resource.type = ML_RESOURCE_TYPE
```
summit: What if a service maps to multiple resource types? Can we avoid using the concept of service_names in v2?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -282,41 +304,56 @@ def configure(conf)
...
-        common_labels["#{COMPUTE_SERVICE}/resource_type"] = 'instance'
-        common_labels["#{COMPUTE_SERVICE}/resource_id"] = @vm_id
+        # Some services have the GCE instance_id and zone as MonitoredResource
+        # labels; for other services we send them as entry labels.
+        if @resource.type == COMPUTE_RESOURCE_TYPE ||
+           @resource.type == CONTAINER_RESOURCE_TYPE
+          @resource.labels['instance_id'] = @vm_id
+          @resource.labels['zone'] = @zone
+        else
+          common_labels["#{COMPUTE_SERVICE}/resource_id"] = @vm_id
```
summit: What case is this for? Why is it using 'compute.googleapis.com' here? In v2, we should encourage using URL-less labels.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: This is what @mr-salty did in his earlier PR, and, as I understand, replicates what the v1 API implementation did with labels.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: All of the `common_labels` values start with the service URL... Are you suggesting that this one should be different?

lib/fluent/plugin/out_google_cloud.rb
qingling128: Yeah, `subservice_name` is a config we support to override subservice detection (each subservice more or less corresponds to a specific resource type). We would need to support the same behavior after V2 migration.

For the other resource types, I was reading the discussion from the original V2 migration thread. It sounds to me: The initial goal is to support the services we currently support, without requiring users to make changes to their configuration. We should eventually enable users to specify the resource type and any required labels explicitly in the configuration. Yet that is not something we are planning to do for this round. And we can only fully deprecate the concept of "service name" then.

lib/fluent/plugin/out_google_cloud.rb
summit: We should avoid using service names when using the v2 API. Labels should just be key value pairs, and the keys should not use a service prefix unless absolutely required to be backwards compatible.

lib/fluent/plugin/out_google_cloud.rb
summit: Given that users can take a long time to update their fluentd on various platforms, and the v1 logging API turndown will be a significant motivation to update, I think it makes better sense to get it right on the first move to v2.

igorpeshansky: One of the main motivations on our side is to make the v2 migration transparent to the user (beyond an agent upgrade). This means that their existing configs should continue working, and, in particular, service name overrides and label mappings should work as expected. Switching from one resource type per service to multiple resource types per service is an orthogonal change that we can make at a later point, when we're already writing to v2.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: As I understand it, outside of the resource labels, all common labels are one big shared pool. This prefixing actually provides a namespace for these metrics, so you could search the logs for, e.g., Cloud Functions version independently from some other, user-defined notion of version.
Are we going to recommend that users use the same, e.g., "version" label for everything? That would also clash with them defining their own label from some payload field, because it'll override the one we compute automatically. Not saying this is the wrong approach, but we do need to tread carefully.
Thoughts?

qingling128: Per discussion offline, we came to consensus that we would keep the prefixes of the common labels in the current phase of migration, so the migration could be transparent to users and the risk of clashing user defined labels and agent generated labels is low (users are aware we reserved these prefixes).

Once a better approach to address these labels is established (requires changes downstream in the Logging API), we would make corresponding changes to replace these url based labels.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -18,10 +18,10 @@
 require 'time'
 require 'yaml'
 require 'google/apis'
-require 'google/apis/logging_v1beta3'
-require 'google/logging/v1/logging_pb'
```
igorpeshansky: You should be able to completely remove `lib/google` on this branch, since we're now using the gRPC bindings from `google-cloud-logging`...

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Have you built a gem with this version already? Do we need to bump the version again?

test/plugin/base_test.rb
```diff
@@ -40,6 +40,24 @@ def setup
   # generic attributes
   HOSTNAME = Socket.gethostname
 
+  WRITE_LOG_ENTRIES_URI = 'https://logging.googleapis.com/v2beta1/entries:write'
+
+  APPENGINE_SERVICE_NAME = 'appengine.googleapis.com'
```
igorpeshansky: Just realized: why are we defining these constants here and not reusing the ones from `out_google_cloud.rb`? And if we do need to redefine them, why the naming inconsistency (`APPENGINE_SERVICE_NAME` here vs. `APPENGINE_SERVICE` in `out_google_cloud.rb`)?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -39,14 +39,24 @@ class GoogleCloudOutput < BufferedOutput
     Fluent::Plugin.register_output('google_cloud', self)
 
     PLUGIN_NAME = 'Fluentd Google Cloud Logging plugin'
-    PLUGIN_VERSION = '0.5.4.grpc.alpha.1'
+    PLUGIN_VERSION = '0.5.5.v2.alpha.1'
 
     # Constants for service names.
     APPENGINE_SERVICE = 'appengine.googleapis.com'
```
igorpeshansky: I'm wondering if it's better to have a single constant for each service/resource, rather than two parallel sets of variables, e.g.:
```
APPENGINE_CONSTANTS = {service: 'appengine.googleapis.com', resource_type: 'gae_app'}
CLOUDFUNCTIONS_CONSTANTS = {service: 'cloudfunctions.googleapis.com', resource_type: 'cloud_function'}
COMPUTE_CONSTANTS = {service: 'compute.googleapis.com', resource_type: 'gce_instance'}
...
```
and then reference them as `APPENGINE_CONSTANTS[:service]` and `APPENGINE_CONSTANTS[:resource_type]`, respectively?

test/plugin/base_test.rb
qingling128: Changed. The full path is pretty long, so I assign them to local constants at the beginning of the file. Seems there is no way to import a class in a module in Ruby.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Just removed `lib/google` folder.

lib/fluent/plugin/out_google_cloud.rb
qingling128: I haven't released any gem from this branch. `0.5.5.pre.1` was released from master. And we are planning to release `0.5.5` soon. So I will update this to `0.5.6.v2.alpha.1`.

test/plugin/base_test.rb
igorpeshansky: A class in a module - no. But we could pull those constants out into their own module and import them into both classes...

igorpeshansky:
LGTM :shipit:

PR #115, opened 2017-01-30
Fix timestamp null value handling with grpc enabled.

qingling128:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -437,10 +437,15 @@ def write(chunk)
 
           ts_secs, ts_nanos = compute_timestamp(record, time)
           if @use_grpc
-            entry.metadata.timestamp = Google::Protobuf::Timestamp.new(
-              seconds: ts_secs,
-              nanos: ts_nanos
-            )
+            # If "seconds" or "nanos" are not integers (e.g. null), we will emit
```
andrewsg: I think "emit" here is intended to be "omit".

qingling128: Oh god, right. Just changed this.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -437,10 +437,15 @@ def write(chunk)
 
           ts_secs, ts_nanos = compute_timestamp(record, time)
           if @use_grpc
-            entry.metadata.timestamp = Google::Protobuf::Timestamp.new(
-              seconds: ts_secs,
-              nanos: ts_nanos
-            )
+            # If "seconds" or "nanos" are not integers (e.g. null), we will omit
+            # the timestamp field and defer the decision of whether to set it to
```
igorpeshansky: s/of whether to set it to the current timestamp/on how too handle it/

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -437,10 +437,15 @@ def write(chunk)
...
+            # the current timestamp to the downstream Logging API.
+            if ts_secs.is_a?(Integer) && ts_nanos.is_a?(Integer)
```
igorpeshansky: This seems to be a much broader check than what you're testing below. Shouldn't you also check non-integer values of the timestamp? Should you check when only one of `seconds` or `nanos` is `nil`?

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Added more tests and changed the behavior a bit to be exactly aligned to how Logging API handles the cases (See details in the comments).

test/plugin/test_out_google_cloud.rb
```diff
@@ -191,6 +191,30 @@ def test_parse_severity
...
+    ].each_with_index do |input, index|
+      setup_logging_stubs do
+        d = create_driver
+        @logs_sent = []
+        d.emit('message' => log_entry(0), 'timestamp' => input)
+        d.run
+      end
+      verify_log_entries(1, COMPUTE_PARAMS) do |entry|
+        assert_equal input, entry['metadata']['timestamp'],
+                     "Index #{index} failed with entry: #{entry.inspect}"
```
igorpeshansky: [Optional] I wouldn't bother with the index here -- the entry is enough.

test/plugin/test_out_google_cloud.rb
```diff
@@ -191,6 +191,30 @@ def test_parse_severity
     assert_equal('DEFAULT', test_obj.parse_severity('er'))
   end
 
+  def test_non_integer_timestamp
+    setup_gce_metadata_stubs
+    time = Time.now
+    [
+      { 'seconds' => nil, 'nanos' => nil },
+      { 'seconds' => nil, 'nanos' => time.tv_nsec },
+      { 'seconds' => nil, 'nanos' => time.tv_nsec },
```
igorpeshansky: Any reason for the duplication?

test/plugin/test_out_google_cloud.rb
igorpeshansky: Ditto in `test_out_google_cloud_grpc.rb`...

qingling128: But neither `entry` nor `expected` is sufficient to decide which test is failing for the grpc side.

igorpeshansky: Ah, you're right. Then perhaps use the stringified input instead of the index in both places?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -155,6 +155,32 @@ def test_struct_payload_non_utf8_log
...
+    }.each do |input, expected|
+      setup_logging_stubs do
+        d = create_driver
+        @logs_sent = []
+        d.emit('message' => log_entry(0), 'timestamp' => input)
+        d.run
+      end
+      verify_log_entries(1, COMPUTE_PARAMS) do |entry|
+        assert_equal expected, entry['metadata']['timestamp'],
+                     "Test with timestamp input '#{input}' failed with " \
```
igorpeshansky: s/timestamp input/timestamp/

igorpeshansky: Let's do the same in `test_out_google_cloud.rb`.

qingling128: Changed.

qingling128: Done.

test/plugin/test_out_google_cloud.rb
```diff
@@ -208,7 +208,9 @@ def test_non_integer_timestamp
         d.run
       end
       verify_log_entries(1, COMPUTE_PARAMS) do |entry|
-        assert_equal timestamp, entry['metadata']['timestamp'], entry
+        assert_equal timestamp, entry['metadata']['timestamp'],
+                     "Test with timestamp '#{timestamp}' failed with " \
```
igorpeshansky: Nit: s/failed with/failed for/. Also in the other file.

test/plugin/test_out_google_cloud.rb
```diff
@@ -191,6 +191,30 @@ def test_parse_severity
...
+    ].each do |timestamp|
+      setup_logging_stubs do
+        d = create_driver
+        @logs_sent = []
+        d.emit('message' => log_entry(0), 'timestamp' => timestamp)
+        d.run
+      end
+      verify_log_entries(1, COMPUTE_PARAMS) do |entry|
+        assert_equal timestamp, entry['metadata']['timestamp'],
+                     "Test with timestamp '#{timestamp}' failed with " \
```
igorpeshansky: Nit: s/failed with/failed for/. Also in the other file.

igorpeshansky:
LGTM :shipit:

PR #116, opened 2017-01-31
Fix non-string tag handling and encode log name for gRPC.

qingling128:


test/plugin/test_out_google_cloud.rb
```diff
@@ -229,10 +242,10 @@ def http_request_message
   end
 
   # Set up http stubs to mock the external calls.
-  def setup_logging_stubs
-    [COMPUTE_PARAMS, VMENGINE_PARAMS, CONTAINER_FROM_TAG_PARAMS,
-     CONTAINER_FROM_METADATA_PARAMS, CLOUDFUNCTIONS_PARAMS, CUSTOM_PARAMS,
-     EC2_PARAMS].each do |params|
+  def setup_logging_stubs(params_list = [])
```
igorpeshansky: s/`params_list`/`extra_stub_params`/

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -155,6 +155,21 @@ def test_struct_payload_non_utf8_log
     end
   end
 
+  def test_tags
+    setup_gce_metadata_stubs
+    [123, 'test', 'germanß', 'chinese中'].each do |tag|
```
igorpeshansky: Why is this missing the special character test?

test/plugin/test_out_google_cloud.rb
```diff
@@ -191,6 +191,19 @@ def test_parse_severity
...
+  def test_tags
+    setup_gce_metadata_stubs
+    [123, 'test', 'germanß', 'chinese中', 'specialCharacter_-.'].each do |tag|
+      setup_logging_stubs([COMPUTE_PARAMS.merge(log_name: tag.to_s)]) do
+        @logs_sent = []
+        d = create_driver(APPLICATION_DEFAULT_CONFIG, tag)
+        d.emit('msg' => log_entry(0))
+        d.run
+      end
+      verify_log_entries(1, COMPUTE_PARAMS, 'structPayload')
```
igorpeshansky: Why don't we check the log name here as we do in the grpc test?

test/plugin/test_out_google_cloud_grpc.rb
qingling128: Because I added that very last minute, and forgot to duplicate it. Just changed.

test/plugin/test_out_google_cloud.rb

test/plugin/test_out_google_cloud.rb
qingling128: For non-grpc, log name is reflected in the url instead of in the log entry itself. I just added some comments to clarify this as well:

`  # For non-grpc path, log name is reflected in the url, so we are stubbing the
  # corresponding url containing the converted tag. To verify the tag is
  # converted properly, we simply check the log entry is recorded successfully,
  # and no exception (e.g. stubbing missing for certain url) is thrown.`

The majority of the test cases are identical between grpc and non-grpc, but the setup_logging_stub part requires some special treatment for non-grpc. Also the way to verify converted tag is slightly different. Thus this ended up in two separate files instead of in the `base_test` (which i find pretty annoying).

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -338,6 +338,19 @@ def write(chunk)
...
       chunk.msgpack_each do |tag, *arr|
+        tag = tag.to_s if tag.is_a?(Integer)
+        # Logging API requires log names to be strings with only alphanumeric
+        # characters. Generate warnings and drop the log if the tag is not a
+        # string or if it contains non-utf8 characters since these will be
+        # rejected by Logging API anyway. For other cases when the tag is a
+        # string with illegal characters, defer the rejection to the downstream
+        # Logging API since it checks against certain special characters as
+        # well, and that list is subject to changes in the future.
+        if !tag.is_a?(String) || convert_to_utf8(tag) != tag
```
qingling128: Good idea. Added.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -338,6 +338,19 @@ def write(chunk)
       # Group the entries since we have to make one call per tag.
       grouped_entries = {}
       chunk.msgpack_each do |tag, *arr|
+        tag = tag.to_s if tag.is_a?(Integer)
```
qingling128: I was debating how strict we should treat the tags. `require_valid_tags` is a good idea, so we can treat it differently based on user's preference. Just changed the implementation to be customizable.

qingling128:
Added a fix for log name encoding here, since the tag related tests are newly created in this pull request. The commit message has been amended to reflect this.

test/plugin/test_out_google_cloud.rb
```diff
@@ -191,6 +191,46 @@ def test_parse_severity
     assert_equal('DEFAULT', test_obj.parse_severity('er'))
   end
 
+  # For non-grpc path, log name is reflected in the url, so we are stubbing the
+  # corresponding url containing the converted tag. To verify the tag is
+  # converted properly, we simply check the log entry is recorded successfully,
+  # and no exception (e.g. stubbing missing for certain url) is thrown.
+  def test_tag_acceptance
```
igorpeshansky: There is a lot of duplication between this and the same test in `test_out_google_cloud_grpc.rb`. Can we factor some of it out into a common method?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -155,6 +155,46 @@ def test_struct_payload_non_utf8_log
...
+  def test_tag_acceptance
+    setup_gce_metadata_stubs
+    [
+      # When require_valid_tags is on, we only convert integer tags to strings.
+      # Other invalid cases will cause the log to be dropped.
+      [123, '123', true],
+      ['test', 'test', true],
+      ['germanß', 'germanß', true],
+      ['chinese中', 'chinese中', true],
+      ['specialCharacter/_-.', 'specialCharacter/_-.', true],
```
igorpeshansky: Are we not supposed to URL-encode the log name here? Also in line 177.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -504,10 +535,11 @@ def write(chunk)
             labels_utf8_pairs = labels.map do |k, v|
               [k.encode('utf-8'), convert_to_utf8(v)]
             end
-            utf8_log_name = convert_to_utf8(log_name)
 
+            # Encode the log name since Logging API relies on '/' as delimiter.
+            log_name = URI.encode(log_name).gsub('/', '%2F')
```
igorpeshansky: Let's use `ERB::Util.url_encode` instead of special-casing `'/'`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -87,6 +87,13 @@ class GoogleCloudOutput < BufferedOutput
     # The subservice_name overrides the subservice detection, if provided.
     config_param :subservice_name, :string, :default => nil
 
+    # Valid tags require strings with only alphanumeric characters. If this
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -87,6 +87,13 @@ class GoogleCloudOutput < BufferedOutput
...
+    # config is set to true, log entries with invalid tags would be dropped and
+    # a warning would be generated in the log. If this config is set to false,
+    # we will force the tag to be valid by converting any non-string tag to
+    # string, and remove any non-utf8 or other invalid characters from it.
+    config_param :require_valid_tags, :bool, :default => true
```
qingling128: Because... technically speaking we are modifying clients' data? On a second thought, this might be better than dropping log entries. I've just changed the default value.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -193,6 +200,8 @@ def configure(conf)
       @common_labels = {}
       @common_labels.merge!(@labels) if @labels
 
+      @invalid_tag_regexp = %r{[^[:alpha:]\d/._-]}
```
qingling128: This is the only accepted characters by the Logging API right now. If we let users customize this tag, doesn't that mean the converted tags could still get rejected by the Logging API?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -338,6 +347,27 @@ def write(chunk)
       # Group the entries since we have to make one call per tag.
       grouped_entries = {}
       chunk.msgpack_each do |tag, *arr|
+        tag = tag.to_s
```
qingling128: Do you mean when `require_valid_tags` is false? I guess you meant when `require_valid_tags` is true, we should not do `tag.to_s` because any non-string tags should be rejected right away? I've modified it to be this way.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -338,6 +347,27 @@ def write(chunk)
...
+        if @require_valid_tags
+          # When require_valid_tags is true, invalid cases will cause the log to
```
qingling128: This comment seems outdated and should have been removed in commit d83ef7e already.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -338,6 +347,27 @@ def write(chunk)
...
+          # be dropped and warnings generated in the log.
+          if convert_to_utf8(tag) != tag || tag.match(@invalid_tag_regexp)
+            @log.warn "Dropping log message(s) with invalid tag: '#{tag}'. " \
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -338,6 +347,27 @@ def write(chunk)
...
+                      'A tag should be a string with alphanumeric characters.'
+            next
+          end
+        else
+          # When require_valid_tags is false, try to convert any invalid tag by
+          # stripping off invalid characters.
+          converted_tag = convert_to_utf8(tag, '').gsub(@invalid_tag_regexp, '')
```
qingling128: Spaces in tags are not allowed by the Logging API. The only characters other than alphanumeric ones that are allowed are: `-`, `_`, `/` and `.`

Our previous implementation (replacing non-utf8 characters with spaces in tags) would actually fail the Logging API requests. The user configured replacement string can contain characters other than this set, thus we can not safely use that.

Potentially we can hard code a string (only includes alphanumeric, `-`, `_`, `/` and `.`) to replace invalid characters with. Or we can add a separate config if we want the users to be able to configure this string, but make them choose from these characters.

Thoughts?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -338,6 +347,27 @@ def write(chunk)
...
+          if converted_tag == ''
+            @log.warn "Dropping log message(s) with invalid tag: '#{tag}'. " \
```
qingling128: I've added the logic to return `'_'` when the original tag or the converted tag turned out to be `''`. Corresponding tests have been modified as well.

lib/fluent/plugin/out_google_cloud.rb

test/plugin/base_test.rb
```diff
@@ -128,6 +128,10 @@ def setup
   )
   # rubocop:enable Metrics/LineLength
 
+  NO_REQUIRE_VALID_TAGS_CONFIG = %(
+    require_valid_tags false
```

test/plugin/base_test.rb
```diff
@@ -626,6 +630,36 @@ def test_struct_payload_json_container_log
     end
   end
 
+  def test_tag_rejection
+    setup_gce_metadata_stubs
+    [
+      # When require_valid_tags is true, for any invalid characters detected,
+      # verify we drop the log instead of trying to convert the tag to something
```

test/plugin/test_out_google_cloud.rb
qingling128: Done. The common method `verify_tag_acceptance` has been extracted and moved to `base_test`.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: This line seems outdated. Encoding should have been added in commit d83ef7e already.

test/plugin/test_out_google_cloud.rb
```diff
@@ -191,6 +191,20 @@ def test_parse_severity
...
+    setup_logging_stubs_block = lambda do |converted_tag, setup_driver_block|
+      setup_logging_stubs([COMPUTE_PARAMS.merge(log_name: converted_tag)]) do
+        setup_driver_block.call
+      end
+    end
+    verify_log_name_block = ->(converted_tag) {}
```
igorpeshansky: FWIW, if we modify `setup_logging_stubs` to add the appropriate `logName` field to each entry in @logs_sent, we should be able to reuse the log name verification code in both modes.

test/plugin/base_test.rb
```diff
@@ -1280,6 +1305,50 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
...
+        if require_valid_tags
+          d = create_driver(REQUIRE_VALID_TAGS_CONFIG, tag)
+        else
+          d = create_driver(APPLICATION_DEFAULT_CONFIG, tag)
+        end
+        d.emit('msg' => log_entry(0))
+        d.run
+      end
+      setup_logging_stubs_block.call(converted_tag, setup_driver_block)
+      verify_log_entries(1, COMPUTE_PARAMS, 'structPayload')
```
igorpeshansky: Remember, `verify_log_entries` takes a block for extra verification, so this could instead be:
```
verify_log_entries(1, COMPUTE_PARAMS, 'structPayload') do |entry|
  assert_equal "projects/#{PROJECT_ID}/logs/#{converted_tag}",
               entry['logName']
end

File: test/plugin/test_out_google_cloud.rb [hunk 21]
```diff
@@ -191,6 +191,20 @@ def test_parse_severity
     assert_equal('DEFAULT', test_obj.parse_severity('er'))
   end
 
+  # For non-grpc path, log name is reflected in the url, so we are stubbing the
+  # corresponding url containing the converted tag. To verify the tag is
+  # converted properly, we simply check the log entry is recorded successfully,
+  # and no exception (e.g. stubbing missing for certain url) is thrown.
+  def test_tag_acceptance
+    setup_logging_stubs_block = lambda do |converted_tag, setup_driver_block|
+      setup_logging_stubs([COMPUTE_PARAMS.merge(log_name: converted_tag)]) do
```
igorpeshansky: Just realized that we only need this one stub in this case. Does it make more sense to default to the set of stubs we usually set up, and allow overriding the whole set (i.e., making this the actual `stub_params`, rather than `extra_stub_params`)?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -155,6 +155,21 @@ def test_struct_payload_non_utf8_log
     end
   end
 
+  # For grpc path, the log name is sent as a part of the log entry, thus can be
+  # verified directly.
+  def test_tag_acceptance
+    setup_logging_stubs_block = lambda do |_converted_tag, setup_driver_block|
+      setup_logging_stubs do
```
igorpeshansky: Just realized that our grpc tests don't actually verify the incoming RPC requests (i.e., our mock will simply accept any request and return the configured response). Probably a good idea to add some expected RPCs there...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,10 +343,38 @@ def format(tag, time, record)
       [tag, time, record].to_msgpack
     end
 
+    # Convert tags.
+    # If 'require_valid_tags' is true, reject any non-string tags or tags with
```
qingling128: Changed to this except the first line (probably there by accident). The method name has been changed as well.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,10 +343,38 @@ def format(tag, time, record)
...
+    # invalid characters by returning nil.
+    # If 'require_valid_tags' is false, convert any non-string tags to strings
+    # and strip off any invalid characters.
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,10 +343,38 @@ def format(tag, time, record)
...
+    def convert_tag(tag)
+      if @require_valid_tags
+        if tag.is_a?(String) && convert_to_utf8(tag) == tag &&
+           !tag.match(@invalid_tag_regexp)
+          return tag
+        else
+          return nil
+        end
+      else
+        converted = convert_to_utf8(tag.to_s, '').gsub(@invalid_tag_regexp, '')
```

test/plugin/base_test.rb
```diff
@@ -626,6 +630,27 @@ def test_struct_payload_json_container_log
     end
   end
 
+  # When require_valid_tags is true, for any invalid characters detected, verify
```

test/plugin/base_test.rb
```diff
@@ -626,6 +630,27 @@ def test_struct_payload_json_container_log
...
+  # that we drop the log entries instead of trying to convert the tag to
+  # something valid.
+  def test_tag_rejection
```

test/plugin/base_test.rb
```diff
@@ -1280,6 +1305,50 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
     assert i == n, "Number of entries #{i} does not match expected number #{n}"
   end
 
+  # Verify tags. Detailed implementation defers between the grpc and non-gprc
```

qingling128: Oops. Changed.

test/plugin/base_test.rb
```diff
@@ -1280,6 +1305,50 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
...
+  # paths, which is reflected in setup_logging_stubs_block and
+  # verify_log_name_block.
+  def verify_tag_acceptance(setup_logging_stubs_block, verify_log_name_block)
```

test/plugin/base_test.rb
```diff
@@ -1280,6 +1305,50 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
...
+    setup_gce_metadata_stubs
+    [
+      # When require_valid_tags is on, we only accept string tags with valid
+      # characters.
+      ['test', 'test', true],
```

test/plugin/base_test.rb
```diff
@@ -1280,6 +1305,50 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
...
+      ['germanß', 'german%C3%9F', true],
+      ['chinese中', 'chinese%E4%B8%AD', true],
+      ['specialCharacter/_-.', 'specialCharacter%2F_-.', true],
+
+      # When require_valid_tags is off, we try to convert invalid tags by
```

test/plugin/base_test.rb
```diff
@@ -1280,6 +1305,50 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
...
+      # stripping off invalid characters.
```

test/plugin/base_test.rb
qingling128: Previously non-grpc path is not able to verify `logName` so I had to add that ugly extra block param. I've just added `logName` assignment in the `setup_logging_stubs` method for the non-grpc path as recommended in the other comment, and removed this mess.

test/plugin/test_out_google_cloud.rb
qingling128: Done.

test/plugin/test_out_google_cloud.rb
qingling128: Great idea! Changed.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: Right. Just added equivalent verification to the grpc side. Thank god I can get rid of those messy lambda definitions now.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: True. Ok, let's keep this hard-coded.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Yes, exactly. Sorry for the confusion.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: I think this has now been taken care of by URL encoding. Resolving.

test/plugin/test_out_google_cloud.rb
```diff
@@ -229,12 +229,15 @@ def http_request_message
...
   # Set up http stubs to mock the external calls.
-  def setup_logging_stubs
-    [COMPUTE_PARAMS, VMENGINE_PARAMS, CONTAINER_FROM_TAG_PARAMS,
-     CONTAINER_FROM_METADATA_PARAMS, CLOUDFUNCTIONS_PARAMS, CUSTOM_PARAMS,
-     EC2_PARAMS].each do |params|
+  def setup_logging_stubs(override_stub_params = nil)
+    stub_params = [COMPUTE_PARAMS, VMENGINE_PARAMS, CONTAINER_FROM_TAG_PARAMS,
+                   CONTAINER_FROM_METADATA_PARAMS, CLOUDFUNCTIONS_PARAMS,
+                   CUSTOM_PARAMS, EC2_PARAMS]
+    stub_params = [override_stub_params] unless override_stub_params.nil?
```

test/plugin/base_test.rb
```diff
@@ -626,6 +630,87 @@ def test_struct_payload_json_container_log
...
+  # When 'require_valid_tags' is true, for any invalid characters detected,
+  # verify that we drop the log entries instead of trying to convert the tag to
+  # something valid.
+  def test_reject_invalid_tags
+    setup_gce_metadata_stubs
+    [
+      123,
+      "nonutf8#{[0x92].pack('C*')}",
+      [1, 2, 3],
+      { key: 'value' }
```
qingling128: Added.

test/plugin/base_test.rb
```diff
@@ -626,6 +630,87 @@ def test_struct_payload_json_container_log
...
+        assert_equal "projects/#{PROJECT_ID}/logs/#{converted_tag}",
+                     @logs_sent[0]['logName']
+      end
+    end
+  end
+
+  # When 'require_valid_tags' is true, for any invalid characters detected,
+  # verify that we drop the log entries instead of trying to convert the tag to
+  # something valid.
+  def test_reject_invalid_tags
```
qingling128: Moved the test and updated the name.

test/plugin/base_test.rb
```diff
@@ -626,6 +630,87 @@ def test_struct_payload_json_container_log
...
+        d.run
+      end
+      verify_log_entries(1, COMPUTE_PARAMS, 'structPayload') do
+        assert_equal "projects/#{PROJECT_ID}/logs/#{converted_tag}",
+                     @logs_sent[0]['logName']
+      end
+    end
+  end
+
+  # When 'require_valid_tags' is true, for any invalid characters detected,
```
qingling128: Updated.

test/plugin/base_test.rb
```diff
@@ -626,6 +630,87 @@ def test_struct_payload_json_container_log
...
+    }.each do |tag, converted_tag|
+      setup_logging_stubs(COMPUTE_PARAMS.merge(log_name: converted_tag)) do
+        @logs_sent = []
+        d = create_driver(APPLICATION_DEFAULT_CONFIG, tag)
+        d.emit('msg' => log_entry(0))
+        d.run
+      end
+      verify_log_entries(1, COMPUTE_PARAMS, 'structPayload') do
+        assert_equal "projects/#{PROJECT_ID}/logs/#{converted_tag}",
+                     @logs_sent[0]['logName']
```
qingling128: The `logName` is not in the `entry` level. So I modified the `verify_log_entries` method to add a param `log_name` to get rid of `@logs_sent[0]`.

test/plugin/base_test.rb
```diff
@@ -626,6 +630,87 @@ def test_struct_payload_json_container_log
...
+      'germanß' => 'german%C3%9F',
+      'chinese中' => 'chinese%E4%B8%AD',
+      'specialCharacter/_-.#@*&^' => 'specialCharacter%2F_-.%23%40%2A%26%5E',
+      "nonutf8#{[0x92].pack('C*')}" => 'nonutf8%20',
+      [1, 2, 3] => '%5B1%2C%202%2C%203%5D',
+      { key: 'value' } => '%7B%22key%22%3D%3E%22value%22%7D',
+      "abc#{[0x92].pack('C*')}" => 'abc%20',
+      "#{[0x92].pack('C*')}" => '%20',
+      'abc@&^$*' => 'abc%40%26%5E%24%2A',
+      '@&^$*' => '%40%26%5E%24%2A'
```
qingling128: Added empty string tests in `test_reject_invalid_tags_with_require_valid_tags_true` and `test_encode_and_sanitize_tags_with_require_valid_tags_false`.

test/plugin/base_test.rb
```diff
@@ -626,6 +630,87 @@ def test_struct_payload_json_container_log
...
+    {
+      'test' => 'test',
+      'germanß' => 'german%C3%9F',
+      'chinese中' => 'chinese%E4%B8%AD',
+      'specialCharacter/_-.' => 'specialCharacter%2F_-.'
+    }.each do |tag, converted_tag|
+      setup_logging_stubs(COMPUTE_PARAMS.merge(log_name: converted_tag)) do
+        @logs_sent = []
+        d = create_driver(REQUIRE_VALID_TAGS_CONFIG, tag)
+        d.emit('msg' => log_entry(0))
```
qingling128: Added tests `test_encode_tags_from_container_name_with_require_valid_tags_true` and `test_sanitize_tags_from_container_name_with_require_valid_tags_false`.

test/plugin/base_test.rb
```diff
@@ -626,6 +630,87 @@ def test_struct_payload_json_container_log
     end
   end
 
+  # Verify that invalid tags are sanitized. Detailed implementation differs
+  # between the grpc and non-gprc paths, which is reflected in
+  # 'setup_logging_stubs_block' and 'verify_log_name_block'.
+  # When 'require_valid_tags' is true, we only accept string tags with valid
+  # characters.
+  def test_sanitize_invalid_tags_with_requires_valid_tags_true
```
qingling128: Changed the method names and updated the comments.

test/plugin/base_test.rb
```diff
@@ -626,6 +630,87 @@ def test_struct_payload_json_container_log
     end
   end
 
+  # Verify that invalid tags are sanitized. Detailed implementation differs
+  # between the grpc and non-gprc paths, which is reflected in
+  # 'setup_logging_stubs_block' and 'verify_log_name_block'.
```
qingling128: Updated.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1133,13 +1160,14 @@ def api_client
     # non-UTF-8 character would be replaced by the string specified by
     # 'non_utf8_replacement_string'. If 'coerce_to_utf8' is set to false, any
     # non-UTF-8 character would trigger the plugin to error out.
-    def convert_to_utf8(input)
+    def convert_to_utf8(input, replacement = nil)
+      replacement = @non_utf8_replacement_string if replacement.nil?
```
qingling128: Nope. Reverted this part.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -504,10 +531,10 @@ def write(chunk)
             labels_utf8_pairs = labels.map do |k, v|
               [k.encode('utf-8'), convert_to_utf8(v)]
             end
-            utf8_log_name = convert_to_utf8(log_name)
 
+            log_name = ERB::Util.url_encode(log_name)
```
qingling128: Yes, `log_name` could be extracted from `common_labels`. I've added sanitization there as well (near `def log_name`)

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,12 +341,32 @@ def format(tag, time, record)
...
+    # Given a tag, returns the corresponding valid tag if possible, or nil if
+    # the tag should be rejected. If 'require_valid_tags' is false, non-string
+    # tags are converted to strings, and invalid characters are sanitized;
+    # otherwise such tags are rejected.
+    def sanitize_tag(tag)
+      if @require_valid_tags &&
+         (!tag.is_a?(String) || tag == '' || convert_to_utf8(tag) != tag)
+        return nil
+      end
+      tag = '_' if tag == ''
```

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -224,12 +224,22 @@ def api_client
...
+    def initialize(stub_params, requests_received)
       super()
       @requests_received = requests_received
+      @expected_requests = stub_params.clone
     end
 
     def write_log_entries(request, _call)
+      matched = @expected_requests.take_while do |expected|
+        "projects/#{expected[:project_id]}/logs/#{expected[:log_name]}" \
+          == request.log_name
```
qingling128: The non-grpc side checks both the `project_id` and log_name` though. Might as well be equivalent?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -224,12 +224,22 @@ def api_client
...
+      end
+      unless matched
+        message = 'This request is not expected by the gRPC mock service: \n' \
```
qingling128: Changed.

lib/fluent/plugin/out_google_cloud.rb
qingling128: `invalid_tag_regexp` no longer exists though. I removed it and rely on url encoding instead.

test/plugin/test_out_google_cloud_grpc.rb
igorpeshansky: I meant that you could construct a list of expected string values for `request.log_name`. Something like:
```
@expected_log_names = stub_params.map { |r| "projects/#{r[:project_id]}/logs/#{r[:log_name]}" }
```
and then iterate over that:
```
matched = @expected_log_names.take_while do |expected|
  expected == request.log_name
end
```
WDYT?

test/plugin/test_out_google_cloud.rb
```diff
@@ -229,12 +229,15 @@ def http_request_message
   end
 
   # Set up http stubs to mock the external calls.
-  def setup_logging_stubs
-    [COMPUTE_PARAMS, VMENGINE_PARAMS, CONTAINER_FROM_TAG_PARAMS,
-     CONTAINER_FROM_METADATA_PARAMS, CLOUDFUNCTIONS_PARAMS, CUSTOM_PARAMS,
-     EC2_PARAMS].each do |params|
+  def setup_logging_stubs(override_stub_params = [])
+    stub_params = [COMPUTE_PARAMS, VMENGINE_PARAMS, CONTAINER_FROM_TAG_PARAMS,
```

test/plugin/test_out_google_cloud.rb
```diff
@@ -229,12 +229,15 @@ def http_request_message
   end
 
   # Set up http stubs to mock the external calls.
-  def setup_logging_stubs
-    [COMPUTE_PARAMS, VMENGINE_PARAMS, CONTAINER_FROM_TAG_PARAMS,
-     CONTAINER_FROM_METADATA_PARAMS, CLOUDFUNCTIONS_PARAMS, CUSTOM_PARAMS,
-     EC2_PARAMS].each do |params|
+  def setup_logging_stubs(override_stub_params = [])
```

test/plugin/base_test.rb
```diff
@@ -626,6 +630,144 @@ def test_struct_payload_json_container_log
...
+    end
+  end
+
+  # Verify that tags are properly encoded and sanitized. When
+  # 'require_valid_tags' is false, we try to convert any non-string tags to
+  # strings, and replace non-utf8 characters with a replacement string.
+  def test_sanitize_tags_with_require_valid_tags_false
+    setup_gce_metadata_stubs
+    {
+      123 => '123',
```
qingling128: The tags have been put into constants and consolidated.

test/plugin/base_test.rb
```diff
@@ -626,6 +630,144 @@ def test_struct_payload_json_container_log
...
+      '' => '_',
+      'test' => 'test',
+      'germanß' => 'german%C3%9F',
```
qingling128: The tags have been put into constants and consolidated.

test/plugin/base_test.rb
```diff
@@ -626,6 +630,144 @@ def test_struct_payload_json_container_log
...
+        d = create_driver(REQUIRE_VALID_TAGS_CONFIG, tag)
+        d.emit('msg' => log_entry(0))
+        d.run
+      end
+      verify_log_entries(0, COMPUTE_PARAMS, 'structPayload')
+    end
+  end
+
+  # A map between tags and their encoded version.
+  def encoded_tags
```
qingling128: Changed this to be a constant.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -334,12 +339,33 @@ def format(tag, time, record)
...
-        grouped_entries[tag] = [] unless grouped_entries.key?(tag)
-        grouped_entries[tag].push(arr)
+        sanitized_tag = sanitize_tag(tag)
+        if sanitized_tag.nil?
+          @log.warn "Dropping log entries with invalid tag: '#{tag}'. " \
+                    'A tag should be a string with utf8 characters.'
+          next
+        end
+        grouped_entries[sanitized_tag] = [] \
+          unless grouped_entries.key?(sanitized_tag)
```
qingling128: Right. Changed.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: Ah, I see. Yeah, this could save the extraction for each write_log_entries execution. Changed.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Moved this into the `def log_name` method. Note that we can only encode the log name for the grpc path, since the non-grpc path already handles encoding.

test/plugin/base_test.rb
qingling128: Reverted and added a comment to explain why we only verify log name in the first element.

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -283,14 +293,19 @@ def _undefined
   end
 
   # Set up grpc stubs to mock the external calls.
-  def setup_logging_stubs(should_fail = false, code = 0, message = 'Ok')
+  def setup_logging_stubs(override_stub_params = nil, should_fail = false,
+                          code = 0, message = 'Ok')
+    stub_params = [COMPUTE_PARAMS, VMENGINE_PARAMS, CONTAINER_FROM_TAG_PARAMS,
```
igorpeshansky: Same here:
```
stub_params = override_stub_params || \
              [COMPUTE_PARAMS, VMENGINE_PARAMS, CONTAINER_FROM_TAG_PARAMS,
               CONTAINER_FROM_METADATA_PARAMS, CLOUDFUNCTIONS_PARAMS,
               CUSTOM_PARAMS, EC2_PARAMS]
```

test/plugin/base_test.rb
```diff
@@ -626,6 +654,118 @@ def test_struct_payload_json_container_log
...
+      verify_log_entries(1, COMPUTE_PARAMS, 'structPayload')
+      # Each batch in the logs sent shares the same log name, so we only verify
+      # the first one.
+      assert_equal "projects/#{PROJECT_ID}/logs/#{encoded_tag}",
+                   @logs_sent[0]['logName']
+    end
+  end
+
+  # Verify that tags extracted from container names are properly encoded.
+  def test_encode_tags_from_container_name_with_require_valid_tags_true
```
igorpeshansky: What about rejecting invalid tags extracted from containers? Is that even possible?

test/plugin/base_test.rb
```diff
@@ -626,6 +654,118 @@ def test_struct_payload_json_container_log
     end
   end
 
+  # Verify that we drop the log entries when 'require_valid_tags' is true and
+  # any non-string tags or tags with non-utf8 characters are detected.
+  def test_reject_invalid_tags_with_require_valid_tags_true
+    setup_gce_metadata_stubs
+    invalid_tags = NON_STRING_TAGS.merge(EMPTY_TAG).merge(NON_UTF8_TAGS)
+    invalid_tags.each do |tag, _|
```
igorpeshansky:
```
invalid_tags.keys.each do |tag|
```

test/plugin/base_test.rb
```diff
@@ -320,6 +324,30 @@ def setup
...
+  # Tags and their sanitized and encoded version.
+  VALID_TAGS = {
+    'test' => 'test',
+    'germanß' => 'german%C3%9F',
+    'chinese中' => 'chinese%E4%B8%AD',
+    'specialCharacter/_-.' => 'specialCharacter%2F_-.',
+    'abc@&^$*' => 'abc%40%26%5E%24%2A',
+    '@&^$*' => '%40%26%5E%24%2A'
+  }
+  NON_STRING_TAGS = {
```
igorpeshansky: You almost always join these three into one hash. How about defining only one `INVALID_TAGS` hash, and then in the one case where you only need nonempty string tags, filtering for those:
```
INVALID_TAGS.select { |k, _| k.is_a? String and !k.empty?}
```

test/plugin/base_test.rb
```diff
@@ -626,6 +654,118 @@ def test_struct_payload_json_container_log
...
+    setup_gce_metadata_stubs
+    VALID_TAGS.each do |tag, encoded_tag|
+      setup_logging_stubs([COMPUTE_PARAMS.merge(log_name: encoded_tag)]) do
+        @logs_sent = []
+        d = create_driver(REQUIRE_VALID_TAGS_CONFIG, tag)
+        d.emit('msg' => log_entry(0))
+        d.run
+      end
+      verify_log_entries(1, COMPUTE_PARAMS, 'structPayload')
+      # Each batch in the logs sent shares the same log name, so we only verify
```
igorpeshansky: s/the logs sent/@logs_sent/. Also everywhere you have this comment.
However, I just realized that the previous line always checks that there's exactly one log entry, making this comment superfluous.

test/plugin/base_test.rb
```diff
@@ -626,6 +654,118 @@ def test_struct_payload_json_container_log
...
+  def test_sanitize_tags_from_container_name_with_require_valid_tags_false
+    setup_gce_metadata_stubs
+    setup_container_metadata_stubs
+    # Log names are derived from container names for containers. And container
+    # names are extracted from the tag based on a regex match pattern. As a
+    # prerequisite, the tag should already be a string, thus we only test
+    # non-empty string cases here.
+    non_empty_string_tags = VALID_TAGS.merge(NON_UTF8_TAGS)
+    non_empty_string_tags.each do |container_name, encoded_container_name|
+      params = CONTAINER_FROM_METADATA_PARAMS.clone
```
igorpeshansky: Do you think something like the following would be more readable?
```
params = CONTAINER_FROM_METADATA_PARAMS.merge(
    {labels: CONTAINER_FROM_METADATA_PARAMS[:labels].merge(
        {"#{CONTAINER_SERVICE_NAME}/container_name" =>
            URI.decode(encoded_container_name)}),
         log_name: encoded_container_name})
```

test/plugin/base_test.rb
```diff
@@ -626,6 +654,118 @@ def test_struct_payload_json_container_log
...
+      params[:labels] = CONTAINER_FROM_METADATA_PARAMS[:labels].clone
+      # Container name in the label is not encoded, while the log name is.
```
igorpeshansky: s/not encoded/sanitized but not encoded/, right?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -224,12 +224,22 @@ def api_client
...
+        "projects/#{param[:project_id]}/logs/#{param[:log_name]}"
+      end
     end
 
     def write_log_entries(request, _call)
+      matched = @expected_log_names.take_while do |expected|
+        expected == request.log_name
+      end
+      unless matched
+        message = "Unexpected request: #{request.inspect}"
```
igorpeshansky: Let's inline `message` into the next call.

test/plugin/base_test.rb
qingling128: I tried to include that test initially, then figured it was kind of redundant:

For empty string case:
The tag fails the kubernetes regex matching, thus fall back to the generic tag as the log name case, which is already covered by `test_reject_invalid_tags_with_require_valid_tags_true`. In other words, it's not really testing container specific stuff.

For non-string cases:
By the time it reaches to the point where the kubernetes regex can be matched and container name being extracted, the tag is certainly gonna be a string.

For non-utf8 cases:
With `require_valid_tags` true, any tag containing non-utf8 strings would be rejected early on before reaching any container specific logic, which is also covered by `test_reject_invalid_tags_with_require_valid_tags_true` already.

Thoughts?

test/plugin/base_test.rb

test/plugin/base_test.rb
qingling128: Fixed.

test/plugin/base_test.rb
qingling128: Ah, right. I've removed these comments.

test/plugin/base_test.rb

test/plugin/base_test.rb

test/plugin/test_out_google_cloud_grpc.rb
qingling128: Oh, yeah, missed this one. Changed now.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: Done.

test/plugin/base_test.rb
```diff
@@ -626,6 +650,109 @@ def test_struct_payload_json_container_log
...
+  # sanitized.
+  def test_sanitize_tags_from_container_name_with_require_valid_tags_false
+    setup_gce_metadata_stubs
+    setup_container_metadata_stubs
+    # Log names are derived from container names for containers. And container
+    # names are extracted from the tag based on a regex match pattern. As a
+    # prerequisite, the tag should already be a string, thus we only test
+    # non-empty string cases here.
+    VALID_TAGS.merge(
+      INVALID_TAGS.select { |tag, _| tag.is_a?(String) && !tag.empty? }
```
igorpeshansky: You could `merge` first and `select` after, making this somewhat simpler:
```
string_tags = VALID_TAGS.merge(INVALID_TAGS).select do |tag, _|
    tag.is_a?(String) && !tag.empty?
end
string_tags.each do ...
```

test/plugin/base_test.rb
```diff
@@ -626,6 +650,109 @@ def test_struct_payload_json_container_log
...
+                   @logs_sent[0]['logName']
+    end
+  end
+
+  # Verify that tags are properly encoded and sanitized. When
+  # 'require_valid_tags' is false, we try to convert any non-string tags to
+  # strings, and replace non-utf8 characters with a replacement string.
+  def test_sanitize_tags_with_require_valid_tags_false
+    setup_gce_metadata_stubs
+    VALID_TAGS.merge(INVALID_TAGS).each do |tag, sanitized_tag|
```
igorpeshansky: How about defining `ALL_TAGS = VALID_TAGS.merge(INVALID_TAGS)` above?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -283,14 +291,20 @@ def _undefined
...
 
   # Set up grpc stubs to mock the external calls.
-  def setup_logging_stubs(should_fail = false, code = 0, message = 'Ok')
+  def setup_logging_stubs(override_stub_params = nil, should_fail = false,
+                          code = 0, message = 'Ok')
+    stub_params = override_stub_params || \
+                  [COMPUTE_PARAMS, VMENGINE_PARAMS, CONTAINER_FROM_TAG_PARAMS,
+                   CONTAINER_FROM_METADATA_PARAMS, CLOUDFUNCTIONS_PARAMS,
+                   CUSTOM_PARAMS, EC2_PARAMS]
+    stub_params = override_stub_params unless override_stub_params.nil?
```
igorpeshansky: Not needed anymore.

test/plugin/base_test.rb
qingling128: Yeah, it's possible. but the container name is a part of the tag, and the tag got rejected (if `require_valid_tags` is true) or converted to non-utf-8 (if `require_valid_tags` is false) before reaching the kubernete specific logic (use the regex to extract the container name).

One of the issues in the bug ticket is that the kubernetes regex `match` function would throw exceptions for non-utf8 characters.

Do you think we should test that (container name with non-utf8 characters) anyway just to make sure it's behaving as expected?

test/plugin/base_test.rb

test/plugin/base_test.rb

test/plugin/test_out_google_cloud_grpc.rb
qingling128: oh! Removed.

test/plugin/base_test.rb
qingling128: Just included such tests in a separate commit `65f0b6e8e9cafb2b11bcffaf01fcf86d94fbb4dc`.

test/plugin/base_test.rb
```diff
@@ -626,6 +654,157 @@ def test_struct_payload_json_container_log
...
+        d.emit('msg' => log_entry(0))
+        d.run
+      end
+      verify_log_entries(0, COMPUTE_PARAMS, 'structPayload')
+    end
+  end
+
+  # Verify that empty string container name should fail the kubernetes regex
+  # match, thus the original tag is used as the log name.
+  def test_handle_empty_container_name_with_require_valid_tags_true
```
igorpeshansky: Is this test dependent on the value of `require_valid_tags`?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -224,12 +224,20 @@ def api_client
...
+    def initialize(stub_params, requests_received)
       super()
       @requests_received = requests_received
+      @expected_log_names = stub_params.map do |param|
+        "projects/#{param[:project_id]}/logs/#{param[:log_name]}"
+      end
     end
 
     def write_log_entries(request, _call)
+      matched = @expected_log_names.take_while do |expected|
```
igorpeshansky: Is `take_while` the right thing to use here? Should we have used `select` instead?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -224,12 +224,20 @@ def api_client
...
+        expected == request.log_name
+      end
+      fail GRPC::BadStatus.new(99, "Unexpected request: #{request.inspect}") \
+        unless matched
```
igorpeshansky: Empty arrays are truthy in Ruby. Did you mean `if matched.empty?` instead?

test/plugin/base_test.rb
```diff
@@ -626,6 +654,157 @@ def test_struct_payload_json_container_log
...
+    end
+  end
+
+  # Verify that empty string container name should fail the kubernetes regex
+  # match, thus the original tag is used as the log name.
+  def test_handle_empty_container_name
+    setup_gce_metadata_stubs
+    setup_container_metadata_stubs
+    container_name = ''
+    # This tag will not match the kubernetes regex because the container name
```
igorpeshansky: ...because it requires a non-empty container name.

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -224,12 +224,20 @@ def api_client
...
+        "projects/#{param[:project_id]}/logs/#{param[:log_name]}"
+      end
     end
 
     def write_log_entries(request, _call)
+      matched = @expected_log_names.select do |expected|
+        expected == request.log_name
+      end
+      fail GRPC::BadStatus.new(99, "Unexpected request: #{request.inspect}") \
+        if matched.empty?
```
igorpeshansky: In fact, now that we're doing straight-up string equality, this is just:
```
fail GRPC::BadStatus.new(99, "Unexpected request: #{request.inspect}") \
  unless @expected_log_names.include?(request.log_name)
```
which lets you get rid of the whole `matched` bit.

test/plugin/base_test.rb
qingling128: Sweet.

test/plugin/base_test.rb
qingling128: Nope. I just changed the test name.

test/plugin/base_test.rb
```diff
@@ -626,6 +654,157 @@ def test_struct_payload_json_container_log
...
+    end
+  end
+
+  # Verify that empty string container name should fail the kubernetes regex
+  # match, thus the original tag is used as the log name.
+  def test_handle_empty_container_name_with_require_valid_tags_true
+    setup_gce_metadata_stubs
+    setup_container_metadata_stubs
+    container_name = ''
+    tag = container_tag_with_container_name(container_name)
```
qingling128: Added.

test/plugin/base_test.rb
```diff
@@ -626,6 +654,111 @@ def test_struct_payload_json_container_log
...
+  # Verify that tags extracted from container names are properly encoded and
+  # sanitized.
+  def test_sanitize_tags_from_container_name_with_require_valid_tags_false
+    setup_gce_metadata_stubs
+    setup_container_metadata_stubs
+    # Log names are derived from container names for containers. And container
+    # names are extracted from the tag based on a regex match pattern. As a
+    # prerequisite, the tag should already be a string, thus we only test
+    # non-empty string cases here.
+    string_tags = VALID_TAGS.merge(INVALID_TAGS).select do |tag, _|
```

test/plugin/test_out_google_cloud_grpc.rb
qingling128: You are right, `select` is what we should use. Originally I was trying to find something like `exists`, but seems Ruby does not have anything similar. `take_while` is definitely not right. Changed.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: Ah, I see. I was just wondering why `take_while` would work... This explains. Changed.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: Changed.

test/plugin/base_test.rb

igorpeshansky:
LGTM :shipit:

PR #117, opened 2017-02-09
Fix non-string tag handling and encode log name.

qingling128:
This is equivalent to the fix in https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/116, but in the `api_v2` branch instead of master.

Main differences:

1. Logging API V2 is using a unified url for different resource types, so the `stub_params` and `override_stub_params` no longer make sense, and have been removed.

2. Logging API V2 non-grpc client lib no longer handles encoding (probably because the log name is no longer used as a url). We just need to encode "/" in the tag.

This brings an open question: Should we continue to encode log names at our side regardless (we would be doing it for both grpc and non-grpc sides)? Or should we skip url encoding, which implies we would need to reintroduce the `invalid_tag_regex` (alphanumeric and ./-_) and sanitize tag based on that. Thoughts?

As to `rebase api_v2 from master` or `cherry pick the bug fix to api_v2`, right now I'm doing the latter, since it's easier to review. If necessary I can do the former once the pull request is approved.


qingling128:
That change added to the grpc mock is no longer needed, because both grpc and non-grpc paths now include `log_name` in the `WriteLogEntriesRequest`, so a `log_name`check has been added to the `verify_log_entries` method which should be sufficient:

```
    def verify_json_log_entries(n, params, payload_type = 'textPayload')
        ...
        log_name = entry['logName'] || request['logName']
        ...
        assert_equal "projects/#{params[:project_id]}/logs/#{params[:log_name]}", log_name
```

Ah, this reminds me I should get rid of all the `@logs_sent[0]['logName']` checks since they are redundant now. Just removed them.

As for the version, we never released a gem with the version we have now (`0.5.6.v2.alpha.1`), so it should be good.

igorpeshansky:
Makes sense, thanks. LGTM :shipit:

qingling128:
Thanks! I've just rebased `api_v2` from `master` and verified it's identical to this branch `lingshi_v2_tag_fix`.

Will delete `lingshi_v2_tag_fix` and also re-branch `api_v1` from the current `master`.

PR #119, opened 2017-02-14
Remove erroneous assertion; fix google-api-client version at 0.9.*.

igorpeshansky:


test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -60,7 +60,6 @@ def test_server_error
         begin
           d.run
         rescue GRPC::Cancelled => error
-          assert_equal "GRPC::#{message}", error.message
```
qingling128: Should we add a comment for the reason `Cancelled` got the standard message?

fluent-plugin-google-cloud.gemspec
```diff
@@ -21,9 +21,9 @@ eos
 
   gem.add_runtime_dependency 'fluentd', '~> 0.10'
   gem.add_runtime_dependency 'googleapis-common-protos', '~> 1.3'
-  gem.add_runtime_dependency 'google-api-client', '> 0.9'
+  gem.add_runtime_dependency 'google-api-client', '> 0.9', '< 0.10'
```
dhrupadb: Was the problem in the google-api-client and not grpc?

dhrupadb: Would `'google-api-client', '~>0.9', '< 0.10'` be better in keeping convention with the rest of the file?

test/plugin/test_out_google_cloud_grpc.rb
dhrupadb: Also is the new message arbitrary that we can't assert on it anymore?

qingling128: The new message is "unknown cause"...

fluent-plugin-google-cloud.gemspec
qingling128: I think there is a separate issue with `google-api-client` (which was released on Feb 13).

test/plugin/test_out_google_cloud_grpc.rb
qingling128: It's kind of ambiguous whether that's a bug or `as designed`...

fluent-plugin-google-cloud.gemspec
igorpeshansky: The problem was in both, I'm just punting on the former now and addressing the latter.

test/plugin/test_out_google_cloud_grpc.rb
igorpeshansky: Done.

fluent-plugin-google-cloud.gemspec
igorpeshansky: If we used `~>`, we probably wouldn't need the `<` constraint. I'll try it out.

dhrupadb: I believe `~>` goes to the highest version in that level. So if we did `~>0.9' it would go to version `0.9` and above e.g. `0.10`,`0.11`.. but not to `1.0`.

igorpeshansky: Nope, looks like we still need the upper bound...

dhrupadb: I think you wouldn't need the upper bound if you did `~>0.9.0` but the upper bound is fine. Just thought it would be better to use the same convention as the rest of the file.

igorpeshansky: You're right. Fixed.

qingling128:
LGTM

Issue #84, opened 2016-09-19
missing HttpRequest fields are erroneously set to 'nil' rather than left unset.


mr-salty:
The code that constructs an HttpRequest sets fields that are not present in the input to 'nil' rather than leaving them unset.  This causes those fields to appear in the request sent to the API with null values rather than being omitted as they should be (the [serialization does make a distinction](https://github.com/google/google-api-ruby-client/blob/1ee88320e91deccf4e69fe589b4959046dcc752e/lib/google/apis/core/json_representation.rb#L49)).

so instead of:

`output.request_method = input.delete('requestMethod')`

we should have

`output.request_method = input.delete('requestMethod') if input.key?('requestMethod')`

(FYI, to look at the actual requests I added a line to lib/google/apis/core/api_command.rb):

```
self.body = request_representation.new(request_object).to_json(skip_undefined: true)
File.open('/tmp/requests', 'a') { |file| file.write(self.body); file.write("\n") }
```

christopherL91:
What is the status on this issue? I have the same problem as found in #85. I use nginx and just output the log to stdout which in turn is captured by stackdriver on GCE.

```json
{
 metadata: {…}   
 textPayload: "2016-12-06 17:16:28 +0000 [warn]: Dropping 1 log message(s) error_class="Google::Apis::ClientError" error="badRequest: Invalid JSON payload received. Unknown name \"validated_with_origin_server\" at 'entries[0].http_request': Cannot find field."
"   
 insertId: "1thtsqzekp5jl"   
 log: "fluentd-cloud-logging"   
}
```

ged15:
Hello!
Any progress regarding this issue?

igorpeshansky:
As far as I know, the latest version of the gem (0.6.0) already does this. @qingling128, could you please verify that and resolve this issue?

qingling128:
Looking.

qingling128:
For gRPC path (if you set `use_grpc true` in the config), this has already been fixed: If a field is `null` or absent from the input, it will be left unset in the constructed proto (Setting it to `null` does not apply to the gRPC path since the proto has these fields defined as strings or integers).

[...]

PR #120, opened 2017-03-02
Leave absent fields unset in 'http_request' for non-grpc path.

qingling128:
Backporting the fix to the non-grpc path. This is to address https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/84.


qingling128:
The master branch now has V2 migration included. I just rebased this branch accordingly. It's ready for reviewing.

qingling128:
@igorpeshansky Got a sec for this?

PR #121, opened 2017-03-16
Add Protobuf gem to resolve glibc issue

dhrupadb:
Adding this in here for the output plugin as well so that even if grpc pulls in not the latest gem our config overrides it.

### DIFF
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -24,6 +24,7 @@ eos
   gem.add_runtime_dependency 'google-api-client', '~> 0.9.0'
   gem.add_runtime_dependency 'google-cloud-logging', '~> 0.23.2'
   gem.add_runtime_dependency 'googleauth', '~> 0.4'
+  gem.add_runtime_dependency 'google-protobuf', '~> 3.2.0.2'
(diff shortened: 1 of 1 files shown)


fluent-plugin-google-cloud.gemspec
```diff
@@ -24,6 +24,7 @@ eos
   gem.add_runtime_dependency 'google-api-client', '~> 0.9.0'
   gem.add_runtime_dependency 'google-cloud-logging', '~> 0.23.2'
   gem.add_runtime_dependency 'googleauth', '~> 0.4'
+  gem.add_runtime_dependency 'google-protobuf', '~> 3.2.0.2'
```
qingling128: Seems `grpc` has it own dependency (https://github.com/grpc/grpc/blob/41373d0457a41801297e5fdd0ed03bfe94d26ff2/grpc.gemspec#L30)

`s.add_dependency 'google-protobuf', '~> 3.1'`

Would that affect us?

dhrupadb:
It should get overridden by this dependency. They're asking for ~>3.1 and
so 3.2 is compatible.

qingling128:
LGTM.

dhrupadb:
Closing branch as the gem that is auto pulled in is fixed.

PR #122, opened 2017-04-07
Add latency field in http_request.

qingling128:


.rubocop_todo.yml
```diff
@@ -19,7 +19,7 @@ Metrics/AbcSize:
 Metrics/ClassLength:
   Max: 1300
 Metrics/ModuleLength:
-  Max: 1400
```
qingling128: -___- :facepalm:

fluent-plugin-google-cloud.gemspec
```diff
@@ -19,6 +19,7 @@ eos
   gem.test_files    = gem.files.grep(/^(test)/)
   gem.require_paths = ['lib']
 
+  gem.add_runtime_dependency 'chronic_duration', '~> 0.10.6'
```
qingling128: This is a third party lib that parses strings to duration: https://github.com/henrypoydar/chronic_duration

Is it a good practice to use something like this? Maybe we can fork it if necessary. Or simply use a regex instead?

qingling128:
@igorpeshansky Got a sec for this one?

fluent-plugin-google-cloud.gemspec
igorpeshansky: I would ask the Google Ruby libraries folks. This is a common issue with all of our APIs, and should have a common solution by now.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -961,6 +964,27 @@ def set_http_request(record, entry)
...
+            # Split the integer and decimal parts in order to calculate seconds
+            # and nanos.
+            (integer_part, decimal_part) = duration_in_seconds.divmod 1
+            output.latency = Google::Protobuf::Duration.new(
+              seconds: integer_part,
+              nanos: (decimal_part * NANOS_IN_A_SECOND).round
+            )
+          else
+            # Logging API only takes a string with digits that ends with 's'.
+            output.latency = "#{duration_in_seconds.round(2)}s"
```
igorpeshansky: You can send a hash with `seconds` and `nanos` here as well, just like we do for the timestamp above.

qingling128: Good to know. Changed.

fluent-plugin-google-cloud.gemspec
qingling128: Asked [Steven](https://github.com/sqrrrl) who is a major contributor of [google-api-ruby-client](https://github.com/google/google-api-ruby-client). It's OK to have dependencies on 3rd party libraries. It's a common thing in our open source projects and API clients. Note that the licenses do matter though. Affero General Public License ([AGPL](https://en.wikipedia.org/wiki/Affero_General_Public_License)) is not allowed.

qingling128:
@igorpeshansky Got a sec for this one?

fluent-plugin-google-cloud.gemspec
qingling128: Per discussion offline, we decided to use simple regex instead of this third-party lib since it is not actively maintained. I just modified the implementation.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -67,6 +66,8 @@ module Constants
         service: 'ml.googleapis.com',
         resource_type: 'ml_job'
       }
+
+      LATENCY_REGEX = /^[ \t]*(\d+)(\.\d+)?[ \t]*s[ \t]*$/
```
igorpeshansky: Let's replace `[ \t]` with `\s`...

igorpeshansky: Instead of putting this into the `Constants` module, I'd suggest making this a class variable and moving it to `confgure`, alongside `@cloudfunctions_tag_regexp` and the like. Also, the naming convention in this file uses `regexp` instead of `regex`...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -967,12 +968,16 @@ def set_http_request(record, entry)
...
       unless latency.nil?
-        # Parse latency to duration. If failed, skip setting latency.
-        duration_in_seconds = ChronicDuration.parse(latency)
-        if duration_in_seconds
+        # Parse latency. If no valid format is detected, skip setting latency.
+        # Format: space/tab (optional) + integer + point & decimal (optional)
+        #       + space/tab (optional) + "s" + space/tab (optional)
+        # e.g.: "1.42 s"
+        parsed_groups = latency.scan(LATENCY_REGEX)
+        if !parsed_groups.empty? && parsed_groups[0].length == 2
```
igorpeshansky: You have exactly two captures in the string. If there's a match (`parsed_groups` is non-empty), when will `parsed_groups[0].length` be anything but 2?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -967,12 +968,16 @@ def set_http_request(record, entry)
 
       latency = input.delete('latency')
       unless latency.nil?
-        # Parse latency to duration. If failed, skip setting latency.
-        duration_in_seconds = ChronicDuration.parse(latency)
-        if duration_in_seconds
+        # Parse latency. If no valid format is detected, skip setting latency.
+        # Format: space/tab (optional) + integer + point & decimal (optional)
```
igorpeshansky: s/`space/tab`/`whitespace`/g

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -967,12 +968,16 @@ def set_http_request(record, entry)
...
+        #       + space/tab (optional) + "s" + space/tab (optional)
+        # e.g.: "1.42 s"
+        parsed_groups = latency.scan(LATENCY_REGEX)
+        if !parsed_groups.empty? && parsed_groups[0].length == 2
           # Split the integer and decimal parts in order to calculate seconds
           # and nanos.
-          (latency_seconds, decimal_part) = duration_in_seconds.divmod 1
+          latency_seconds = parsed_groups[0][0].to_i
```
igorpeshansky: This might be a bit more readable if we change the regexp to `/^\s*(?<seconds>\d+)(?<decimal>\.\d+)?\s*s\s*$/`, and then use:
```
match = LATENCY_REGEX.match(latency)
if match
  latency_seconds = match['seconds'].to_i
  decimal_part = match['decimal'].to_f
  ...
end
```

test/plugin/base_test.rb
```diff
@@ -1628,18 +1628,11 @@ def http_request_message_with_absent_referer
...
       '32s' => { 'seconds' => 32, 'nanos' => 0 },
-      '32 second' => { 'seconds' => 32, 'nanos' => 0 },
-      '32 seconds' => { 'seconds' => 32, 'nanos' => 0 },
-      '32 sec' => { 'seconds' => 32, 'nanos' => 0 },
-      '32 secs' => { 'seconds' => 32, 'nanos' => 0 },
       '0.32s' => { 'seconds' => 0, 'nanos' => 320_000_000 },
-      '123' => { 'seconds' => 123, 'nanos' => 0 },
-      '1.3442' => { 'seconds' => 1, 'nanos' => 344_200_000 }
+      ' 123 s ' => { 'seconds' => 123, 'nanos' => 0 },
+      '1.3442 s' => { 'seconds' => 1, 'nanos' => 344_200_000 }
```
igorpeshansky: You mentioned tabs in the latency format description, but aren't testing for those... With `\s`, we could also test for other whitespace...

qingling128: Added.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Changed the location and the name.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Yeah, it's kind of redundant. I just removed it. I was mainly worried what if someone mistakenly touches the regex and the length changes? It might crash the agent if the tests failed to catch it. Yet since we've changed the regex to have the groups named as `seconds` and `decimal`, I think the chance of such mistake should be quite low. I'll stop worrying about it.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -244,6 +242,7 @@ def configure(conf)
         \[(?<timestamp>.{24})\]
         (?:\[(?<execution_id>[^\]]+)\])?
         [ ](?<text>.*)$/x
+      @http_latency_regexp = /^\s*(?<seconds>\d+)(?<decimal>\.\d+)?\s*s\s*$/
```
igorpeshansky: Add a blank line before this, as it's a separate regexp group.

test/plugin/base_test.rb
igorpeshansky: Can we make the tabs explicit, please? I.e., use `"\t\t123\t\ts\t\t"` instead of embedding tabs inline.
Also use `\r`, `\n`, `\v`, `\f`...

qingling128: Added.

lib/fluent/plugin/out_google_cloud.rb

test/plugin/base_test.rb
igorpeshansky: I would have just dumped them all in one test case, but this is fine too.

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -152,6 +152,18 @@ def test_non_integer_timestamp
...
 
+  # The conversions from user input to output.
+  def latency_conversion
+    # TODO(lingshi): Use http://ruby-doc.org/core-2.4.0/Hash.html#method-i-transform_values
+    # instead when we upgrade to ruby 2.4.0.
+    # super.transform_values do |expected|
+    #   expected.reject { |_, value| value == 0 }
+    # end
+    super.map do |input, expected|
+      [input, expected.reject { |_, value| value == 0 }]
```
igorpeshansky: Does the REST path require us to send 'nanos' => 0? If not, can we just get rid of those throughout?

qingling128: Did some experiment. The REST path does not requrie `'nanos' => 0` to be set. Changed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -961,6 +965,30 @@ def set_http_request(record, entry)
...
+          # Split the integer and decimal parts in order to calculate seconds
+          # and nanos.
+          latency_seconds = match['seconds'].to_i
+          latency_nanos = (match['decimal'].to_f * NANOS_IN_A_SECOND).round
+          if @use_grpc
+            output.latency = Google::Protobuf::Duration.new
+          else
+            output.latency = {}
+          end
+          output.latency['seconds'] = latency_seconds \
```
igorpeshansky: This will happen automagically for protobufs, so I would just simplify the hash case:
```
            output.latency = {
              seconds: latency_seconds,
              nanos: latency_nanos
            }.delete_if { |k, v| v == 0 }
```

test/plugin/base_test.rb
```diff
@@ -1368,6 +1368,39 @@ def test_http_request_from_record_with_referer_nil_or_absent
...
+        assert_equal HTTP_REQUEST_MESSAGE.merge('latency' => expected),
+                     entry['httpRequest'], entry
+        assert_nil get_fields(entry['jsonPayload'])['httpRequest'], entry
+      end
+    end
+  end
+
+  def test_http_request_with_null_latency
+    setup_gce_metadata_stubs
+    ['', nil, 'null'].each do |input|
```
igorpeshansky: How about also testing something that doesn't match the regex?

qingling128: Added.

lib/fluent/plugin/out_google_cloud.rb

test/plugin/base_test.rb
```diff
@@ -1368,6 +1368,42 @@ def test_http_request_from_record_with_referer_nil_or_absent
...
+        assert_nil get_fields(entry['jsonPayload'])['httpRequest'], entry
+      end
+    end
+  end
+
+  # Skip setting latency when the field is null or with invalid format.
+  def test_http_request_skip_setting_latency
+    setup_gce_metadata_stubs
+    [
+      '', ' ', nil, 'null', '123', '1.23 seconds', '1min', 'abc&^!$*('
```
igorpeshansky: Let's also add `' 123  s     econds  '` or something like that.

test/plugin/base_test.rb
```diff
@@ -1368,6 +1368,42 @@ def test_http_request_from_record_with_referer_nil_or_absent
...
+      end
+      verify_log_entries(1, COMPUTE_PARAMS, 'httpRequest') do |entry|
+        assert_equal HTTP_REQUEST_MESSAGE.merge('latency' => expected),
+                     entry['httpRequest'], entry
+        assert_nil get_fields(entry['jsonPayload'])['httpRequest'], entry
+      end
+    end
+  end
+
+  # Skip setting latency when the field is null or with invalid format.
```
igorpeshansky: s/`with`/`has`/

test/plugin/base_test.rb
qingling128: Added.

test/plugin/base_test.rb

PR #128, opened 2017-05-15
Clarify how users should run outside of GCP and AWS

witoff:
**Problem**

Running with the default configuration in a test environment fails with "Unable to obtain metadata parameters".  Reading the [docs](https://cloud.google.com/logging/docs/agent/) clearly notes that this agent is _only_ supported in AWS and EC2 which led me to believe that it wouldn't work at all without a metadata service running.  Considering initial use of this agent is likely to be in a local or test environment, this is a misleading first impression.

**Solution**

Clearly document how to run this outside of GCP/AWS in the initial docs.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -26,15 +26,28 @@ will also install and configure the gem.
 To send logs to Google Cloud Logging, specify <code>type google_cloud</code>
 in a
 {match clause}[http://docs.fluentd.org/articles/config-file#2-ldquomatchrdquo-tell-fluentd-what-to-do]
-of your fluentd configuration file, for example:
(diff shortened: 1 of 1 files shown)


witoff:
I signed it!  (i.e. i'm a googler and per docs don't need to sign it.  please let me know if anything else is needed.)

igorpeshansky:
The docs are correct — running outside of GCP or AWS is not supported today, and any hacks that make the agent run in other environments are at the users' own risk and cannot be officially recommended. Your suggested approach will break some of the Stackdriver functionality, e.g., logs-based metrics. Sorry, but we can't accept this change.

PR #129, opened 2017-05-18
When constructing a monitored resource, translate namespace_name and pod_name to namespace_id and pod_id, respectively.

igorpeshansky:
@crassirostris @summit FYI.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -450,12 +450,9 @@ module Fluent
         # Do this here to avoid having to repeat it for each record.
         match_data = @compiled_kubernetes_tag_regexp.match(tag)
         if match_data
-          group_resource.labels['container_name'] =
-            match_data['container_name']
-          %w(namespace_name pod_name).each do |field|
-            group_common_labels["#{CONTAINER_CONSTANTS[:service]}/#{field}"] =
-              match_data[field]
-          end
+          group_resource.labels['container_name'] = match_data['container_name']
+          group_resource.labels['namespace_id'] = match_data['namespace_name']
+          group_resource.labels['pod_id'] = match_data['pod_name']
         end
       end
 
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -453,6 +453,10 @@ module Fluent
           group_resource.labels['container_name'] = match_data['container_name']
           group_resource.labels['namespace_id'] = match_data['namespace_name']
           group_resource.labels['pod_id'] = match_data['pod_name']
+          %w(namespace_name pod_name).each do |field|
+            group_common_labels["#{CONTAINER_CONSTANTS[:service]}/#{field}"] =
+              match_data[field]
+          end
         end
       end
 
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -270,13 +270,16 @@ module BaseTest
     }
   }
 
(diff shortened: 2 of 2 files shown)


crassirostris:
Thanks, LGTM

test/plugin/base_test.rb
```diff
@@ -270,23 +270,23 @@ def setup
...
         'instance_id' => VM_ID,
+        'pod_id' => CONTAINER_POD_NAME,
         'container_name' => CONTAINER_CONTAINER_NAME,
         'zone' => ZONE
       }
     },
     log_name: CONTAINER_CONTAINER_NAME,
     project_id: PROJECT_ID,
     labels: {
-      "#{CONTAINER_CONSTANTS[:service]}/namespace_name" =>
```
qingling128: Is removing these two labels going to break anything?

igorpeshansky: That's exactly the question that @summit was planning to answer.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -450,12 +450,9 @@ def compute_group_resource_and_labels(tag)
...
         if match_data
-          group_resource.labels['container_name'] =
-            match_data['container_name']
-          %w(namespace_name pod_name).each do |field|
-            group_common_labels["#{CONTAINER_CONSTANTS[:service]}/#{field}"] =
-              match_data[field]
-          end
+          group_resource.labels['container_name'] = match_data['container_name']
+          group_resource.labels['namespace_id'] = match_data['namespace_name']
+          group_resource.labels['pod_id'] = match_data['pod_name']
```
qingling128: BTW, if `pod_id` and `namespace_id` are present in the `kubernetes` field in the log entry, the values of resource labels `pod_id` and `namespace_id` here would be overwritten. Do we want that? Sounds like the `kubernetes` field is to be deprecated or something (I just read the original ticket).

crassirostris: > Sounds like the `kubernetes` field is to be deprecated or something

It's not used in the fluentd for GCP deployment, yes

qingling128: @igorpeshansky: Should we clean it up then? A follow up ticket or TODO is fine.

igorpeshansky: @qingling128: That would involve deprecating (support for) the [kubernetes_metadata_filter](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter) plugin. I would rather tackle that separately. As things stand, the `container_name` field from the `kubernetes` record will have been overwritten by this code already -- we're just adding a couple more.

qingling128: Ah, good to know about the filter plugin.

For the overwriting part, I think it's the other way around though. Data extracted from `tag` is set when processing group label level, while data extracted from `kubernetes` record is set later on when processing entry level labels. But I guess the order doesn't really matter If a user does decide to specify different values in different places...

qingling128:
LGTM

igorpeshansky:
Waiting for @summit to confirm that deleting the labels is harmless.

igorpeshansky:
Ping @summit.

igorpeshansky:
As discussed with @summit, we need to leave the labels around as users are actually filtering on them. We'll deprecate and remove them later.

test/plugin/base_test.rb
igorpeshansky: You were right -- @summit found some uses of these, so we're leaving them in. Pushed a new commit. PTAL.

igorpeshansky:
Thanks all. Merging.

PR #132, opened 2017-05-31
Add Google Cloud Dataproc support.

jmikula:
Adding Cloud Dataproc support directly to plugin (currently Cloud Dataproc image uses a forked version of the plugin).

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -318,14 +318,6 @@ module Fluent
             @resource.type = DATAFLOW_CONSTANTS[:resource_type]
           elsif @subservice_name == ML_CONSTANTS[:service]
             @resource.type = ML_CONSTANTS[:resource_type]
-          elsif @subservice_name = DATAPROC_CONSTANTS[:service]
-            @resource.type = DATAPROC_CONSTANTS[:resource_type]
-            @resource.labels['cluster_uuid'] =
-              fetch_gce_metadata('instance/attributes/dataproc-cluster-uuid')
-            @resource.labels['cluster_name'] =
-              fetch_gce_metadata('instance/attributes/dataproc-cluster-name')
-            @resource.labels['region'] =
-              fetch_gce_metadata('instance/attributes/dataproc-region')
           end
         elsif @detect_subservice
           # Check for specialized GCE environments.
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -314,6 +318,14 @@ def configure(conf)
             @resource.type = DATAFLOW_CONSTANTS[:resource_type]
           elsif @subservice_name == ML_CONSTANTS[:service]
             @resource.type = ML_CONSTANTS[:resource_type]
+          elsif @subservice_name = DATAPROC_CONSTANTS[:service]
```
igorpeshansky: Why do both? If you are automatically detecting Dataproc, you shouldn't need this block.

igorpeshansky: There's also a bug here:
```
lib/fluent/plugin/out_google_cloud.rb:321:34: W: Assignment in condition - you probably meant to use ==.
          elsif @subservice_name = DATAPROC_CONSTANTS[:service]
```
but since you're removing this block, the bug is not really relevant, right?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -339,6 +351,16 @@ def configure(conf)
             @resource.labels['cluster_name'] =
               cluster_name_from_kube_env(@kube_env)
             detect_cloudfunctions(attributes)
+          elsif attributes.include?('dataproc-cluster-uuid') &&
```
qingling128: Some unit tests for this as well? Should be similar to `test_one_container_log_metadata_from_plugin`.

jmikula: Added a unit test. PTAL

lib/fluent/plugin/out_google_cloud.rb
jmikula: Removed this block (and bug) as suggested.

jmikula:
Addressed comments. PTAL

test/plugin/base_test.rb
```diff
@@ -1003,6 +1003,17 @@ def test_http_request_skip_setting_latency
     end
   end
 
+  def test_dataproc_log
```
igorpeshansky: Can you please move this right after Cloud Functions​ tests (above `test_http_request_from_record`)?

jmikula: Done. PTAL

qingling128:
LGTM. Please wait for Igor's approval as well.

PR #133, opened 2017-06-05
Refactoring and cleaning up.

qingling128:
* Remove unnecessary instance variables: `@running_on_managed_vm`,
  `@gae_backend_name` and `@gae_backend_version`.
* Refactor the `configure` function to improve readability by separating
  different sections into smaller functions.
* Separate the logic to set labels for different monitored resources to improve
  readability and maintainability.
* Add comments for confusing logic in the code and explain the story behind them
  to improve readability and also serve as a reminder that this might be
  something we could revisit.
* Combine tests.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -264,79 +264,34 @@ module Fluent
                extra.join(' ')
       end
 
-      # TODO: Send instance tags as labels as well?
-      @common_labels = {}
-      @common_labels.merge!(@labels) if @labels
+      # Set regex patterns that is used to parse tags and logs.
+      set_regexp_patterns
 
-      # TODO: Construct Google::Api::MonitoredResource when @use_grpc is
-      # true after the protobuf map corruption issue is fixed.
+      @platform = detect_platform
+
+      # EC2 Metadata server returns everything in one call. Store it to avoid
+      # making multiple calls.
+      @ec2_metadata = fetch_ec2_metadata if @platform == Platform::EC2
+
+      # Set required variables: @project_id, @vm_id, @vm_name and @zone by
+      # making some requests to metadata server.
+      #
+      # Note: Once we support metadata injection at Logging API side, we might
+      # no longer need to require all these metadata in logging agent. But for
+      # now, they are still required.
+      #
+      # TODO(qingling128) After Metadata Agent support is added, try extracting
+      # these info from responses from Metadata Agent first.
+      set_required_metadata_variables
+
+      # TODO(qingling128): Construct Google::Api::MonitoredResource when
+      # @use_grpc is true after the protobuf map corruption issue is fixed.
       @resource = Google::Apis::LoggingV2beta1::MonitoredResource.new(
         labels: {})
 
-      @compiled_kubernetes_tag_regexp = nil
-      if @kubernetes_tag_regexp
-        @compiled_kubernetes_tag_regexp = Regexp.new(@kubernetes_tag_regexp)
-      end
-
-      @cloudfunctions_tag_regexp =
-        /\.(?<encoded_function_name>.+)\.\d+-[^-]+_default_worker$/
-      @cloudfunctions_log_regexp = /^
-        (?:\[(?<severity>.)\])?
-        \[(?<timestamp>.{24})\]
-        (?:\[(?<execution_id>[^\]]+)\])?
-        [ ](?<text>.*)$/x
-
-      @http_latency_regexp = /^\s*(?<seconds>\d+)(?<decimal>\.\d+)?\s*s\s*$/
-
-      # set attributes from metadata (unless overriden by static config)
-      @vm_name = Socket.gethostname if @vm_name.nil?
-      @platform = detect_platform
-      case @platform
-      when Platform::GCE
-        if @project_id.nil?
-          @project_id = fetch_gce_metadata('project/project-id')
-        end
-        if @zone.nil?
-          # this returns "projects/<number>/zones/<zone>"; we only want
-          # the part after the final slash.
-          fully_qualified_zone = fetch_gce_metadata('instance/zone')
-          @zone = fully_qualified_zone.rpartition('/')[2]
-        end
-        @vm_id = fetch_gce_metadata('instance/id') if @vm_id.nil?
-      when Platform::EC2
-        metadata = fetch_ec2_metadata
-        if @zone.nil? && metadata.key?('availabilityZone')
-          @zone = 'aws:' + metadata['availabilityZone']
-        end
-        if @vm_id.nil? && metadata.key?('instanceId')
-          @vm_id = metadata['instanceId']
-        end
-        if metadata.key?('accountId')
-          @resource.labels['aws_account'] = metadata['accountId']
-        end
-      when Platform::OTHER
-        # do nothing
-      else
-        fail Fluent::ConfigError, 'Unknown platform ' + @platform
-      end
-
-      # If we still don't have a project ID, try to obtain it from the
-      # credentials.
-      if @project_id.nil?
-        @project_id = CredentialsInfo.project_id
-        @log.info 'Set Project ID from credentials: ', @project_id unless
-          @project_id.nil?
-      end
-
-      # all metadata parameters must now be set
-      unless @project_id && @zone && @vm_id
-        missing = []
-        missing << 'project_id' unless @project_id
-        missing << 'zone' unless @zone
-        missing << 'vm_id' unless @vm_id
-        fail Fluent::ConfigError, 'Unable to obtain metadata parameters: ' +
-          missing.join(' ')
-      end
+      # TODO: Send instance tags as labels as well?
+      @common_labels = {}
+      @common_labels.merge!(@labels) if @labels
 
       # Default this to false; it is only overwritten if we detect Cloud
       # Functions.
@@ -402,6 +357,8 @@ module Fluent
         @resource.type = EC2_CONSTANTS[:resource_type]
         @resource.labels['instance_id'] = @vm_id
         @resource.labels['region'] = @zone
+        @resource.labels['aws_account'] = @ec2_metadata['accountId'] if
+          @ec2_metadata.key?('accountId')
         # the aws_account label is populated above.
         common_labels["#{EC2_CONSTANTS[:service]}/resource_name"] = @vm_name
       when Platform::OTHER
@@ -821,6 +778,24 @@ module Fluent
       nil
     end
 
+    # Set regexp patterns to parse tags and logs.
+    def set_regexp_patterns
+      @compiled_kubernetes_tag_regexp = nil
+      if @kubernetes_tag_regexp
+        @compiled_kubernetes_tag_regexp = Regexp.new(@kubernetes_tag_regexp)
+      end
+
+      @cloudfunctions_tag_regexp =
+        /\.(?<encoded_function_name>.+)\.\d+-[^-]+_default_worker$/
+      @cloudfunctions_log_regexp = /^
+        (?:\[(?<severity>.)\])?
+        \[(?<timestamp>.{24})\]
+        (?:\[(?<execution_id>[^\]]+)\])?
+        [ ](?<text>.*)$/x
+
+      @http_latency_regexp = /^\s*(?<seconds>\d+)(?<decimal>\.\d+)?\s*s\s*$/
+    end
+
     # "enum" of Platform values
     module Platform
       OTHER = 0  # Other/unkown platform
@@ -872,6 +847,82 @@ module Fluent
         contents = f.read
         return JSON.parse(contents)
       end
+    end
+
+    # Set required variables like @project_id, @vm_id, @vm_name and @zone.
+    def set_required_metadata_variables
+      set_project_id
+      set_vm_id
+      set_vm_name
+      set_location
+
+      # All metadata parameters must now be set.
+      return if @project_id && @zone && @vm_id
+      missing = []
+      missing << 'project_id' unless @project_id
+      missing << 'zone' unless @zone
+      missing << 'vm_id' unless @vm_id
+      fail Fluent::ConfigError, 'Unable to obtain metadata parameters: ' +
+        missing.join(' ')
+    end
+
+    def set_project_id
+      # 1. Check if the value is explicitly in the config thus already set.
+      return unless @project_id.nil?
+      # 2. Try to retrieve it by calling metadata servers directly.
+      @project_id = fetch_gce_metadata('project/project-id') if
+        @platform == Platform::GCE
+      # 3. Try to obtain it from the credentials.
+      @project_id ||= CredentialsInfo.project_id
+    rescue StandardError => e
+      @log.debug 'Failed to obtain project id: ', error: e
+    end
+
+    def set_vm_id
+      # 1. Check if the value is explicitly in the config thus already set.
+      return unless @vm_id.nil?
+      # 2. Check if the response from Metadata Agent includes this info.
+      @vm_id = @resource.labels['instance_id'] if
+        !@resource.nil? && @resource.labels.key?('instance_id')
+      # 3. Try to retrieve it by calling metadata servers directly.
+      @vm_id ||= fetch_gce_metadata('instance/id') if @platform == Platform::GCE
+      @vm_id ||= @ec2_metadata['instanceId'] if @platform == Platform::EC2
+    rescue StandardError => e
+      @log.debug 'Failed to obtain vm_id: ', error: e
+    end
+
+    def set_vm_name
+      # 1. Check if the value is explicitly in the config thus already set.
+      return unless @vm_name.nil?
+      # 2. Check if the response from Metadata Agent includes this info.
+      @vm_name ||= @resource.labels['instance_name'] if
+        !@resource.nil? && @resource.labels.key?('instance_name')
+      # 3. Detect it locally.
+      @vm_name ||= Socket.gethostname
+    rescue StandardError => e
+      @log.debug 'Failed to obtain vm name: ', error: e
+    end
+
+    def set_location
+      # 1. Check if the value is explicitly in the config thus already set.
+      return unless @zone.nil?
+      # 2. Check if the response from Metadata Agent includes this info.
+      unless @resource.nil?
+        @zone ||= @resource.labels['location'] if
+          @resource.labels.key?('location')
+        @zone ||= @resource.labels['zone'] if
+          @platform == Platform::GCE && @resource.labels.key?('zone')
+        @zone ||= @resource.labels['region'] if
+          @platform == Platform::EC2 && @resource.labels.key?('region')
+      end
+      # 3. Detect it locally.
+      # Response format: "projects/<number>/zones/<zone>"
+      @zone ||= fetch_gce_metadata('instance/zone').rpartition('/')[2] if
+        @platform == Platform::GCE
+      @zone ||= 'aws:' + @ec2_metadata['availabilityZone'] if
+        @platform == Platform::EC2 && @ec2_metadata.key?('availabilityZone')
+    rescue StandardError => e
+      @log.debug 'Failed to obtain location: ', error: e
     end
 
     # TODO: This functionality should eventually be available in another
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -218,9 +218,6 @@ module Fluent
     attr_reader :project_id
     attr_reader :zone
     attr_reader :vm_id
-    attr_reader :running_on_managed_vm
-    attr_reader :gae_backend_name
-    attr_reader :gae_backend_version
     attr_reader :resource
     attr_reader :common_labels
 
@@ -341,9 +338,6 @@ module Fluent
           missing.join(' ')
       end
 
-      # Default this to false; it is only overwritten if we detect Managed VM.
-      @running_on_managed_vm = false
-
       # Default this to false; it is only overwritten if we detect Cloud
       # Functions.
       @running_cloudfunctions = false
@@ -368,15 +362,11 @@ module Fluent
           # Do nothing, just don't populate other service's labels.
           if attributes.include?('gae_backend_name') &&
              attributes.include?('gae_backend_version')
-            # Managed VM
-            @running_on_managed_vm = true
-            @gae_backend_name =
-                fetch_gce_metadata('instance/attributes/gae_backend_name')
-            @gae_backend_version =
-                fetch_gce_metadata('instance/attributes/gae_backend_version')
             @resource.type = APPENGINE_CONSTANTS[:resource_type]
-            @resource.labels['module_id'] = @gae_backend_name
-            @resource.labels['version_id'] = @gae_backend_version
+            @resource.labels['module_id'] = fetch_gce_metadata(
+              'instance/attributes/gae_backend_name')
+            @resource.labels['version_id'] = fetch_gce_metadata(
+              'instance/attributes/gae_backend_version')
           elsif attributes.include?('kube-env')
             # Kubernetes/Container Engine
             @resource.type = CONTAINER_CONSTANTS[:resource_type]
@@ -1316,7 +1306,7 @@ module Fluent
     def log_name(tag, resource)
       if resource.type == CLOUDFUNCTIONS_CONSTANTS[:resource_type]
         tag = 'cloud-functions'
-      elsif @running_on_managed_vm
+      elsif resource.type == APPENGINE_CONSTANTS[:resource_type]
         # Add a prefix to Managed VM logs to prevent namespace collisions.
         tag = "#{APPENGINE_CONSTANTS[:service]}/#{tag}"
       elsif resource.type == CONTAINER_CONSTANTS[:resource_type]
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -54,11 +54,14 @@ module Fluent
       }
       CONTAINER_CONSTANTS = {
         service: 'container.googleapis.com',
-        resource_type: 'container'
+        resource_type: 'container',
+        extra_resource_labels: %w(namespace_id pod_id container_name),
+        extra_common_labels: %w(namespace_name pod_name)
       }
       DATAFLOW_CONSTANTS = {
         service: 'dataflow.googleapis.com',
-        resource_type: 'dataflow_step'
+        resource_type: 'dataflow_step',
+        extra_common_labels: %w(region job_name job_id step_id)
       }
       DATAPROC_CONSTANTS = {
         service: 'cluster.dataproc.googleapis.com',
@@ -70,8 +73,10 @@ module Fluent
       }
       ML_CONSTANTS = {
         service: 'ml.googleapis.com',
-        resource_type: 'ml_job'
+        resource_type: 'ml_job',
+        extra_common_labels: %w(job_id task_name)
       }
+      SUB_SERVICE_LIST = [DATAFLOW_CONSTANTS, ML_CONSTANTS]
 
       # Default value for trace_key config param to set "trace" LogEntry field.
       DEFAULT_TRACE_KEY = 'logging.googleapis.com/trace'
@@ -275,31 +280,36 @@ module Fluent
       # no longer need to require all these metadata in logging agent. But for
       # now, they are still required.
       #
-      # TODO(qingling128) After Metadata Agent support is added, try extracting
+      # TODO(qingling128): After Metadata Agent support is added, try extracting
       # these info from responses from Metadata Agent first.
       set_required_metadata_variables
 
       # Retrieve monitored resource.
       #
-      # TODO(qingling128) After Metadata Agent support is added, try retrieving
+      # TODO(qingling128): After Metadata Agent support is added, try retrieving
       # the monitored resource from Metadata Agent first.
       @resource = determine_agent_level_monitored_resource_via_legacy
 
-      # Set variables specific to CLoud Functions. This has to be called after
-      # we have determined the resource type. The purpose is to avoid repeated
-      # calls to metadata server.
-      @running_cloudfunctions = false
-      # We only support Cloud Functions logs for GKE right now.
-      if @resource.type == CONTAINER_CONSTANTS[:resource_type] &&
-         fetch_gce_metadata('instance/attributes/').split.include?('gcf_region')
-        # We are not setting resource type as Cloud Functions here because
-        # whether a log entry is truly coming from a Cloud Functions function
-        # depends on the log tag. Only when @running_cloudfunctions is true will
-        # we try to match log tags against Cloud Functions tag regexp when
-        # processing log entries.
-        @running_cloudfunctions = true
-        # Fetch this info and store it to avoid recurring metadata server calls.
-        @gcf_region = fetch_gce_metadata('instance/attributes/gcf_region')
+      # Set regexp that we should match tags against later on. Using a list
+      # instead of a map to ensure order. For example, tags will be matched
+      # against Cloud Functions first, then GKE.
+      @tag_regexp_list = []
+      if @resource.type == CONTAINER_CONSTANTS[:resource_type]
+        # We only support Cloud Functions logs for GKE right now.
+        if fetch_gce_metadata('instance/attributes/'
+                             ).split.include?('gcf_region')
+          # Fetch this info and store it to avoid recurring
+          # metadata server calls.
+          @gcf_region = fetch_gce_metadata('instance/attributes/gcf_region')
+          @tag_regexp_list << {
+            tag_regexp: @cloudfunctions_tag_regexp,
+            derived_type: CLOUDFUNCTIONS_CONSTANTS[:resource_type]
+          }
+        end
+        @tag_regexp_list << {
+          tag_regexp: @compiled_kubernetes_tag_regexp,
+          derived_type: CONTAINER_CONSTANTS[:resource_type]
+        }
       end
 
       # Determine the common labels that should be added to all log entries
@@ -651,9 +661,8 @@ module Fluent
     # Set regexp patterns to parse tags and logs.
     def set_regexp_patterns
       @compiled_kubernetes_tag_regexp = nil
-      if @kubernetes_tag_regexp
-        @compiled_kubernetes_tag_regexp = Regexp.new(@kubernetes_tag_regexp)
-      end
+      @compiled_kubernetes_tag_regexp = Regexp.new(@kubernetes_tag_regexp) if
+        @kubernetes_tag_regexp
 
       @cloudfunctions_tag_regexp =
         /\.(?<encoded_function_name>.+)\.\d+-[^-]+_default_worker$/
@@ -674,11 +683,11 @@ module Fluent
       set_location
 
       # All metadata parameters must now be set.
-      return if @project_id && @zone && @vm_id
       missing = []
       missing << 'project_id' unless @project_id
       missing << 'zone' unless @zone
       missing << 'vm_id' unless @vm_id
+      return if missing.empty?
       fail Fluent::ConfigError, 'Unable to obtain metadata parameters: ' +
         missing.join(' ')
     end
@@ -738,108 +747,106 @@ module Fluent
 
     # Determine agent level monitored resource type.
     def determine_agent_level_monitored_resource_type
-      # EC2 instance.
-      return EC2_CONSTANTS[:resource_type] if
-        @platform == Platform::EC2
+      case @platform
+      when Platform::OTHER
+        # Unknown platform will be defaulted to GCE instance.
+        return COMPUTE_CONSTANTS[:resource_type]
 
-      # Unknown platform will be defaulted to GCE instance..
-      return COMPUTE_CONSTANTS[:resource_type] if
-        @platform == Platform::OTHER
+      when Platform::EC2
+        return EC2_CONSTANTS[:resource_type]
 
-      # Resource types determined by @subservice_name config.
-      # Cloud Dataflow.
-      return DATAFLOW_CONSTANTS[:resource_type] if
-        @subservice_name == DATAFLOW_CONSTANTS[:service]
-      # Cloud ML.
-      return ML_CONSTANTS[:resource_type] if
-        @subservice_name == ML_CONSTANTS[:service]
-      # Default back to GCE if invalid value is detected.
-      return COMPUTE_CONSTANTS[:resource_type] if
-        @subservice_name
-
-      # Resource types determined by @detect_subservice config.
-      if @detect_subservice
-        begin
-          attributes = fetch_gce_metadata('instance/attributes/').split
-        rescue StandardError => e
-          @log.error 'Failed to detect subservice: ', error: e
+      when Platform::GCE
+        # Resource types determined by @subservice_name config.
+        if @subservice_name
+          SUB_SERVICE_LIST.each do |resource_constants|
+            return resource_constants[:resource_type] if
+              @subservice_name == resource_constants[:service]
+          end
+          # Default back to GCE if invalid value is detected.
+          return COMPUTE_CONSTANTS[:resource_type]
         end
-        # GAE app.
-        return APPENGINE_CONSTANTS[:resource_type] if
-          attributes.include?('gae_backend_name') &&
-          attributes.include?('gae_backend_version')
-        # GKE container.
-        return CONTAINER_CONSTANTS[:resource_type] if
-          attributes.include?('kube-env')
-        # Cloud Dataproc.
-        return DATAPROC_CONSTANTS[:resource_type] if
-          attributes.include?('dataproc-cluster-uuid') &&
-          attributes.include?('dataproc-cluster-name')
+
+        # Resource types determined by @detect_subservice config.
+        if @detect_subservice
+          begin
+            attributes = fetch_gce_metadata('instance/attributes/').split
+          rescue StandardError => e
+            @log.error 'Failed to detect subservice: ', error: e
+          end
+          # GAE app.
+          return APPENGINE_CONSTANTS[:resource_type] if
+            attributes.include?('gae_backend_name') &&
+            attributes.include?('gae_backend_version')
+          # GKE container.
+          return CONTAINER_CONSTANTS[:resource_type] if
+            attributes.include?('kube-env')
+          # Cloud Dataproc.
+          return DATAPROC_CONSTANTS[:resource_type] if
+            attributes.include?('dataproc-cluster-uuid') &&
+            attributes.include?('dataproc-cluster-name')
+        end
+
+        # GCE instance.
+        return COMPUTE_CONSTANTS[:resource_type]
       end
-      # GCE instance.
-      COMPUTE_CONSTANTS[:resource_type]
     end
 
     # Determine agent level monitored resource labels based on the resource
     # type. Each resource type has its own labels that need to be filled in.
     def determine_agent_level_monitored_resource_labels(type)
-      labels = {}
-
       case type
-
       # GAE app.
       when APPENGINE_CONSTANTS[:resource_type]
-        begin
-          labels['module_id'] = fetch_gce_metadata(
-            'instance/attributes/gae_backend_name')
-          labels['version_id'] = fetch_gce_metadata(
+        labels = {
+          'module_id' => fetch_gce_metadata(
+            'instance/attributes/gae_backend_name'),
+          'version_id' => fetch_gce_metadata(
             'instance/attributes/gae_backend_version')
-        rescue StandardError => e
-          @log.error 'Failed to set monitored resource labels for GAE: ',
-                     error: e
-        end
+        }
 
       # GCE.
       when COMPUTE_CONSTANTS[:resource_type]
-        labels['instance_id'] = @vm_id
-        labels['zone'] = @zone
+        labels = {
+          'instance_id' => @vm_id,
+          'zone' => @zone
+        }
 
       # GKE container.
       when CONTAINER_CONSTANTS[:resource_type]
-        labels['instance_id'] = @vm_id
-        labels['zone'] = @zone
-        begin
-          raw_kube_env = fetch_gce_metadata('instance/attributes/kube-env')
-          kube_env = YAML.load(raw_kube_env)
-          labels['cluster_name'] =
-            cluster_name_from_kube_env(kube_env)
-        rescue StandardError => e
-          @log.error 'Failed to set monitored resource labels for GKE: ',
-                     error: e
-        end
+        raw_kube_env = fetch_gce_metadata('instance/attributes/kube-env')
+        kube_env = YAML.load(raw_kube_env)
+        labels = {
+          'instance_id' => @vm_id,
+          'zone' => @zone,
+          'cluster_name' => cluster_name_from_kube_env(kube_env)
+        }
 
       # Cloud Dataproc.
       when DATAPROC_CONSTANTS[:resource_type]
-        begin
-          labels['cluster_uuid'] =
-            fetch_gce_metadata('instance/attributes/dataproc-cluster-uuid')
-          labels['cluster_name'] =
-            fetch_gce_metadata('instance/attributes/dataproc-cluster-name')
-          labels['region'] =
+        labels = {
+          'cluster_uuid' =>
+            fetch_gce_metadata('instance/attributes/dataproc-cluster-uuid'),
+          'cluster_name' =>
+            fetch_gce_metadata('instance/attributes/dataproc-cluster-name'),
+          'region' =>
             fetch_gce_metadata('instance/attributes/dataproc-region')
-        rescue StandardError => e
-          @log.error 'Failed to set monitored resource labels for Cloud ' \
-                     'Dataproc: ', error: e
-        end
+        }
 
       # EC2.
       when EC2_CONSTANTS[:resource_type]
-        labels['instance_id'] = @vm_id
-        labels['region'] = @zone
+        labels = {
+          'instance_id' => @vm_id,
+          'region' => @zone
+        }
         labels['aws_account'] = ec2_metadata['accountId'] if
           ec2_metadata.key?('accountId')
       end
-      labels
+
+      labels ||= {}
+      return labels
+    rescue StandardError => e
+      @log.error "Failed to set monitored resource labels for #{type}: ",
+                 error: e
     end
 
     # Determine the common labels that should be added to all log entries
@@ -847,38 +854,30 @@ module Fluent
     def determine_agent_level_common_labels
       labels = {}
       # User can specify labels via config. We want to capture those as well.
-      # TODO: Send instance tags as labels as well?
       labels.merge!(@labels) if @labels
 
       case @resource.type
+      # GAE, Cloud Dataflow, Cloud Dataproc and Cloud ML.
+      when APPENGINE_CONSTANTS[:resource_type],
+           DATAFLOW_CONSTANTS[:resource_type],
+           DATAPROC_CONSTANTS[:resource_type],
+           ML_CONSTANTS[:resource_type]
+        labels.merge!(
+          "#{COMPUTE_CONSTANTS[:service]}/resource_id" => @vm_id,
+          "#{COMPUTE_CONSTANTS[:service]}/resource_name" => @vm_name,
+          "#{COMPUTE_CONSTANTS[:service]}/zone" => @zone
+        )
 
-      # GAE app.
-      when APPENGINE_CONSTANTS[:resource_type]
-        labels["#{COMPUTE_CONSTANTS[:service]}/resource_id"] = @vm_id
-        labels["#{COMPUTE_CONSTANTS[:service]}/resource_name"] = @vm_name
-        labels["#{COMPUTE_CONSTANTS[:service]}/zone"] = @zone
-
-      # GCE and GKE container.
+      # GCE instance and GKE container.
       when COMPUTE_CONSTANTS[:resource_type],
            CONTAINER_CONSTANTS[:resource_type]
-        labels["#{COMPUTE_CONSTANTS[:service]}/resource_name"] = @vm_name
-
-      # Cloud Dataflow and Cloud Dataproc.
-      when DATAFLOW_CONSTANTS[:resource_type],
-           DATAPROC_CONSTANTS[:resource_type]
-        labels["#{COMPUTE_CONSTANTS[:service]}/resource_id"] = @vm_id
-        labels["#{COMPUTE_CONSTANTS[:service]}/resource_name"] = @vm_name
-        labels["#{COMPUTE_CONSTANTS[:service]}/zone"] = @zone
+        labels.merge!(
+          "#{COMPUTE_CONSTANTS[:service]}/resource_name" => @vm_name)
 
       # EC2.
       when EC2_CONSTANTS[:resource_type]
-        labels["#{EC2_CONSTANTS[:service]}/resource_name"] = @vm_name
-
-      # Cloud ML.
-      when ML_CONSTANTS[:resource_type]
-        labels["#{COMPUTE_CONSTANTS[:service]}/resource_id"] = @vm_id
-        labels["#{COMPUTE_CONSTANTS[:service]}/resource_name"] = @vm_name
-        labels["#{COMPUTE_CONSTANTS[:service]}/zone"] = @zone
+        labels.merge!(
+          "#{EC2_CONSTANTS[:service]}/resource_name" => @vm_name)
       end
       labels
     end
@@ -893,8 +892,8 @@ module Fluent
         determine_group_level_monitored_resource_type(tag)
 
       # Determine group level monitored resource labels and common labels.
-      group_resource_labels, group_common_labels = determine_group_level_labels(
-        group_resource_type, matched_regex_group)
+      group_resource_labels, group_common_labels =
+        determine_group_level_labels(group_resource_type, matched_regex_group)
 
       group_resource = Google::Apis::LoggingV2beta1::MonitoredResource.new(
         type: group_resource_type,
@@ -912,71 +911,64 @@ module Fluent
 
     # Determine group level monitored resource type shared by a collection of
     # entries.
-    # Returns the resource type and tag regexp matched groups. The matched
-    # groups only apply to some resource types. Return nil if not applicable or
-    # if there is no match.
+    # Return the resource type and tag regexp matched groups. The matched groups
+    # only apply to some resource types. Return nil if not applicable or if
+    # there is no match.
     def determine_group_level_monitored_resource_type(tag)
-      # Match tag against Cloud Functions format.
-      if @running_cloudfunctions
-        matched_regex_group = @cloudfunctions_tag_regexp.match(tag)
-        return [CLOUDFUNCTIONS_CONSTANTS[:resource_type], matched_regex_group] \
-          if matched_regex_group
+      @tag_regexp_list.each do |rule_pair|
+        matched_regex_group = rule_pair[:tag_regexp].match(tag)
+        return [rule_pair[:derived_type], matched_regex_group] if
+          matched_regex_group
       end
-
-      # Match tag against GKE Container format.
-      if @resource.type == CONTAINER_CONSTANTS[:resource_type] &&
-         @compiled_kubernetes_tag_regexp
-        # Container logs in Kubernetes are tagged based on where they came from,
-        # so we can extract useful metadata from the tag. Do this here to avoid
-        # having to repeat it for each record.
-        matched_regex_group = @compiled_kubernetes_tag_regexp.match(tag)
-        return [@resource.type, matched_regex_group] if matched_regex_group
-      end
-
-      # Otherwise, return the original type.
       [@resource.type, nil]
     end
 
     # Determine group level monitored resource labels and common labels. These
     # labels will be shared by a collection of entries.
-    def determine_group_level_labels(group_resource_type,
-                                     matched_regex_group)
+    def determine_group_level_labels(group_resource_type, matched_regex_group)
       group_resource_labels = @resource.labels.dup
       group_common_labels = @common_labels.dup
 
       case group_resource_type
-
       # Cloud Functions.
       when CLOUDFUNCTIONS_CONSTANTS[:resource_type]
-        group_resource_labels['region'] = @gcf_region
-        group_resource_labels['function_name'] =
-          decode_cloudfunctions_function_name(
+        group_resource_labels.merge!(
+          'region' => @gcf_region,
+          'function_name' => decode_cloudfunctions_function_name(
             matched_regex_group['encoded_function_name'])
+        )
+
         instance_id = group_resource_labels.delete('instance_id')
-        group_common_labels["#{CONTAINER_CONSTANTS[:service]}/cluster_name"] =
-          group_resource_labels.delete('cluster_name')
-        group_common_labels["#{CONTAINER_CONSTANTS[:service]}/instance_id"] =
-          instance_id
-        group_common_labels["#{COMPUTE_CONSTANTS[:service]}/resource_id"] =
-          instance_id
-        group_common_labels["#{COMPUTE_CONSTANTS[:service]}/zone"] =
-          group_resource_labels.delete('zone')
+        group_common_labels.merge!(
+          "#{CONTAINER_CONSTANTS[:service]}/instance_id" => instance_id,
+          "#{COMPUTE_CONSTANTS[:service]}/resource_id" => instance_id,
+          "#{CONTAINER_CONSTANTS[:service]}/cluster_name" =>
+            group_resource_labels.delete('cluster_name'),
+          "#{COMPUTE_CONSTANTS[:service]}/zone" =>
+            group_resource_labels.delete('zone')
+        )
 
       # GKE container.
       when CONTAINER_CONSTANTS[:resource_type]
-        if matched_regex_group
-          group_resource_labels['container_name'] =
-            matched_regex_group['container_name']
-          # The kubernetes_tag_regexp is poorly named. 'namespace_name' is in
-          # fact 'namespace_id'. 'pod_name' is in fact 'pod_id'.
-          group_resource_labels['namespace_id'] =
-            matched_regex_group['namespace_name']
-          group_resource_labels['pod_id'] =
-            matched_regex_group['pod_name']
-          %w(namespace_name pod_name).each do |field|
-            group_common_labels["#{CONTAINER_CONSTANTS[:service]}/#{field}"] =
-              matched_regex_group[field]
-          end
+        unless matched_regex_group.nil?
+          matched_regex_group_hash = Hash[
+            matched_regex_group.names.zip(matched_regex_group.captures)]
+          group_resource_labels.merge!(
+            delete_and_extract_labels(
+              matched_regex_group_hash, {
+                # The kubernetes_tag_regexp is poorly named. 'namespace_name' is
+                # in fact 'namespace_id'. 'pod_name' is in fact 'pod_id'.
+                'container_name' => 'container_name',
+                'namespace_name' => 'namespace_id',
+                'pod_name' => 'pod_id'
+              }, false))
+
+          group_common_labels.merge!(
+            delete_and_extract_labels(
+              matched_regex_group_hash,
+              CONTAINER_CONSTANTS[:extra_common_labels]
+                .map { |l| [l, "#{CONTAINER_CONSTANTS[:service]}/#{l}"] }.to_h,
+              false))
         end
       end
 
@@ -1013,17 +1005,18 @@ module Fluent
         if record.key?('kubernetes')
           resource_labels.merge!(
             delete_and_extract_labels(
-              record['kubernetes'], %w(namespace_id pod_id container_name)
+              record['kubernetes'], CONTAINER_CONSTANTS[:extra_resource_labels]
                 .map { |l| [l, l] }.to_h))
           common_labels.merge!(
             delete_and_extract_labels(
-              record['kubernetes'], %w(namespace_name pod_name)
+              record['kubernetes'], CONTAINER_CONSTANTS[:extra_common_labels]
                 .map { |l| [l, "#{CONTAINER_CONSTANTS[:service]}/#{l}"] }.to_h))
           # Prepend label/ to all user-defined labels' keys.
           if record['kubernetes'].key?('labels')
-            record['kubernetes']['labels'].each do |key, value|
-              common_labels["label/#{key}"] = value
-            end
+            common_labels.merge!(
+              delete_and_extract_labels(
+                record['kubernetes']['labels'], record['kubernetes']['labels']
+                  .map { |key, _| [key, "label/#{key}"] }.to_h))
           end
           # We've explicitly consumed all the fields we care about -- don't
           # litter the log entries with the remaining fields that the kubernetes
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -36,8 +36,10 @@ class Map
 module Fluent
   # fluentd output plugin for the Stackdriver Logging API
   class GoogleCloudOutput < BufferedOutput
-    # Constants for service names and resource types.
+    # Constants used by the plugin and its tests.
```
igorpeshansky: s/used/shared/? But I fail to see the reason behind this change...
How about renaming this module to `ServiceConstants` and leaving the comment as-is?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -207,6 +206,24 @@ def initialize
       @log = $log # rubocop:disable Style/GlobalVars
     end
 
+    # Set up regexp patterns to parse tags and logs.
+    def setup_regexp_patterns
```
igorpeshansky: Why did you add this before `configure`, when all of the other new functions are after? Does this function even need to be public?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -489,12 +431,14 @@ def extract_entry_labels(group_resource, record)
...
+      # Cloud Functions.
       if group_resource.type == CLOUDFUNCTIONS_CONSTANTS[:resource_type] &&
          record.key?('log')
         @cloudfunctions_log_match =
           @cloudfunctions_log_regexp.match(record['log'])
       end
 
+      # GKE containers.
       if group_resource.type == CONTAINER_CONSTANTS[:resource_type]
         # Move the stdout/stderr annotation from the record into a label
```
igorpeshansky: Since you're cleaning up, might as well add the missing trailing "." here...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -225,160 +242,47 @@ def configure(conf)
...
-          @resource.labels['aws_account'] = metadata['accountId']
-        end
-      when Platform::OTHER
-        # do nothing
-      else
-        fail Fluent::ConfigError, 'Unknown platform ' + @platform
-      end
+      # Retrieve monitored resource.
+      #
+      # TODO(qingling128) After Metadata Agent support is added, try retrieving
```
igorpeshansky: Did you mean to include this change?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -225,160 +242,47 @@ def configure(conf)
...
-        @log.info 'Set Project ID from credentials: ', @project_id unless
-          @project_id.nil?
-      end
+      # Set variables specific to CLoud Functions. This has to be called after
+      # we have determined resource type. The purpose is to avoid repeated calls
+      # to metadata server.
+      #
+      # TODO(qingling128) Revisit this after Metadata Agent supports Cloud
+      # Functions or after we support Cloud Functions in more platforms
+      # (Right now we only supports GKE).
```
igorpeshansky: Nit: s/supports/support/.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -225,160 +242,47 @@ def configure(conf)
...
-      if @project_id.nil?
-        @project_id = CredentialsInfo.project_id
-        @log.info 'Set Project ID from credentials: ', @project_id unless
-          @project_id.nil?
-      end
+      # Set variables specific to CLoud Functions. This has to be called after
+      # we have determined resource type. The purpose is to avoid repeated calls
+      # to metadata server.
+      #
+      # TODO(qingling128) Revisit this after Metadata Agent supports Cloud
```
igorpeshansky: Did you mean to include this change?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -225,160 +242,47 @@ def configure(conf)
...
 
-      # If we still don't have a project ID, try to obtain it from the
-      # credentials.
-      if @project_id.nil?
-        @project_id = CredentialsInfo.project_id
-        @log.info 'Set Project ID from credentials: ', @project_id unless
-          @project_id.nil?
-      end
+      # Set variables specific to CLoud Functions. This has to be called after
+      # we have determined resource type. The purpose is to avoid repeated calls
```
igorpeshansky: Nit: the resource type.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -225,160 +242,47 @@ def configure(conf)
...
+      # to metadata server.
+      #
+      # TODO(qingling128) Revisit this after Metadata Agent supports Cloud
+      # Functions or after we support Cloud Functions in more platforms
+      # (Right now we only supports GKE).
+      set_cloudfunctions_variables
```
igorpeshansky: Can this be folded into another `set_*_variables` function (to be called here)? I'm just wondering why Cloud Functions have to be treated specially.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -225,160 +242,47 @@ def configure(conf)
...
-      when Platform::OTHER
-        # do nothing
-      else
-        fail Fluent::ConfigError, 'Unknown platform ' + @platform
-      end
+      # Retrieve monitored resource.
+      #
+      # TODO(qingling128) After Metadata Agent support is added, try retrieving
+      # the monitored resource from Metadata Agent first.
+      @resource = determine_agent_level_monitored_resource_via_legacy
```
igorpeshansky: Should we combine this with `determine_agent_level_common_labels`? I vaguely recall that you had a function that returned a pair -- what happened to that?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -512,17 +456,20 @@ def extract_entry_labels(group_resource, record)
...
-      # If a field is present in the label_map, send its value as a label
-      # (mapping the field name to label name as specified in the config)
-      # and do not send that field as part of the payload.
+      # If the name of a field in the record is present in the @label_map
+      # configured by users, report its value as a label and do not send that
+      # field as part of the payload.
       common_labels.merge!(fields_to_labels(record, @label_map))
 
+      # Cloud Functions.
+      # TODO TODOTODO
```
igorpeshansky: ???

igorpeshansky:
Just to clarify, I'm only asking for separate commits in this PR, not separate PRs.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Oops, this change is not part of the refactoring. It only makes sense after adding Metadata Agent support. I just reverted it.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Oh, I forgot to move this to where it should belong. Just moved. Also rename it to be `set_regexp_patterns` instead of `setup_regexp_patterns`.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Yeah, I'm trying to give a hint why the naming is `via_legacy`.

lib/fluent/plugin/out_google_cloud.rb
qingling128: I put these in two separate files because `determine_agent_level_monitored_resource_via_legacy` should be equivalent to `determine_agent_level_monitored_resource_via_metadata_agent` which is to be added soon (in the other PR).

The functions that return a pair of `resource_labels` and `common_labels` are at the `group` and `entry` level.

Note that at the `agent` level we don't need to determine the monitored resource type before calling Metadata Agent. But at `group` and `entry` levels we need to determine monitored resource type locally based on tags first, then call Metadata Agent if necessary (only when we detect special resource type).

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Removed.

lib/fluent/plugin/out_google_cloud.rb
qingling128: This comment is removed.

lib/fluent/plugin/out_google_cloud.rb
qingling128: I thought about it. Yet this is the only variable that has to be set after resource_type is determined. Maybe in the future when other resource types need to set certain variables as well, we can group them. Another option is to fold this into `determine_agent_level_common_labels` in a `Cloud Functions` section. But I leaned towards keeping them separate.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Oh, this is a reminder for myself to double check the influence and move this part to be right after the Cloud Functions tag parsing above. Completed the TODO.

qingling128:
I've split the PR into different commits. I did not do a force push because I'm not sure how GitHub will react to that. So right now an ugly revert commit is present. I will make sure the commit message is clean when I merge. When merging, do we want to just squash everything into one single commit or keep different sections in different commits?

qingling128:
PTAL.

qingling128:
Friendly ping.

test/plugin/base_test.rb
```diff
@@ -701,80 +701,33 @@ def test_multiple_managed_vm_logs
...
-        assert_equal CONTAINER_SECONDS_EPOCH, entry['timestamp']['seconds'],
-                     entry
-        assert_equal CONTAINER_NANOS, entry['timestamp']['nanos'], entry
-        assert_equal CONTAINER_SEVERITY, entry['severity'], entry
-      end
-    end
-  end
-
-  def test_multiple_container_logs_metadata_from_tag
+  def test_container_log_metadata
```
qingling128: This has been split into two tests as suggested in the other comment.

test/plugin/base_test.rb
```diff
@@ -701,80 +701,33 @@ def test_multiple_managed_vm_logs
...
-        n.times { |i| d.emit(container_log_entry(log_entry(i))) }
-        d.run
-      end
-      verify_log_entries(n, CONTAINER_FROM_TAG_PARAMS) do |entry|
-        assert_equal CONTAINER_SECONDS_EPOCH, entry['timestamp']['seconds'],
-                     entry
-        assert_equal CONTAINER_NANOS, entry['timestamp']['nanos'], entry
-        assert_equal CONTAINER_SEVERITY, entry['severity'], entry
+    {
+      # Metadata from metadata.
```
qingling128: Just clarified this comment. The metadata is in fact from the record:
```
# Test container logs when metadata is extracted from the 'kubernetes' field
# in the log record.
```

test/plugin/base_test.rb
```diff
@@ -701,80 +701,33 @@ def test_multiple_managed_vm_logs
...
+      method(:container_log_entry_with_metadata) =>
+        CONTAINER_FROM_METADATA_PARAMS,
+      # Metadata from tag.
+      method(:container_log_entry) => CONTAINER_FROM_TAG_PARAMS
+    }.each do |log_entry_method, expected_params|
```
qingling128: Good point. Just refactored this into two methods with shared utility function.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -225,160 +222,43 @@ def configure(conf)
                extra.join(' ')
       end
 
-      # TODO: Send instance tags as labels as well?
-      @common_labels = {}
-      @common_labels.merge!(@labels) if @labels
+      # Set regex patterns that is used to parse tags and logs.
```
qingling128: Changed this one and all other occurrences.

qingling128: Right. Removed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -225,160 +222,43 @@ def configure(conf)
...
-        labels: {})
+      @platform = detect_platform
 
-      @compiled_kubernetes_tag_regexp = nil
-      if @kubernetes_tag_regexp
-        @compiled_kubernetes_tag_regexp = Regexp.new(@kubernetes_tag_regexp)
-      end
+      # EC2 Metadata server returns everything in one call. Store it to avoid
+      # making multiple calls.
+      @ec2_metadata = fetch_ec2_metadata if @platform == Platform::EC2
```
qingling128: Good idea. Changed. BTW, I combined `ec2_metadata` and `fetch_ec2_metadata`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -225,160 +222,43 @@ def configure(conf)
...
-
-      @http_latency_regexp = /^\s*(?<seconds>\d+)(?<decimal>\.\d+)?\s*s\s*$/
+      # Set required variables: @project_id, @vm_id, @vm_name and @zone by
+      # making some requests to metadata server.
+      #
+      # Note: Once we support metadata injection at Logging API side, we might
+      # no longer need to require all these metadata in logging agent. But for
+      # now, they are still required.
+      #
+      # TODO(qingling128) After Metadata Agent support is added, try extracting
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -225,160 +222,43 @@ def configure(conf)
...
+      # these info from responses from Metadata Agent first.
+      set_required_metadata_variables
```
qingling128: Yes. The current workflow with metadata agent support added is:

1. Get monitored resource from metadata agent.
2. Set these required variables: `project_id`, `vm_id`, `vm_name`, `zone`, etc. In this step, we will make use of the monitored resource retrieved from metadata agent if available. Only when a variable is not available from the metadata agent response will we try to retrieve it locally.
3. Execute the legacy monitored resource construction code which will make use of the variables that have been set in step 2.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -764,6 +520,24 @@ def parse_json_or_nil(input)
       nil
     end
 
+    # Set regexp patterns to parse tags and logs.
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -817,6 +591,409 @@ def fetch_ec2_metadata
...
+      missing << 'project_id' unless @project_id
+      missing << 'zone' unless @zone
+      missing << 'vm_id' unless @vm_id
+      fail Fluent::ConfigError, 'Unable to obtain metadata parameters: ' +
+        missing.join(' ')
+    end
+
+    def set_project_id
+      # 1. Check if the value is explicitly in the config thus already set.
+      return unless @project_id.nil?
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -817,6 +591,409 @@ def fetch_ec2_metadata
...
+      # 2. Try to retrieve it by calling metadata servers directly.
+      @project_id = fetch_gce_metadata('project/project-id') if
+        @platform == Platform::GCE
+      # 3. Try to obtain it from the credentials.
+      @project_id ||= CredentialsInfo.project_id
+    rescue StandardError => e
+      @log.debug 'Failed to obtain project id: ', error: e
```
qingling128: Changed this and all other occurrences.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -817,6 +591,409 @@ def fetch_ec2_metadata
...
+    end
+
+    def set_vm_id
+      # 1. Check if the value is explicitly in the config thus already set.
+      return unless @vm_id.nil?
```
qingling128: Changed. Also removed the part to extract the value from `@resource`, which is only meaningful after Metadata agent support is added. I left it here by mistake.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -817,6 +591,409 @@ def fetch_ec2_metadata
...
+    end
+
+    def set_location
+      # 1. Check if the value is explicitly in the config thus already set.
+      return unless @zone.nil?
+      # 2. Check if the response from Metadata Agent includes this info.
+      unless @resource.nil?
+        @zone ||= @resource.labels['location'] if
+          @resource.labels.key?('location')
+        @zone ||= @resource.labels['zone'] if
```
qingling128: Changed. In fact I removed the entire section to extract the value from `@resource`. This is only meaningful after Metadata agent support is added. I left it here by mistake.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -817,6 +591,409 @@ def fetch_ec2_metadata
...
+          @platform == Platform::GCE && @resource.labels.key?('zone')
+        @zone ||= @resource.labels['region'] if
```
qingling128: Changed. In fact I removed the entire section to extract the value from `@resource`. This is only meaningful after Metadata agent support is added. I left it here by mistake.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -858,6 +829,189 @@ def set_location
...
+      # We only support Cloud Functions logs for GKE right now.
+      if @resource.type == CONTAINER_CONSTANTS[:resource_type] &&
+         fetch_gce_metadata('instance/attributes/').split.include?('gcf_region')
+        # We are not setting resource type as Cloud Functions here because
+        # whether a log entry is truly coming from a Cloud Functions function
+        # depends on the log tag.
+
+        # Only when @running_cloudfunctions is true will we try to match log
+        # tags against Cloud Functions tag regex when processing log entries.
+        @running_cloudfunctions = true
```
qingling128: Removed `set_cloudfunctions_variables` and inlined the code.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -858,6 +829,189 @@ def set_location
...
+        # Fetch this info and store it to avoid recurring metadata server calls.
+        @gcf_region = fetch_gce_metadata('instance/attributes/gcf_region')
```
qingling128: ? `def fetch_gce_metadata(metadata_path)`? This method has many occurrences. `fetch_ec2_metadata` has only one occurrence, and I just inlined that.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -817,6 +591,409 @@ def fetch_ec2_metadata
...
+      # Cloud ML.
+      when ML_CONSTANTS[:resource_type]
+        labels["#{COMPUTE_CONSTANTS[:service]}/resource_id"] = @vm_id
+        labels["#{COMPUTE_CONSTANTS[:service]}/resource_name"] = @vm_name
+        labels["#{COMPUTE_CONSTANTS[:service]}/zone"] = @zone
+      end
+      labels
+    end
+
+    # Determin the group level monitored resource and common labels shared by a
```
qingling128: Oops. Changed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -225,160 +222,50 @@ def configure(conf)
...
-        common_labels["#{EC2_CONSTANTS[:service]}/resource_name"] = @vm_name
-      when Platform::OTHER
-        # Use GCE as the default environment.
-        @resource.type = COMPUTE_CONSTANTS[:resource_type]
-        @resource.labels['instance_id'] = @vm_id
-        @resource.labels['zone'] = @zone
-        common_labels["#{COMPUTE_CONSTANTS[:service]}/resource_name"] = @vm_name
+      # We only support Cloud Functions logs for GKE right now.
+      if @resource.type == CONTAINER_CONSTANTS[:resource_type] &&
+         fetch_gce_metadata('instance/attributes/').split.include?('gcf_region')
```
bmoyles0117: Are the values of fetch_gce_metadata memoized? If not, can we memoize them or store a local value and reuse the value in the if statement to avoid fetching data twice?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -806,15 +569,403 @@ def fetch_gce_metadata(metadata_path)
...
+        labels["#{COMPUTE_CONSTANTS[:service]}/resource_id"] = @vm_id
+        labels["#{COMPUTE_CONSTANTS[:service]}/resource_name"] = @vm_name
+        labels["#{COMPUTE_CONSTANTS[:service]}/zone"] = @zone
+
+      # GCE.
+      when COMPUTE_CONSTANTS[:resource_type]
+        labels["#{COMPUTE_CONSTANTS[:service]}/resource_name"] = @vm_name
+
+      # GKE container.
+      when CONTAINER_CONSTANTS[:resource_type]
```
bmoyles0117: Can GKE and GCE be blended into a single conditional like Dataflow and Dataproc are?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -806,15 +569,403 @@ def fetch_gce_metadata(metadata_path)
...
+          group_resource_labels.delete('zone')
+
+      # GKE container.
+      when CONTAINER_CONSTANTS[:resource_type]
+        if matched_regex_group
+          group_resource_labels['container_name'] =
+            matched_regex_group['container_name']
+          group_resource_labels['namespace_id'] =
+            matched_regex_group['namespace_name']
+          group_resource_labels['pod_id'] =
```
bmoyles0117: Should we be consistent with naming across our regex fields and labels?

lib/fluent/plugin/out_google_cloud.rb
qingling128: Unlike AWS, the GCE metadata server does not return everything in one call. Instead we retrieve different pieces of metadata via different urls. And these metadata are only retrieved once at agent start time. So we do not memorize them.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -806,15 +569,403 @@ def fetch_gce_metadata(metadata_path)
...
+          @ec2_metadata = JSON.parse(contents)
+        end
+      end
+
+      @ec2_metadata
+    end
+
+    # Set regexp patterns to parse tags and logs.
+    def set_regexp_patterns
+      @compiled_kubernetes_tag_regexp = nil
```
qingling128: I guess we were trying to differentiate it from the config name `kubernetes_tag_regexp` by prefixing it with `compiled`. At this point, we can not change the config name because that will affect users. `compiled` is not ideal yet kind of makes sense. Any other suggestion?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -806,15 +569,403 @@ def fetch_gce_metadata(metadata_path)
...
+    end
+
+    # 1. Return the value if it is explicitly set in the config already.
+    # 2. If not, try to retrieve it by calling metadata server directly.
+    # 3. If still not set, try to obtain it from the credentials.
+    def set_project_id
+      @project_id ||= fetch_gce_metadata('project/project-id') if
+        @platform == Platform::GCE
+      @project_id ||= CredentialsInfo.project_id
+    rescue StandardError => e
```
qingling128: Yup, that can totally happen. The detailed error and its associated message will be included in the log line though (we pass the error to the logger as well). The only thing we do when there an error occurs is to log it. So I think this should be enough. WDYT?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -806,15 +569,403 @@ def fetch_gce_metadata(metadata_path)
...
+      @log.error 'Failed to obtain project id: ', error: e
+    end
+
+    # 1. Return the value if it is explicitly set in the config already.
+    # 2. If not, try to retrieve it by calling metadata servers directly.
+    def set_vm_id
+      @vm_id ||= fetch_gce_metadata('instance/id') if @platform == Platform::GCE
+      @vm_id ||= ec2_metadata['instanceId'] if @platform == Platform::EC2
+    rescue StandardError => e
```
qingling128: See above.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -806,15 +569,403 @@ def fetch_gce_metadata(metadata_path)
...
+
+    # 1. Return the value if it is explicitly set in the config already.
+    # 2. If not, try to retrieve it locally.
+    def set_location
+      # Response format: "projects/<number>/zones/<zone>"
+      @zone ||= fetch_gce_metadata('instance/zone').rpartition('/')[2] if
+        @platform == Platform::GCE
+      @zone ||= 'aws:' + ec2_metadata['availabilityZone'] if
+        @platform == Platform::EC2 && ec2_metadata.key?('availabilityZone')
+    rescue StandardError => e
```
qingling128: See above.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Sure thing. Just changed this.

lib/fluent/plugin/out_google_cloud.rb
qingling128: `namespace_name` and `pod_name` are implicitly mapped to `namespace_id` and `pod_id` per GKE's request. The `namespace_id` and `pod_id` are poorly named. They are in fact names. Yet it's not possible to change it without affecting users at this moment. See details in b/38411250.

qingling128:
PTAL

qingling128:
Friendly ping.

bmoyles0117:
LGTM

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -306,11 +306,6 @@ def configure(conf)
       # processed by this logging agent.
       @common_labels = determine_agent_level_common_labels
 
-      # For each resource type, there is a list of labels that we want to report
-      # as monitored resource instead of metadata labels. Move them if present.
-      @resource.labels.merge!(
```
qingling128: We are doing the same thing at `agent` level and at `entry` level. The extraction at `agent` level should be redundant because `@label_map` is processed at `entry` level, and we are required to extract these labels after that step anyway. Simply extracting these labels at `entry` level should be sufficient.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1357,14 +1367,16 @@ def sanitize_tag(tag)
     end
 
     # For every original_label => new_label pair in the label_map, delete the
-    # original_label from the record if it exists, and extract the value to form
-    # a map with the new_label as the key.
-    def fields_to_labels(record, label_map)
```
qingling128: Renamed this function to make it more generic. Also make it obvious what is actually being done.

qingling128:
PTAL.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+      end
+      labels
+    end
+
+    # Determine the common labels that should be added to all log entries
+    # processed by this logging agent.
+    def determine_agent_level_common_labels
+      labels = {}
+      # User can specify labels via config. We want to capture those as well.
+      # TODO: Send instance tags as labels as well?
```
igorpeshansky: Now that we've refactored this code, I don't think this TODO makes sense anymore -- we know that the instance labels will be sent via metadata.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+      labels.merge!(@labels) if @labels
+
+      case @resource.type
+
```
igorpeshansky: Same here: this blank line doesn't do anything for readability.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+      # GAE app.
+      when APPENGINE_CONSTANTS[:resource_type]
+        labels["#{COMPUTE_CONSTANTS[:service]}/resource_id"] = @vm_id
```
igorpeshansky: Let's do `labels.merge!({<hash_literal>})` instead, here and below.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+        labels["#{COMPUTE_CONSTANTS[:service]}/resource_name"] = @vm_name
+        labels["#{COMPUTE_CONSTANTS[:service]}/zone"] = @zone
+
+      # EC2.
+      when EC2_CONSTANTS[:resource_type]
+        labels["#{EC2_CONSTANTS[:service]}/resource_name"] = @vm_name
+
+      # Cloud ML.
+      when ML_CONSTANTS[:resource_type]
```
igorpeshansky: This case seems identical to the one for Cloud Dataflow and Cloud Dataproc. Combine?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+    def determine_agent_level_common_labels
+      labels = {}
+      # User can specify labels via config. We want to capture those as well.
+      # TODO: Send instance tags as labels as well?
+      labels.merge!(@labels) if @labels
+
+      case @resource.type
+
+      # GAE app.
+      when APPENGINE_CONSTANTS[:resource_type]
```
igorpeshansky: This case seems identical to the one for Cloud Dataflow and Cloud Dataproc. Combine?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+    # collection of entries.
+    def determine_group_level_monitored_resource_and_labels(tag)
+      # Determine group level monitored resource type. For certain types,
+      # extract useful info from the tag and store those in
+      # matched_regex_group.
+      group_resource_type, matched_regex_group =
+        determine_group_level_monitored_resource_type(tag)
+
+      # Determine group level monitored resource labels and common labels.
+      group_resource_labels, group_common_labels = determine_group_level_labels(
```
igorpeshansky: Let's make this formatting the same as in lines 892-893 above.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+      group_resource.labels.freeze
+      group_common_labels.freeze
+
+      [group_resource, group_common_labels]
+    end
+
+    # Determine group level monitored resource type shared by a collection of
+    # entries.
+    # Returns the resource type and tag regexp matched groups. The matched
+    # groups only apply to some resource types. Return nil if not applicable or
```
igorpeshansky: Consistency: you first say `Returns`, and then `Return` in a later sentence.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+    # if there is no match.
+    def determine_group_level_monitored_resource_type(tag)
+      # Match tag against Cloud Functions format.
+      if @running_cloudfunctions
```
igorpeshansky: It seems like we could set a single `@tag_regexp` value at the beginning to either `@cloudfunctions_tag_regexp` if `@running_cloudfunctions`, or `@compiled_kubernetes_tag_regexp` if the resource is a container resource, and then this method would become simply:
```ruby
  def determine_group_level_monitored_resource_type(tag)
    matched_regex_group = @tag_regexp ? @tag_regexp.match(tag) : nil
    resource_type = @running_cloudfunctions ? CLOUDFUNCTIONS_CONSTANTS[:resource_type]
                                            : @resource.type
    return [resource_type, matched_regex_group]
  end
```
And with the method becoming this simple, you can even inline it at the call site.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+      end
+
+      # Otherwise, return the original type.
+      [@resource.type, nil]
+    end
+
+    # Determine group level monitored resource labels and common labels. These
+    # labels will be shared by a collection of entries.
+    def determine_group_level_labels(group_resource_type,
+                                     matched_regex_group)
```
igorpeshansky: Definitely fits on the previous line...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+        if matched_regex_group
+          group_resource_labels['container_name'] =
+            matched_regex_group['container_name']
+          # The kubernetes_tag_regexp is poorly named. 'namespace_name' is in
+          # fact 'namespace_id'. 'pod_name' is in fact 'pod_id'.
+          group_resource_labels['namespace_id'] =
+            matched_regex_group['namespace_name']
+          group_resource_labels['pod_id'] =
+            matched_regex_group['pod_name']
+          %w(namespace_name pod_name).each do |field|
```
igorpeshansky: This list could also go into `CONTAINER_CONSTANTS`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+    def determine_group_level_labels(group_resource_type,
+                                     matched_regex_group)
+      group_resource_labels = @resource.labels.dup
+      group_common_labels = @common_labels.dup
+
+      case group_resource_type
+
+      # Cloud Functions.
+      when CLOUDFUNCTIONS_CONSTANTS[:resource_type]
+        group_resource_labels['region'] = @gcf_region
```
igorpeshansky: Might be more readable as `<labels_var>.merge!({<hash_literal>})`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+        group_resource_labels['function_name'] =
+          decode_cloudfunctions_function_name(
+            matched_regex_group['encoded_function_name'])
+        instance_id = group_resource_labels.delete('instance_id')
+        group_common_labels["#{CONTAINER_CONSTANTS[:service]}/cluster_name"] =
```
igorpeshansky: This seems reminiscent of `delete_and_extract_labels`...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+          instance_id
+        group_common_labels["#{COMPUTE_CONSTANTS[:service]}/resource_id"] =
+          instance_id
+        group_common_labels["#{COMPUTE_CONSTANTS[:service]}/zone"] =
+          group_resource_labels.delete('zone')
+
+      # GKE container.
+      when CONTAINER_CONSTANTS[:resource_type]
+        if matched_regex_group
+          group_resource_labels['container_name'] =
```
igorpeshansky: If you don't care about destroying `matched_regex_group`, this also seems reminiscent of `delete_and_extract_labels`...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+          delete_and_extract_labels(
+            record, 'stream' => "#{CONTAINER_CONSTANTS[:service]}/stream"))
+
+        # If the record has been annotated by the kubernetes_metadata_filter
+        # plugin, then use that metadata. Otherwise, rely on commonLabels
+        # populated at the grouped_entries level from the group's tag.
+        if record.key?('kubernetes')
+          resource_labels.merge!(
+            delete_and_extract_labels(
+              record['kubernetes'], %w(namespace_id pod_id container_name)
```
igorpeshansky: This list can go into `CONTAINER_CONSTANTS` too (under a different key from the next list :-) ).

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+                .map { |l| [l, l] }.to_h))
+          common_labels.merge!(
+            delete_and_extract_labels(
+              record['kubernetes'], %w(namespace_name pod_name)
+                .map { |l| [l, "#{CONTAINER_CONSTANTS[:service]}/#{l}"] }.to_h))
+          # Prepend label/ to all user-defined labels' keys.
+          if record['kubernetes'].key?('labels')
+            record['kubernetes']['labels'].each do |key, value|
```
igorpeshansky: Seems reminiscent of `delete_and_extract_labels` as well.

test/plugin/base_test.rb
```diff
@@ -1329,6 +1268,25 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
     assert i == n, "Number of entries #{i} does not match expected number #{n}"
   end
 
+  def verify_container_logs(log_entry_method, expected_params)
```
qingling128: Renamed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+
+    # Set required variables like @project_id, @vm_id, @vm_name and @zone.
+    def set_required_metadata_variables
+      set_project_id
+      set_vm_id
+      set_vm_name
+      set_location
+
+      # All metadata parameters must now be set.
+      return if @project_id && @zone && @vm_id
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+      resource.type = determine_agent_level_monitored_resource_type
+      resource.labels = determine_agent_level_monitored_resource_labels(
+        resource.type)
+      resource
+    end
+
+    # Determine agent level monitored resource type.
+    def determine_agent_level_monitored_resource_type
+      # EC2 instance.
+      return EC2_CONSTANTS[:resource_type] if
```
qingling128: Yup, `case` is more readable. Changed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
         @platform == Platform::EC2
-      # See http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
-      open('http://' + METADATA_SERVICE_ADDR +
-           '/latest/dynamic/instance-identity/document') do |f|
-        contents = f.read
-        return JSON.parse(contents)
+
+      # Unknown platform will be defaulted to GCE instance..
```
qingling128: Removed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+      return COMPUTE_CONSTANTS[:resource_type] if
+        @platform == Platform::OTHER
+
+      # Resource types determined by @subservice_name config.
+      # Cloud Dataflow.
+      return DATAFLOW_CONSTANTS[:resource_type] if
+        @subservice_name == DATAFLOW_CONSTANTS[:service]
```
qingling128: Changed to use a `SUB_SERVICE_MAP`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+      COMPUTE_CONSTANTS[:resource_type]
+    end
+
+    # Determine agent level monitored resource labels based on the resource
+    # type. Each resource type has its own labels that need to be filled in.
+    def determine_agent_level_monitored_resource_labels(type)
+      labels = {}
+
+      case type
+
```
qingling128: Removed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+      # GAE app.
+      when APPENGINE_CONSTANTS[:resource_type]
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+        begin
+          labels['module_id'] = fetch_gce_metadata(
+            'instance/attributes/gae_backend_name')
+          labels['version_id'] = fetch_gce_metadata(
+            'instance/attributes/gae_backend_version')
+        rescue StandardError => e
+          @log.error 'Failed to set monitored resource labels for GAE: ',
```
qingling128: On a second thought, we can just log with the resource type name. So we can achieve having one `begin` and `rescue` without introducing an extra map. PTAL.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Removed.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Removed.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Combined.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+
+      case @resource.type
+
+      # GAE app.
+      when APPENGINE_CONSTANTS[:resource_type]
+        labels["#{COMPUTE_CONSTANTS[:service]}/resource_id"] = @vm_id
+        labels["#{COMPUTE_CONSTANTS[:service]}/resource_name"] = @vm_name
+        labels["#{COMPUTE_CONSTANTS[:service]}/zone"] = @zone
+
+      # GCE and GKE container.
```

lib/fluent/plugin/out_google_cloud.rb
qingling128: Combined.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: In fact, when `@running_cloudfunctions` is true, we want to match tags against both regexps. So I changed it to be a list instead of a map. Then I realized it's more readable without inlining the function since we need to return early when a regexp is matched.

I also got rid of the `@running_cloudfunctions` variable`.

Initial set up:
```
      # Set regexp that we should match tags against later on. Using a list
      # instead of a map to ensure order. For example, tags will be matched
      # against Cloud Functions first, then GKE.
      @tag_regexp_list = []
      if @resource.type == CONTAINER_CONSTANTS[:resource_type]
        # We only support Cloud Functions logs for GKE right now.
        if fetch_gce_metadata('instance/attributes/'
                             ).split.include?('gcf_region')
          # Fetch this info and store it to avoid recurring
          # metadata server calls.
          @gcf_region = fetch_gce_metadata('instance/attributes/gcf_region')
          @tag_regexp_list << {
            tag_regexp: @cloudfunctions_tag_regexp,
            derived_type: CLOUDFUNCTIONS_CONSTANTS[:resource_type]
          }
        end
        @tag_regexp_list << {
          tag_regexp: @compiled_kubernetes_tag_regexp,
          derived_type: CONTAINER_CONSTANTS[:resource_type]
        }
      end
```

Tag parsing:
```
    def determine_group_level_monitored_resource_type(tag)
      @tag_regexp_list.each do |rule_pair|
        matched_regex_group = rule_pair[:tag_regexp].match(tag)
        return [rule_pair[:derived_type], matched_regex_group] if
          matched_regex_group
      end
      [@resource.type, nil]
    end
```

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Seems we can't simply apply that function because we are setting two labels to the same thing `instance_id`.

```
        instance_id = group_resource_labels.delete('instance_id')
        group_common_labels.merge(
          "#{CONTAINER_CONSTANTS[:service]}/instance_id" => instance_id,
          "#{COMPUTE_CONSTANTS[:service]}/resource_id" => instance_id,
...
        )
```

lib/fluent/plugin/out_google_cloud.rb
qingling128: Changed to use `delete_and_extract_labels`.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Moved this list to `CONTAINER_CONSTANCE[:extra_common_labels]`.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Moved this list to `CONTAINER_CONSTANCE[:extra_resource_labels]`.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +631,434 @@ def fetch_gce_metadata(metadata_path)
...
+      # field as part of the payload.
+      common_labels.merge!(delete_and_extract_labels(record, @label_map))
+
+      # Cloud Dataflow.
+      # These labels can be set via configuring 'labels' or 'label_map'.
+      # Report them as monitored resource labels instead of common labels.
+      if group_resource.type == DATAFLOW_CONSTANTS[:resource_type]
+        resource_labels.merge!(
+          delete_and_extract_labels(
+            common_labels, %w(region job_name job_id step_id)
```
qingling128: Moved this list to `DATAFLOW_CONSTANTS[:extra_resource_labels]`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1178,43 +1348,35 @@ def decode_cloudfunctions_function_name(function_name)
...
     end
 
     # For every original_label => new_label pair in the label_map, delete the
-    # original_label from the record if it exists, and extract the value to form
-    # a map with the new_label as the key.
-    def fields_to_labels(record, label_map)
-      return {} if label_map.nil? || !label_map.is_a?(Hash)
+    # original_label from the original_resource map if it exists, and extract
+    # the value to form a map with the new_label as the key.
+    def delete_and_extract_labels(original_resource, label_map)
```
qingling128: Renamed `original_resource` to `hash`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +641,409 @@ def fetch_gce_metadata(metadata_path)
...
+            group_resource_labels.delete('cluster_name'),
+          "#{COMPUTE_CONSTANTS[:service]}/zone" =>
+            group_resource_labels.delete('zone')
+        )
+
+      # GKE container.
+      when CONTAINER_CONSTANTS[:resource_type]
+        unless matched_regex_group.nil?
+          matched_regex_group_hash = Hash[
+            matched_regex_group.names.zip(matched_regex_group.captures)]
```
igorpeshansky: Might be worth commenting that we only expect one occurrence of each key in the match group (see [this](https://stackoverflow.com/questions/11688726/convert-named-matches-in-matchdata-to-hash#answer-33837669) for an example of when it would matter).

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +641,409 @@ def fetch_gce_metadata(metadata_path)
...
+          "#{COMPUTE_CONSTANTS[:service]}/resource_id" => instance_id,
+          "#{CONTAINER_CONSTANTS[:service]}/cluster_name" =>
+            group_resource_labels.delete('cluster_name'),
+          "#{COMPUTE_CONSTANTS[:service]}/zone" =>
+            group_resource_labels.delete('zone')
+        )
+
+      # GKE container.
+      when CONTAINER_CONSTANTS[:resource_type]
+        unless matched_regex_group.nil?
```
igorpeshansky: This could just be `if matched_regex_group`, right?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +641,409 @@ def fetch_gce_metadata(metadata_path)
...
+                # in fact 'namespace_id'. 'pod_name' is in fact 'pod_id'.
+                'container_name' => 'container_name',
+                'namespace_name' => 'namespace_id',
+                'pod_name' => 'pod_id'
+              }, false))
+
+          group_common_labels.merge!(
+            delete_and_extract_labels(
+              matched_regex_group_hash,
+              CONTAINER_CONSTANTS[:extra_common_labels]
```
igorpeshansky: [Optional] I find `Hash[list.map { ... }]` easier to read than `list.map { ... }.to_h`, because the hash construction is upfront. Also in other places below.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1178,43 +1333,39 @@ def decode_cloudfunctions_function_name(function_name)
...
     end
 
     # For every original_label => new_label pair in the label_map, delete the
-    # original_label from the record if it exists, and extract the value to form
-    # a map with the new_label as the key.
-    def fields_to_labels(record, label_map)
-      return {} if label_map.nil? || !label_map.is_a?(Hash)
+    # original_label from the hash map if it exists, and extract the value to
+    # form a map with the new_label as the key.
+    def delete_and_extract_labels(hash, label_map, should_delete = true)
```
igorpeshansky: Why do you need the `should_delete` flag? In all cases where you set it to `false`, we don't actually care if we destroy the hash.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -70,8 +73,10 @@ module Constants
       }
       ML_CONSTANTS = {
         service: 'ml.googleapis.com',
-        resource_type: 'ml_job'
+        resource_type: 'ml_job',
+        extra_common_labels: %w(job_id task_name)
       }
+      SUB_SERVICE_LIST = [DATAFLOW_CONSTANTS, ML_CONSTANTS]
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -267,162 +269,52 @@ def configure(conf)
...
+      # Retrieve monitored resource.
+      #
+      # TODO(qingling128): After Metadata Agent support is added, try retrieving
+      # the monitored resource from Metadata Agent first.
+      @resource = determine_agent_level_monitored_resource_via_legacy
+
+      # Set regexp that we should match tags against later on. Using a list
+      # instead of a map to ensure order. For example, tags will be matched
+      # against Cloud Functions first, then GKE.
+      @tag_regexp_list = []
```
qingling128: The original [commit](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/commit/44dccaf724841f90116eb28a1bebf0877a1ed5d7) was from 2 years ago.

```
+        if @running_cloudfunctions
 +          # If the current group of entries is coming from a Cloud Functions
 +          # function, the function name can be extracted from the tag.
 +          match_data = @cloudfunctions_tag_regexp.match(tag)
 +          if match_data
 +            # Service name is set to Cloud Functions only for logs actually
 +            # coming from a function.
 +            @service_name = CLOUDFUNCTIONS_SERVICE
 +            labels = write_log_entries_request['commonLabels']
 +            labels["#{CLOUDFUNCTIONS_SERVICE}/function_name"] =
 +              match_data['function_name']
 +          else
 +            # Other logs are considered as coming from the Container Engine
 +            # service.
 +            @service_name = CONTAINER_SERVICE
 +          end
 +        end
```

`Cloud Functions` is only supported on `GKE`. So we default back to the `GKE` resource type when the tag does not match with `Cloud Function` format. Then we match the tag against `GKE` regex.

I think this makes sense because `@running_cloudfunctions` is true as long as `gcf_region` is available from metadata server. Yet we still want other GKE system logs to be tagged properly.

Also these two tests (`test_multiple_cloudfunctions_logs_tag_not_matched` and `test_multiple_cloudfunctions_logs_tag_not_matched`) were added in this second [commit](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/commit/cb0c64257586a44d50cafc3c1e9ca969c9f1988c) to explicitly ensure the behavior.

WDYT?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -267,162 +269,52 @@ def configure(conf)
...
+      if @resource.type == CONTAINER_CONSTANTS[:resource_type]
+        # We only support Cloud Functions logs for GKE right now.
+        if fetch_gce_metadata('instance/attributes/'
```
qingling128: I changed `tag_regexp_list` to a list of pairs. Also changed the order of `resource type` and `tag regex`.

As for the error handling change, I found it a bit more confusing after the change. Because catching an error here seems to imply `instance/attributes/gcf_region` is supposed to be present for the most common cases. But in fact it's the opposite. Also, it will make it hard to tell the difference between the `instance/attributes/gcf_region` not being present or the metadata server going down. So I left this part the way it was. Does that make sense?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -267,162 +269,52 @@ def configure(conf)
...
+                             ).split.include?('gcf_region')
+          # Fetch this info and store it to avoid recurring
+          # metadata server calls.
+          @gcf_region = fetch_gce_metadata('instance/attributes/gcf_region')
+          @tag_regexp_list << {
+            tag_regexp: @cloudfunctions_tag_regexp,
```
qingling128: I think `cloudfunctions_tag_regexp` is already a regex, not a string. We have to do

```
@compiled_kubernetes_tag_regexp = Regexp.new(@kubernetes_tag_regexp)
```

because `@kubernetes_tag_regexp` is a string configurable by users.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +641,409 @@ def fetch_gce_metadata(metadata_path)
...
+          SUB_SERVICE_LIST.each do |resource_constants|
+            return resource_constants[:resource_type] if
+              @subservice_name == resource_constants[:service]
+          end
+          # Default back to GCE if invalid value is detected.
+          return COMPUTE_CONSTANTS[:resource_type]
+        end
+
+        # Resource types determined by @detect_subservice config.
+        if @detect_subservice
```
qingling128: I renamed `SUBSERVICE_MAP` to `SUBSERVICE_NAME_MAP`. And added a `SUBSERVICE_ATTRIBUTE_MAP` to add these items (because they are kind of different). The refactored code is as below.

```
      # The map between a resource type and expected subservice attributes.
      SUBSERVICE_ATTRIBUTE_MAP = {
        # GAE.
        APPENGINE_CONSTANTS[:resource_type] =>
          %w(gae_backend_name gae_backend_version),
        # GKE.
        CONTAINER_CONSTANTS[:resource_type] => %w(kube-env),
        # Cloud Dataproc.
        DATAPROC_CONSTANTS[:resource_type] =>
          %w(dataproc-cluster-uuid dataproc-cluster-name)
      }
```
```
        if @detect_subservice
          begin
            attributes = fetch_gce_metadata('instance/attributes/').split
            SUBSERVICE_ATTRIBUTE_MAP.map do |resource_type, expected_attributes|
              return resource_type if
                expected_attributes.all? { |attr| attributes.include?(attr) }
            end
          rescue StandardError => e
            @log.error 'Failed to detect subservice: ', error: e
          end
        end
```

PTAL.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +641,409 @@ def fetch_gce_metadata(metadata_path)
...
+      end
+    end
+
+    # Determine agent level monitored resource labels based on the resource
+    # type. Each resource type has its own labels that need to be filled in.
+    def determine_agent_level_monitored_resource_labels(type)
+      case type
+      # GAE app.
+      when APPENGINE_CONSTANTS[:resource_type]
+        labels = {
```

lib/fluent/plugin/out_google_cloud.rb
qingling128: Oops. Right. Changed.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Interesting. Just added a comment `We only expect one occurrence of each key in the match group.`

lib/fluent/plugin/out_google_cloud.rb
qingling128: I was also debating among using `Hash[]`, `to_hash` and `to_h`. After some googling, it appeared to me that `to_h` is more popular. And I did see that more often in Ruby doc examples.

Some discussion around it:
https://stackoverflow.com/questions/28302482/hash-vs-to-h-are-they-the-same-and-if-so-which-is-better
http://zverok.github.io/blog/2016-01-18-implicit-vs-expicit.html

WDYT?

lib/fluent/plugin/out_google_cloud.rb
qingling128: Ah, I added this because in order to apply `delete_and_extract_labels` to `matched_regex_group`. Right now we are setting matched_regex_group['pod_name'] to both resource label `pod_id` and common label `container.googleapis.com/pod_name`. So we can not destroy the hash on the first run.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +656,397 @@ def fetch_gce_metadata(metadata_path)
...
+        if matched_regex_group
+          # We only expect one occurrence of each key in the match group.
+          matched_regex_group_hash =
+            matched_regex_group.names.zip(matched_regex_group.captures).to_h
+          group_resource_labels.merge!(
+            delete_and_extract_labels(
+              matched_regex_group_hash, {
+                # The kubernetes_tag_regexp is poorly named. 'namespace_name' is
+                # in fact 'namespace_id'. 'pod_name' is in fact 'pod_id'.
+                'container_name' => 'container_name',
```
igorpeshansky: The keys in this hash are `CONTAINER_CONSTANTS[:extra_resource_labels]`. Let's add a TODO to figure out how to reuse them.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: This seems like an unnecessary complication. Instead, you can either construct the hash before every call to `delete_and_extract_labels`, or just pass in the `.dup` of the hash.

lib/fluent/plugin/out_google_cloud.rb
qingling128: I renamed the maps to avoid confusion.

Also changed
`return resource_type if expected_attributes.all? { |attr| attributes.include?(attr) }`
to 
`return resource_type if attributes.superset?(expected)`.

I don't get why we should put `service -> attributes` in `SUBSERVICE_METADATA_ATTRIBUTES`, then do another map lookup to get `service -> resource_type` though. Can't we just put `resource_type -> attributes` in `SUBSERVICE_METADATA_ATTRIBUTES`? Ultimately we want to get the resource types.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Just renamed the variables.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Cool.

lib/fluent/plugin/out_google_cloud.rb
qingling128: No worries.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +656,397 @@ def fetch_gce_metadata(metadata_path)
...
+    end
+
+    # Determine agent level monitored resource labels based on the resource
+    # type. Each resource type has its own labels that need to be filled in.
+    def determine_agent_level_monitored_resource_labels(type)
+      case type
+      # GAE app.
+      when APPENGINE_CONSTANTS[:resource_type]
+        return {
+          'module_id' => fetch_gce_metadata(
```

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Cool, I'll leave it in the common format that Ruby developers use (`to_h`) then.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Removed the `should_delete` variable and use `.dup` instead.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +656,397 @@ def fetch_gce_metadata(metadata_path)
...
+        labels = {
+          'instance_id' => @vm_id,
+          'region' => @zone
+        }
+        labels['aws_account'] = ec2_metadata['accountId'] if
+          ec2_metadata.key?('accountId')
+        return labels
+      end
+
+      return {}
```
qingling128: Seems it's fine with this format. Yet I just changed to `{}` anyway.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +656,397 @@ def fetch_gce_metadata(metadata_path)
...
+
+      when Platform::GCE
+        # Resource types determined by @subservice_name config.
+        return SUBSERVICE_NAME_MAP[@subservice_name] if @subservice_name
+
+        # Resource types determined by @detect_subservice config.
+        if @detect_subservice
+          begin
+            attributes = fetch_gce_metadata('instance/attributes/').split
+            SUBSERVICE_ATTRIBUTE_MAP.map do |resource_type, expected_attributes|
```
qingling128: Ah, right. `map` is an overkill here. Changed to `each`.

qingling128:
I've addressed all the feedback. Yet seems some of my comments did not show up in the main comment history (probably because I rebased the branch and Github got confused). But if you click on the actual commit, that view should show all the commits. PTAL.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -70,7 +73,28 @@ module Constants
...
+      # The map between a subservice name and a resource type.
+      SUBSERVICE_MAP = [
+        APPENGINE_CONSTANTS, CONTAINER_CONSTANTS, DATAFLOW_CONSTANTS,
+        DATAPROC_CONSTANTS, ML_CONSTANTS
+      ].map { |consts| [consts[:service], consts[:resource_type]] }.to_h
+      # Default back to GCE if invalid value is detected.
+      SUBSERVICE_MAP.default = COMPUTE_CONSTANTS[:resource_type]
+
+      # The map between a resource type and expected subservice attributes.
+      SUBSERVICE_METADATA_ATTRIBUTES = {
```
bmoyles0117: Could/should we define metadata attributes as a property of the constants and define this map the same way we define the SUBSERVICE_MAP?

lib/fluent/plugin/out_google_cloud.rb
rbuskens: I agree. service -> resource_type and resource_type -> attributes makes more sense to me.

Naive question: where do you actually use SUBSERVICE_MAP?

lib/fluent/plugin/out_google_cloud.rb
bmoyles0117: Are both`@compiled_kubernetes_tag_regexp` necessary here? Will it default to nil anyways if kubernetes_tag_regexp is not set?

lib/fluent/plugin/out_google_cloud.rb
bmoyles0117: I'm fine with this as long as the user can distinguish the actual here. SGTM.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +659,399 @@ def fetch_gce_metadata(metadata_path)
...
+      when COMPUTE_CONSTANTS[:resource_type]
+        return {
+          'instance_id' => @vm_id,
+          'zone' => @zone
+        }
+
+      # GKE container.
+      when CONTAINER_CONSTANTS[:resource_type]
+        raw_kube_env = fetch_gce_metadata('instance/attributes/kube-env')
+        kube_env = YAML.load(raw_kube_env)
```
bmoyles0117: Could kube_env end up being defined as nil in an error situation? If so, I think `cluster_name_from_kube_env` might blow up attempting to access the key.

lib/fluent/plugin/out_google_cloud.rb
qingling128: That's not naive at all. Turned out I did not rename the variable at the other side. I just changed it. I also created a ticket to add unit tests for logic around `@subservice_name` so this can be captured in the future.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Yeah. That's why we have the `rescue` below. Any error that is thrown would be caught there and logged.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Removed the first one.

qingling128:
Feedback addressed. PTAL.

lib/fluent/plugin/out_google_cloud.rb
rbuskens: LGTM.

qingling128:
THX!

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -76,7 +80,20 @@ module Constants
...
+        [APPENGINE_CONSTANTS, CONTAINER_CONSTANTS, DATAFLOW_CONSTANTS,
+         DATAPROC_CONSTANTS, ML_CONSTANTS]
+        .map { |consts| [consts[:service], consts[:resource_type]] }.to_h
+      # Default back to GCE if invalid value is detected.
+      SUBSERVICE_MAP.default = COMPUTE_CONSTANTS[:resource_type]
+
+      # The map between a resource type and expected subservice attributes.
+      SUBSERVICE_METADATA_ATTRIBUTES = \
+        [APPENGINE_CONSTANTS, CONTAINER_CONSTANTS, DATAPROC_CONSTANTS]
+        .map { |consts| [consts[:resource_type], consts[:metadata_attributes]] }
```
igorpeshansky: If you added `.to_set` here, you could have avoided it in the constants themselves...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +657,398 @@ def fetch_gce_metadata(metadata_path)
...
+        labels['aws_account'] = ec2_metadata['accountId'] if
+          ec2_metadata.key?('accountId')
+        return labels
+      end
+
+      {}
+    rescue StandardError => e
+      @log.error "Failed to set monitored resource labels for #{type}: ",
+                 error: e
+      return {}
```
igorpeshansky: This could have been just `{}` as well...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +657,398 @@ def fetch_gce_metadata(metadata_path)
...
+    # Determine group level monitored resource type shared by a collection of
+    # entries.
+    # Return the resource type and tag regexp matched groups. The matched groups
+    # only apply to some resource types. Return nil if not applicable or if
+    # there is no match.
+    def determine_group_level_monitored_resource_type(tag)
+      @tag_regexp_list.each do |derived_type, tag_regexp|
+        matched_regex_group = tag_regexp.match(tag)
+        return [derived_type, matched_regex_group] if
+          matched_regex_group
```
igorpeshansky: Fits on the previous line.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -873,15 +657,398 @@ def fetch_gce_metadata(metadata_path)
...
+        )
+
+      # GKE container.
+      when CONTAINER_CONSTANTS[:resource_type]
+        if matched_regex_group
+          # We only expect one occurrence of each key in the match group.
+          resource_labels_candidates =
+            matched_regex_group.names.zip(matched_regex_group.captures).to_h
+          common_labels_candidates =
+            resource_labels_candidates.dup
```
igorpeshansky: Fits on the previous line.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1178,43 +1338,35 @@ def decode_cloudfunctions_function_name(function_name)
...
+    # original_label from the hash map if it exists, and extract the value to
+    # form a map with the new_label as the key.
+    def delete_and_extract_labels(hash, label_map)
+      return {} if label_map.nil? || !label_map.is_a?(Hash) ||
+                   hash.nil? || !hash.is_a?(Hash)
       label_map.each_with_object({}) \
         do |(original_label, new_label), extracted_labels|
-        extracted_labels[new_label] = convert_to_utf8(
-          record.delete(original_label).to_s) if record.key?(original_label)
+        extracted_labels[new_label] =
```
igorpeshansky: Alternatively:
```ruby
        value = hash.delete(original_label)
        extracted_labels[new_label] = convert_to_utf8(value.to_s) if value
```

lib/fluent/plugin/out_google_cloud.rb
qingling128: Addressed in the follow up PR: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/148/commits/f3fbf25a37cf73dc44576d57be30ea77f7a0c322.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Addressed in the follow up PR: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/148/commits/f3fbf25a37cf73dc44576d57be30ea77f7a0c322.

lib/fluent/plugin/out_google_cloud.rb
qingling128: This no longer applies in the follow up PR: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/148/commits/f3fbf25a37cf73dc44576d57be30ea77f7a0c322.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Addressed in the follow up PR: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/148/commits/f3fbf25a37cf73dc44576d57be30ea77f7a0c322.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Addressed in the follow up PR: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/148/commits/f3fbf25a37cf73dc44576d57be30ea77f7a0c322.

PR #135, opened 2017-06-30
Add optional prometheus monitoring

crassirostris:
Add an option to the plugin which, when enabled, registers and propagates metrics in the default Prometheus registry to be later picked up and exposed on some endpoint by e.g. https://github.com/kazegusuri/fluent-plugin-prometheus. Prometheus-specific part is abstracted away from the working code to be easily extended to facilitate others monitoring solutions.

/cc @igorpeshansky @qingling128 @piosz @fgrzadkowski

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -186,6 +186,12 @@ module Fluent
                  :default => nil,
                  :secret => true
 
+    # Whether to use prometheus client to initialize and set metrics. Setting
+    # this option to true doesn't expose an endpoint with metrics in prometheus
+    # format, it only uses default prometheus client to store values. To
+    # actually expose metrics, a separate fluentd plugin should be used.
+    config_param :prometheus_monitoring_enabled, :bool, :default => false
+
     # rubocop:enable Style/HashSyntax
 
     # TODO: Add a log_name config option rather than just using the tag?
@@ -387,6 +393,33 @@ module Fluent
       @resource.freeze
       @resource.labels.freeze
       @common_labels.freeze
+
+      # If prometheus monitoring is enabled, register metrics in the default
+      # registry and store metric objects for future use. If metrics are
+      # registed already, just store the metric objects.
+      if @prometheus_monitoring_enabled
+        require 'prometheus/client'
+
+        prometheus = Prometheus::Client.registry
+
+        # Exception-driven behavior to avoid synchronization errors
+        begin
+          @prometheus_http_requests_count = prometheus.counter(
+            :stackdriver_http_requests_count,
+            'A number of http requests to Stackdriver Logging API')
+        rescue Prometheus::Client::Registry::AlreadyRegisteredError
+          @prometheus_http_requests_count = prometheus.get(
+            :stackdriver_http_requests_count)
+        end
+        begin
+          @prometheus_log_entries_count = prometheus.counter(
+            :stackdriver_log_entries_count,
+            'A number of log entries sent to Stackdriver Logging API')
+        rescue Prometheus::Client::Registry::AlreadyRegisteredError
+          @prometheus_log_entries_count = prometheus.get(
+            :stackdriver_log_entries_count)
+        end
+      end
 
       # Log an informational message containing the Logs viewer URL
       @log.info 'Logs viewer address: https://console.cloud.google.com/logs/',
@@ -652,6 +685,8 @@ module Fluent
             )
 
             client.write_log_entries(write_request)
+            increment_requests_metric(GRPC::Core::StatusCodes::OK)
+            increment_log_entries_metric(entries.length, true)
 
             # Let the user explicitly know when the first call succeeded,
             # to aid with verification and troubleshooting.
@@ -661,10 +696,12 @@ module Fluent
             end
 
           rescue GRPC::Cancelled => error
+            increment_requests_metric(GRPC::Core::StatusCodes::CANCELLED)
             # RPC cancelled, so retry via re-raising the error.
             raise error
 
           rescue GRPC::BadStatus => error
+            increment_requests_metric(error.code)
             case error.code
             when GRPC::Core::StatusCodes::CANCELLED,
                  GRPC::Core::StatusCodes::UNAVAILABLE,
@@ -679,6 +716,7 @@ module Fluent
               # Most client errors indicate a problem with the request itself
               # and should not be retried.
               dropped = entries.length
+              increment_log_entries_metric(dropped, false)
               @log.warn "Dropping #{dropped} log message(s)",
                         error: error.to_s, error_code: error.code.to_s
             when GRPC::Core::StatusCodes::UNAUTHENTICATED
@@ -686,12 +724,14 @@ module Fluent
               # These are usually solved via a `gcloud auth` call, or by
               # modifying the permissions on the Google Cloud project.
               dropped = entries.length
+              increment_log_entries_metric(dropped, false)
               @log.warn "Dropping #{dropped} log message(s)",
                         error: error.to_s, error_code: error.code.to_s
             else
               # Assume this is a problem with the request itself
               # and don't retry.
               dropped = entries.length
+              increment_log_entries_metric(dropped, false)
               @log.error "Unknown response code #{error.code} from the "\
                          "server, dropping #{dropped} log message(s)",
                          error: error.to_s, error_code: error.code.to_s
@@ -707,7 +747,14 @@ module Fluent
                 entries: entries)
 
             # TODO: RequestOptions
-            client.write_entry_log_entries(write_request)
+            begin
+              client.write_entry_log_entries(write_request)
+            rescue Google::Apis::Error => error
+              increment_requests_metric(error.status_code)
+              raise error
+            end
+            increment_requests_metric(200)
+            increment_log_entries_metric(entries.length, true)
 
             # Let the user explicitly know when the first call succeeded,
             # to aid with verification and troubleshooting.
@@ -725,6 +772,7 @@ module Fluent
             # These are usually solved via a `gcloud auth` call, or by modifying
             # the permissions on the Google Cloud project.
             dropped = entries.length
+            increment_log_entries_metric(dropped, false)
             @log.warn "Dropping #{dropped} log message(s)",
                       error_class: error.class.to_s, error: error.to_s
 
@@ -732,6 +780,7 @@ module Fluent
             # Most ClientErrors indicate a problem with the request itself and
             # should not be retried.
             dropped = entries.length
+            increment_log_entries_metric(dropped, false)
             @log.warn "Dropping #{dropped} log message(s)",
                       error_class: error.class.to_s, error: error.to_s
           end
@@ -1344,6 +1393,22 @@ module Fluent
                      'to true.'
           raise
         end
+      end
+    end
+
+    # Increment the metric for the number of requests, labeled by
+    # the provided status code.
+    def increment_requests_metric(code)
+      if @prometheus_monitoring_enabled
+        @prometheus_http_requests_count.increment({ grpc: @use_grpc, code: code })
+      end
+    end
+
+    # Increment the metric for the number of log entries, labeled by
+    # the success status of their ingestion.
+    def increment_log_entries_metric(count, success)
+      if @prometheus_monitoring_enabled
+        @prometheus_log_entries_count.increment({ success: success }, count)
       end
     end
   end
--- a/test/plugin/constants.rb
+++ b/test/plugin/constants.rb
@@ -140,6 +140,10 @@ module Constants
     detect_subservice false
   )
 
+  PROMETHEUS_ENABLE_CONFIG = %(
+    prometheus_monitoring_enabled true
+  )
+
   CUSTOM_METADATA_CONFIG = %(
     project_id #{CUSTOM_PROJECT_ID}
     zone #{CUSTOM_ZONE}
--- a/test/plugin/test_out_google_cloud.rb
+++ b/test/plugin/test_out_google_cloud.rb
@@ -70,6 +70,76 @@ class GoogleCloudOutputTest < Test::Unit::TestCase
     assert_equal 1, exception_count
   end
 
+  def test_prometheus_successful_call
+    setup_gce_metadata_stubs
+    setup_prometheus
+    stub_request(:post, WRITE_LOG_ENTRIES_URI)
+      .to_return(status: 200, body: 'OK')
+    d = create_driver(PROMETHEUS_ENABLE_CONFIG)
+    d.emit('message' => log_entry(0))
+    d.run
+    assert_prometheus_metric_value(:stackdriver_http_requests_count, 1, { grpc: false, code: 200 })
+  end
+
+  def test_prometheus_failed_call
+    setup_gce_metadata_stubs
+    setup_prometheus
+    stub_request(:post, WRITE_LOG_ENTRIES_URI)
+      .to_return(status: 401, body: 'Unauthorized')
+    d = create_driver(PROMETHEUS_ENABLE_CONFIG)
+    d.emit('message' => log_entry(0))
+    begin
+      d.run
+    rescue Google::Apis::AuthorizationError => error
+      assert_equal 'Unauthorized', error.message
+    end
+    assert_prometheus_metric_value(:stackdriver_http_requests_count, 1, { grpc: false, code: 401 })
+  end
+
+  def test_prometheus_successfully_ingested_entries
+    setup_gce_metadata_stubs
+    setup_prometheus
+    stub_request(:post, WRITE_LOG_ENTRIES_URI)
+      .to_return(status: 200, body: 'OK')
+    d = create_driver(PROMETHEUS_ENABLE_CONFIG)
+    d.emit('message' => log_entry(0))
+    d.run
+    assert_prometheus_metric_value(:stackdriver_log_entries_count, 1, { success: true })
+    assert_prometheus_metric_value(:stackdriver_log_entries_count, 0, { success: false })
+  end
+
+  def test_prometheus_dropped_entries
+    setup_gce_metadata_stubs
+    setup_prometheus
+    stub_request(:post, WRITE_LOG_ENTRIES_URI)
+      .to_return(status: 401, body: 'Unauthorized')
+    d = create_driver(PROMETHEUS_ENABLE_CONFIG)
+    d.emit('message' => log_entry(0))
+    begin
+      d.run
+    rescue Google::Apis::AuthorizationError => error
+      assert_equal 'Unauthorized', error.message
+    end
+    assert_prometheus_metric_value(:stackdriver_log_entries_count, 0, { success: true })
+    assert_prometheus_metric_value(:stackdriver_log_entries_count, 1, { success: false })
+  end
+
+  def test_prometheus_retry_without_ingesting
+    setup_gce_metadata_stubs
+    setup_prometheus
+    stub_request(:post, WRITE_LOG_ENTRIES_URI)
+      .to_return(status: 500, body: 'Server Error')
+    d = create_driver(PROMETHEUS_ENABLE_CONFIG)
+    d.emit('message' => log_entry(0))
+    begin
+      d.run
+    rescue Google::Apis::ServerError => error
+      assert_equal 'Server error', error.message
+    end
+    assert_prometheus_metric_value(:stackdriver_log_entries_count, 0, { success: true })
+    assert_prometheus_metric_value(:stackdriver_log_entries_count, 0, { success: false })
+  end
+
   # This test looks similar between the grpc and non-grpc paths except that when
   # parsing "105", the grpc path responds with "DEBUG", while the non-grpc path
   # responds with "100".
--- a/test/plugin/test_out_google_cloud_grpc.rb
+++ b/test/plugin/test_out_google_cloud_grpc.rb
@@ -72,6 +72,78 @@ class GoogleCloudOutputGRPCTest < Test::Unit::TestCase
     end
   end
 
+  def test_prometheus_successful_call
+    setup_gce_metadata_stubs
+    setup_prometheus
+    setup_logging_stubs(false, 0, 'OK') do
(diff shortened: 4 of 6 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -388,6 +394,33 @@ def configure(conf)
...
+      # registry and store metric objects for future use. If metrics are
+      # registed already, just store the metric objects.
+      if @prometheus_monitoring_enabled
+        require 'prometheus/client'
+
+        prometheus = Prometheus::Client.registry
+
+        # Exception-driven behavior to avoid synchronization errors
+        begin
+          @prometheus_http_requests_count = prometheus.counter(
```
igorpeshansky: This is generally useful beyond Prometheus. How about creating a `monitoring` module, which uses the prometheus client if available, and changing these to `monitoring.counter` instead. Something like this:
```ruby
module Monitoring
  begin
    require 'prometheus/client'
    @prometheus = Prometheus::Client.registry
  rescue LoadError
    @prometheus = nil
  end
  class Counter
    def initialize(name, desc)
      @name = name
      @desc = desc
      @map = {}
      @map.default = 0
    end
    def increment(labels, inc=1)
      @map[labels] = @map[labels] + inc
    end
  end
  def counter(name, desc)
    return @prometheus ? @prometheus.counter(name, desc) : Counter.new(name, desc)
  end
end
```
Then these variables become just `@http_requests_count` (or even just `@requests_count`) and `@log_entries_count`, and the option becomes just `@monitoring_enabled`...

We'd probably want to test the prometheus path end-to-end anyway, though.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -388,6 +394,33 @@ def configure(conf)
       @resource.labels.freeze
       @common_labels.freeze
 
+      # If prometheus monitoring is enabled, register metrics in the default
+      # registry and store metric objects for future use. If metrics are
+      # registed already, just store the metric objects.
+      if @prometheus_monitoring_enabled
```
igorpeshansky: Let's also fix the rubocop complaints (the build [failed](https://travis-ci.org/GoogleCloudPlatform/fluent-plugin-google-cloud/jobs/248830189)).

crassirostris: Ack, I just didn't see the results yet :)

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -388,6 +394,33 @@ def configure(conf)
       @resource.labels.freeze
       @common_labels.freeze
 
+      # If prometheus monitoring is enabled, register metrics in the default
+      # registry and store metric objects for future use. If metrics are
+      # registed already, just store the metric objects.
```
crassirostris: Ack

lib/fluent/plugin/out_google_cloud.rb
crassirostris: I didn't want to re-implement the "default solution", b/c there's actually pretty much synchronization code in the Prometheus library and copying it would result in a lot of code.

I don't see much profit in having several monitoring solution used at the same time. What do you think about the following: I create a base Monitoring class & a Prometheus implementation and you will create a custom implementation independently; then I change the `prometheus_monitoring_enabled` flag to `monitoring_type` with the string type, that can be one of the following: `none`, `prometheus` and later `json`?

igorpeshansky: Fair enough -- I admit I omitted the synchronization bits. But you missed my point -- I was asking to separate enabling monitoring from selecting Prometheus. I'm happy to have the `Monitoring` module be a no-op for now when `prometheus/client` can't be loaded. Then the non-Prometheus code becomes even simpler.

crassirostris: OK, I see. Do you mean smth like bool flag `enable_monitoring` and at the same time string flag `monitoring_type` defaulted to something?

igorpeshansky: Nah, even simpler -- if the prometheus library is available, use it, otherwise a no-op (which we'll later fill in with JSON).

crassirostris: I see. I just like the explicit mechanisms better, but it's your code :) I'll update the PR shortly

crassirostris: Oh, I've just realized that the implicit mechanism is much harder to test, since you have a dependency on the environment and you cannot test non-prometheus pipeline if prometheus gem is installed and vice versa (at least to my limited knowledge in ruby). Maybe `monitoring_type` is not such a bad idea?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -707,7 +740,14 @@ def write(chunk)
...
 
             # TODO: RequestOptions
-            client.write_entry_log_entries(write_request)
+            begin
+              client.write_entry_log_entries(write_request)
+            rescue Google::Apis::Error => error
+              increment_requests_metric(error.status_code)
+              raise error
+            end
+            increment_requests_metric(200)
```
crassirostris: I would be grateful if you could suggest something better

crassirostris:
@igorpeshansky PTAL. I took the liberty and introduced an explicit flag for the monitoring type, to make unit testing easier

crassirostris:
@igorpeshansky Friendly ping

fgrzadkowski:
@igorpeshansky let's try to merge this change asap, so that we can use it in prod to see where we are in terms of reliability. This is really important IMO.

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -72,6 +72,86 @@ def test_server_error
     end
   end
 
+  def test_prometheus_successful_call
+    setup_gce_metadata_stubs
+    setup_prometheus
+    setup_logging_stubs(false, 0, 'OK') do
```
igorpeshansky: I'm a little concerned about the code duplication here, but not enough to block this PR on that.

lib/fluent/plugin/monitoring.rb
```diff
@@ -0,0 +1,67 @@
...
+require 'google/apis'
+require 'google/apis/logging_v2beta1'
+require 'google/logging/v2/logging_pb'
+require 'google/logging/v2/logging_services_pb'
+require 'google/logging/v2/log_entry_pb'
+require 'googleauth'
+
+module Monitoring
+  # Factory that is used to create a monitoring registry based on
+  # the monitoring solution name
```
crassirostris: Ack

lib/fluent/plugin/monitoring.rb
```diff
@@ -0,0 +1,67 @@
...
+  class MonitoringRegistryFactory
+    def self.create(name)
+      case name
```
crassirostris: Ack

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -210,6 +224,18 @@ def initialize
     def configure(conf)
       super
 
+      # If monitoring is enabled, register metrics in the default registry
+      # and store metric objects for future use.
+      if @monitoring_enabled
+        registry = Monitoring::MonitoringRegistryFactory.create @monitoring_type
+        @http_requests_count = registry.counter(
```
crassirostris: Ack

lib/fluent/plugin/out_google_cloud.rb
crassirostris: Ack, works for us

test/plugin/test_out_google_cloud.rb
```diff
@@ -70,6 +70,84 @@ def test_server_error
...
 
+  def test_prometheus_successful_call
+    setup_gce_metadata_stubs
+    setup_prometheus
+    stub_request(:post, WRITE_LOG_ENTRIES_URI)
+      .to_return(status: 200, body: 'OK')
+    d = create_driver(PROMETHEUS_ENABLE_CONFIG)
+    d.emit('message' => log_entry(0))
+    d.run
+    assert_prometheus_metric_value(:stackdriver_http_requests_count, 1,
```
crassirostris: Ack

test/plugin/test_out_google_cloud_grpc.rb
crassirostris: Ack

crassirostris:
@igorpeshansky PTAL

test/plugin/test_out_google_cloud.rb
```diff
@@ -70,6 +70,70 @@ def test_server_error
...
+      # Several successful requests.
+      [200, 2, 2, 0],
+      # Single failed request.
+      [401, 1, 0, 1]
+    ].each do |code, emit_count, successful_count, failed_count|
+      setup_prometheus
+      stub_request(:post, WRITE_LOG_ENTRIES_URI)
+        .to_return(status: code, body: 'Some Message')
+      (1..emit_count).each do |i|
+        d = create_driver(PROMETHEUS_ENABLE_CONFIG)
```
igorpeshansky: Do you need to create the driver every time?

test/plugin/test_out_google_cloud.rb
```diff
@@ -70,6 +70,70 @@ def test_server_error
...
+  end
+
+  def test_prometheus_ingested_entries
+    setup_gce_metadata_stubs
+    [
+      # Single successful request.
+      [200, 1, 1, 0],
+      # Several successful requests.
+      [200, 2, 2, 0],
+      # Single failed request that caused logs dropping.
```
igorpeshansky: Nit: s/logs dropping/dropping logs/. Also below.

test/plugin/test_out_google_cloud.rb
```diff
@@ -70,6 +70,70 @@ def test_server_error
...
+        # rubocop:enable Lint/HandleExceptions
+      end
+      assert_prometheus_metric_value(:stackdriver_successful_requests_count,
+                                     successful_count, grpc: false)
+      assert_prometheus_metric_value(:stackdriver_failed_requests_count,
+                                     failed_count, grpc: false, code: code)
+    end
+  end
+
+  def test_prometheus_ingested_entries
```
igorpeshansky: Why wouldn't you fold this test into the previous one?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -72,6 +72,70 @@ def test_server_error
...
+  def test_prometheus_requests_count
+    setup_gce_metadata_stubs
+    [
+      # Single successful request.
+      [false, 0, 1, 1, 0],
+      # Several successful requests.
+      [false, 0, 2, 2, 0],
+      # Single failed request.
+      [true, 13, 1, 0, 1]
+    ].each do |is_fail, code, emit_count, successful_count, failed_count|
```
igorpeshansky: s/`is_fail`/`should_fail`/.

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -72,6 +72,70 @@ def test_server_error
...
+        end
+      end
+      assert_prometheus_metric_value(:stackdriver_successful_requests_count,
+                                     successful_count, grpc: true)
+      assert_prometheus_metric_value(:stackdriver_failed_requests_count,
+                                     failed_count, grpc: true, code: code)
+    end
+  end
+
+  def test_prometheus_ingested_entries
```
igorpeshansky: Same comments as in the other test file.

test/plugin/test_out_google_cloud.rb
```diff
@@ -70,6 +70,70 @@ def test_server_error
...
+      [401, 1, 0, 1],
+      # Single failed request that was escalated without logs dropping.
+      [500, 1, 0, 0]
+    ].each do |code, emit_count, successful_count, failed_count|
+      setup_prometheus
+      stub_request(:post, WRITE_LOG_ENTRIES_URI)
+        .to_return(status: code, body: 'Some Message')
+      (1..emit_count).each do |i|
+        d = create_driver(PROMETHEUS_ENABLE_CONFIG)
+        d.emit('message' => log_entry(i.to_s))
```
igorpeshansky: Should you be testing more than one request per entry?

test/plugin/test_out_google_cloud_grpc.rb
crassirostris: Ack

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -228,12 +228,15 @@ def configure(conf)
...
         registry = Monitoring::MonitoringRegistryFactory.create @monitoring_type
-        @http_requests_count = registry.counter(
-          :stackdriver_http_requests_count,
-          'A number of http requests to Stackdriver Logging API')
+        @successful_requests_count = registry.counter(
+          :stackdriver_successful_requests_count,
+          'A number of successful requests to the Stackdriver Logging API')
+        @failed_requests_count = registry.counter(
+          :stackdriver_failed_requests_count,
+          'A number of failed requests to the Stackdriver Logging API')
```
crassirostris: Ack

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -210,6 +224,21 @@ def initialize
...
+        registry = Monitoring::MonitoringRegistryFactory.create @monitoring_type
+        @successful_requests_count = registry.counter(
+          :stackdriver_successful_requests_count,
+          'A number of successful requests to the Stackdriver Logging API')
+        @failed_requests_count = registry.counter(
+          :stackdriver_failed_requests_count,
+          'A number of failed requests to the Stackdriver Logging API')
+        @log_entries_count = registry.counter(
+          :stackdriver_log_entries_count,
+          'A number of log entries sent to the Stackdriver Logging API')
```
crassirostris: Well, that ARE sent, just some of them were ingested and some -- weren't

But ack, I'll change

test/plugin/test_out_google_cloud.rb
crassirostris: If I run it only once, I don't know how to make several entries with the same tag to be batches into the separate requests

If I run it every time, but don't recreate the driver, entries emitted earlier are still there and change the metric values

test/plugin/test_out_google_cloud.rb
crassirostris: Tbh -- b/c rubocop doesn't allow more than 5 values in the parameters list :)

Ack, I'll change it to one method

test/plugin/test_out_google_cloud.rb
crassirostris: Ack, thanks

test/plugin/test_out_google_cloud.rb
crassirostris: Ack

test/plugin/test_out_google_cloud.rb
igorpeshansky: Sorry, I was asking whether you have to re-create the driver for each entry. Can you do it inside the outer loop?

test/plugin/test_out_google_cloud.rb
igorpeshansky: Whoops, I meant "more than one entry per request"...

test/plugin/test_out_google_cloud.rb
crassirostris: As I've mentioned, I have to create a new driver for each request I want it to issue, since otherwise all entries get bundled to a single request

I would really appreciate if you could advise how to make it issue several requests with one driver

igorpeshansky: Multiple `emit` calls will put entries into a single bundle (request). There is no easy way to force the driver to flush entries, so you would have to re-create the driver for each request. I guess that answers my question, modulo the other comment about testing larger bundles.

crassirostris:
@igorpeshansky Thanks! Addressed all new comments

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -72,6 +72,50 @@ def test_server_error
...
+    setup_gce_metadata_stubs
+    [
+      # Single successful request.
+      [false, 0, 1, 1, [1, 0, 1, 0]],
+      # Several successful requests.
+      [false, 0, 2, 1, [2, 0, 2, 0]],
+      # Single successful request with several entries.
+      [false, 0, 1, 2, [1, 0, 2, 0]],
+      # Single failed request that causes logs to be dropped.
+      [true, 16, 1, 1, [0, 1, 0, 1]],
```
igorpeshansky: Technically we should be using `GRPC::BadStatus::UNAUTHENTICATED`, etc, instead of these int literals. But that's beyond the scope of this PR...

test/plugin/test_out_google_cloud.rb
```diff
@@ -70,6 +70,50 @@ def test_server_error
     assert_equal 1, exception_count
   end
 
+  def test_prometheus_metrics
```
igorpeshansky: As I said before, the code in the non-gRPC and gRPC tests is nearly identical. Can you please add a TODO to refactor it later? Don't put a name -- we'll pick it up.

test/plugin/test_out_google_cloud.rb
```diff
@@ -70,6 +70,50 @@ def test_server_error
...
+      [200, 2, 1, [2, 0, 2, 0]],
+      # Single successful request with several entries.
+      [200, 1, 2, [1, 0, 2, 0]],
+      # Single failed request that causes logs to be dropped.
+      [401, 1, 1, [0, 1, 0, 1]],
+      # Single failed request that escalates without logs being dropped.
+      [500, 1, 1, [0, 1, 0, 0]]
+    ].each do |code, request_count, entry_count, metric_values|
+      setup_prometheus
+      stub_request(:post, WRITE_LOG_ENTRIES_URI)
```
igorpeshansky: We should be able to do this as part of `setup_logging_stubs`. Can you please add a TODO to refactor it later? Don't put a name -- we'll pick it up.

crassirostris:
@igorpeshansky Done, thanks! I guess the refactoring will also allow to easily add tests for other monitoring implementations, since the testing mechanism is pretty similar

test/plugin/test_out_google_cloud.rb
```diff
@@ -70,6 +70,53 @@ def test_server_error
     assert_equal 1, exception_count
   end
 
+  # TODO: The code in the non-gRPC and gRPC tests is nearly identical.
+  # Refactor and remove duplciation.
```
igorpeshansky: Typo: duplciation.

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -72,6 +72,52 @@ def test_server_error
     end
   end
 
+  # TODO: The code in the non-gRPC and gRPC tests is nearly identical.
+  # Refactor and remove duplciation.
```
igorpeshansky: Typo: duplciation.

test/plugin/test_out_google_cloud.rb
igorpeshansky: How about this TODO?

crassirostris:
@igorpeshansky Whoops, thanks for noticing. Do you want me to add a TODO for the status codes?

test/plugin/test_out_google_cloud.rb
igorpeshansky: Sorry about the earlier "How about this TODO?" comment. It was supposed to go here.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -186,6 +188,18 @@ module Constants
                  :default => nil,
                  :secret => true
 
+    # Whether to collect metrics about the plugin usage. Use configuration
+    # parameter monitoring_type to select a monitoring system you want to use.
```
dhrupadb: Let's move the second sentence of this comment to above the next config_param.

test/plugin/constants.rb
```diff
@@ -140,6 +140,11 @@ module Constants
     detect_subservice false
   )
 
+  PROMETHEUS_ENABLE_CONFIG = %(
+    enable_monitoring true
```
dhrupadb: The code above suggests the default for `enable_monitoring` is `false` but it's true here - should they be consistent?

dhrupadb: Never mind. I misread this. please ignore.

lib/fluent/plugin/monitoring.rb
```diff
@@ -0,0 +1,67 @@
...
+require 'grpc'
+require 'json'
+require 'open-uri'
+require 'socket'
+require 'time'
+require 'yaml'
+require 'google/apis'
+require 'google/apis/logging_v2beta1'
+require 'google/logging/v2/logging_pb'
+require 'google/logging/v2/logging_services_pb'
```
crassirostris: Oh, thanks, I wanted to remove them and forgot

lib/fluent/plugin/out_google_cloud.rb
crassirostris: Thb, I don't understand that comment :)

The second sentence is just referring to the different parameter that changes the behavior of this one. It kinda binds them together. Do you want to remove it?

dhrupadb: Ah. I felt the first sentence "Whether to collect metrics ..." referred to the `enable_monitoring` config param whereas the second sentence "Use configuration parameter monitoring_type..." is associated with the config param `monitoring_type` . This is a nit comment.

crassirostris: I'll rewrite the sentence to make it clearer

crassirostris:
@dhrupadb Thanks! PTAL

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -186,6 +188,18 @@ module Constants
                  :default => nil,
                  :secret => true
 
+    # Whether to collect metrics about the plugin usage. The way metrics are
+    # collected and exposed is controlled by the parameter monitoring_type.
```
igorpeshansky: Being extra nit-picky... How about: `The mechanism for collecting and exposing metrics is controlled by the monitoring_type parameter.`?

crassirostris: Sounds cool, I like it

crassirostris:
@igorpeshansky Done

PR #137, opened 2017-07-11
Add docs for set project_id in conf

frankyaorenjie:
Sometimes our GCE and Logging service are in different project, actually we can set this parameter in conf but not documented.
Pull Request: #137


frankyaorenjie:
I signed it!

jkohen:
@qingling128 does this seem reasonable?

qingling128:
This is documented already at https://cloud.google.com/logging/docs/agent/configuration#advanced_non-default_configuration_definitions. The `README.rdoc` has a link to that page.

Issue #126, opened 2017-04-21
Sending operation object to stackdriver logging


kdabir:
As per Stackdriver Logging API, we could pass an `operation` object to connect relevant log lines that are part of one operation. (Reference: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogEntryOperation)

Here are some example log lines:
```
{"timestamp":"2017-04-21T22:24:33.739+05:30","message":"message 1","logger":"LoggerRequestListener","thread":"http-nio-8080-exec-1","severity":"INFO","level_value":20000,"operation":{"producer":"api","id":"nqTUHtJueuw45jcAVpfU"}}
{"timestamp":"2017-04-21T22:24:33.739+05:30","message":"message 1","logger":"LoggerRequestListener","thread":"http-nio-8080-exec-1","severity":"INFO","level_value":20000,"operation":{"producer":"api","id":"nqTUHtJueuw45jcAVpfU"}}
{"timestamp":"2017-04-21T22:24:33.739+05:30","message":"message 3","logger":"LoggerRequestListener","thread":"http-nio-8080-exec-1","severity":"INFO","level_value":20000,"operation":{"producer":"api","id":"nqTUHtJueuw45jcAVpfU"}}
```

But when viewed in Stackdriver Logging, `operation` is nested inside `jsonPayload`. 

![image](https://cloud.githubusercontent.com/assets/735240/25289650/24c95f06-26e8-11e7-898e-579302b711f5.png)

When used similar feature for `httpRequest`, google fluentd plugin moves the object at top level but `operation` gets no such love from the plugin. 

![image](https://cloud.githubusercontent.com/assets/735240/25289659/340159ec-26e8-11e7-88a0-d88a27fe003b.png)

How do we send `operation` object such that it is received at top level by Stackdriver logging?

kdabir:
any updates on this one? I hope the request makes sense. Let me know if I can help with anything.

kdabir:
I am losing hopes on getting updates on this. If it can not be supported, kindly update the issue, we'll stop watching it.

j-walker23:
It's sad no one has responded to this. This was the number one coolest logging feature anyone i have ever talked to loved. Only thing i truly miss from my old appengine days.

[...]

Issue #4, opened 2015-04-16
Support specifying 'message_key' for text payload, 'severity_key' for severity, and 'time_key' for the timestamp


mr-salty:
this is based on @filipjs comment on #3 

  About format none:
  """The key field is “message” by default, but you can specify a different value using the message_key parameter as shown below"""
  is it possible to read that 'message_key' in the plugin and use it here instead of hard-coding 'message'?

I noted that other output plugins support this (see http://docs.fluentd.org/articles/out_file) so we should as well.

mr-salty:
We should also allow specifying the field name for 'severity'.

mr-salty:
We should also add keep_severity_key, etc. - analogous to keep_time_key for http://docs.fluentd.org/articles/in_tail

tonglil:
Is there a way to also specify the key for JSON log messages?
The log driver I use formats the json log message key as `msg` instead of `message`, and I don't really have a way to transform that.

This is in addition to the severity key.

Overall, there should either be some transformers available in SD logging to format log fields into native SD fields better, or SD logging should make it easier to parse logs (like Papertrail).

igorpeshansky:
There are transformers available: http://docs.fluentd.org/articles/filter_record_transformer. However, in some environments (e.g., GKE) users don't have access to the logging agent configuration, so they can't use these.

tonglil:
Yes, this is a GKE environment. 

Ideally these transformers can be applied and adjusted in stackdriver as well.

igorpeshansky:
Unfortunately, this is not possible on GKE today. This is a limitation of the GKE configuration. It's possible to override it by deploying the whole logging subsystem yourself, but there are caveats.
I believe one workaround is to output your logs to a file and run a sidecar container to transform them into the correct form and dump them to stdout. @crassirostris and @piosz may have more recommendations.

crassirostris:
@tonglil GKE eng here. The problem is, we in GKE want to make it work as smoothly as possible and maintain the logging infrastructure for you. When there's a bug in the agent or a new Kubernetes version requires a new way of scraping logs, we fix it for you. If GKE configuration is allowed to be changed, it's not clear how to maintain it on GKE side, i.e update the image or configuration without breaking changes, introduced by users. It would be strange for example if each update would swipe out all the changes you made last time.

[...]

PR #138, opened 2017-07-21
Allow setting LogEntry trace field via record

draffensperger:
This fixes part of https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/126. This is just for the `trace` field, but I plan to do a similar change for `operation`.

This makes the `trace_key` default to `logging.googleapis.com/trace` to not conflict with any users who happen to use `"trace"` as a log entry field.

There is also a `keep_trace_key` setting to control whether the trace value is retained in the `jsonPayload` in addition to being set on the LogEntry itself. That's in line with the discussion at https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/4.

This upgrades the `google-cloud-logging` gem dependency as the new version was needed to support the `trace` field in the LogEntry proto for gRPC.

### DIFF
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -22,7 +22,7 @@ eos
   gem.add_runtime_dependency 'fluentd', '~> 0.10'
   gem.add_runtime_dependency 'googleapis-common-protos', '~> 1.3'
   gem.add_runtime_dependency 'google-api-client', '~> 0.9.0'
-  gem.add_runtime_dependency 'google-cloud-logging', '~> 1.2.1'
+  gem.add_runtime_dependency 'google-cloud-logging', '~> 0.24'
   gem.add_runtime_dependency 'googleauth', '~> 0.4'
   gem.add_runtime_dependency 'grpc', '~> 1.0', '< 1.3'
   gem.add_runtime_dependency 'json', '~> 1.8'
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -22,7 +22,7 @@ eos
   gem.add_runtime_dependency 'fluentd', '~> 0.10'
   gem.add_runtime_dependency 'googleapis-common-protos', '~> 1.3'
   gem.add_runtime_dependency 'google-api-client', '~> 0.9.0'
-  gem.add_runtime_dependency 'google-cloud-logging', '~> 0.23.2'
+  gem.add_runtime_dependency 'google-cloud-logging', '~> 1.2.1'
   gem.add_runtime_dependency 'googleauth', '~> 0.4'
   gem.add_runtime_dependency 'grpc', '~> 1.0', '< 1.3'
   gem.add_runtime_dependency 'json', '~> 1.8'
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -22,8 +22,8 @@ eos
   gem.add_runtime_dependency 'fluentd', '~> 0.10'
   gem.add_runtime_dependency 'googleapis-common-protos', '~> 1.3'
   gem.add_runtime_dependency 'google-api-client', '~> 0.9.0'
-  gem.add_runtime_dependency 'google-cloud-logging', '~> 0.24'
-  gem.add_runtime_dependency 'googleauth', '~> 0.4'
+  gem.add_runtime_dependency 'google-cloud-logging', '0.24.1'
+  gem.add_runtime_dependency 'googleauth', '0.5.1'
   gem.add_runtime_dependency 'grpc', '~> 1.0', '< 1.3'
   gem.add_runtime_dependency 'json', '~> 1.8'
 
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -72,6 +72,9 @@ module Fluent
         service: 'ml.googleapis.com',
         resource_type: 'ml_job'
       }
+
+      # Default value for trace_key config param to set "trace" LogEntry field.
+      DEFAULT_TRACE_KEY = 'logging.googleapis.com/trace'
     end
 
     include self::Constants
@@ -86,9 +89,6 @@ module Fluent
 
     # Address of the metadata service.
     METADATA_SERVICE_ADDR = '169.254.169.254'
-
-    # Default value for trace_key config param to set "trace" LogEntry field.
-    DEFAULT_TRACE_KEY = 'logging.googleapis.com/trace'
 
     # Disable this warning to conform to fluentd config_param conventions.
     # rubocop:disable Style/HashSyntax
@@ -629,12 +629,12 @@ module Fluent
           severity = compute_severity(
             entry_resource.type, record, entry_common_labels)
 
-          # Get trace resource for LogEntry "trace" field per specified config.
-          trace_resource = if @keep_trace_key
-                             record[@trace_key]
-                           else
-                             record.delete(@trace_key)
-                           end
+          # Get fully-qualified trace id for LogEntry "trace" field per config.
+          fq_trace_id = if @keep_trace_key
+                          record[@trace_key]
+                        else
+                          record.delete(@trace_key)
+                        end
 
           if @use_grpc
             entry = Google::Logging::V2::LogEntry.new(
@@ -645,7 +645,7 @@ module Fluent
               ),
               severity: grpc_severity(severity)
             )
-            entry.trace = trace_resource if trace_resource
+            entry.trace = fq_trace_id if fq_trace_id
             # If "seconds" is null or not an integer, we will omit the timestamp
             # field and defer the decision on how to handle it to the downstream
             # Logging API. If "nanos" is null or not an integer, it will be set
@@ -671,7 +671,7 @@ module Fluent
                 nanos: ts_nanos
               }
             )
-            entry.trace = trace_resource if trace_resource
+            entry.trace = fq_trace_id if fq_trace_id
             set_http_request(record, entry)
             set_payload(entry_resource.type, record, entry, is_json)
           end
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -87,6 +87,9 @@ module Fluent
     # Address of the metadata service.
     METADATA_SERVICE_ADDR = '169.254.169.254'
 
+    # Default value for trace_key config param to set "trace" LogEntry field.
+    DEFAULT_TRACE_KEY = 'logging.googleapis.com/trace'
+
     # Disable this warning to conform to fluentd config_param conventions.
     # rubocop:disable Style/HashSyntax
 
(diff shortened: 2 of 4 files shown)


test/plugin/base_test.rb
```diff
@@ -1015,6 +1015,67 @@ def test_http_request_skip_setting_latency
     end
   end
 
+  def test_log_with_trace
+    setup_gce_metadata_stubs
+    message = log_entry(0)
+    trace = 'projects/project-1/traces/1234567890abcdef1234567890abcdef'
+    default_trace_key = 'logging.googleapis.com/trace'
```
qingling128: Can we put `'logging.googleapis.com/trace'` in a constant and reference that in both places?

draffensperger: Done.

fluent-plugin-google-cloud.gemspec
```diff
@@ -22,7 +22,7 @@ eos
   gem.add_runtime_dependency 'fluentd', '~> 0.10'
   gem.add_runtime_dependency 'googleapis-common-protos', '~> 1.3'
   gem.add_runtime_dependency 'google-api-client', '~> 0.9.0'
-  gem.add_runtime_dependency 'google-cloud-logging', '~> 0.23.2'
+  gem.add_runtime_dependency 'google-cloud-logging', '~> 1.2.1'
```
draffensperger: This actually only needs to be upgraded to 0.24.2 as the commit that added the "trace" LogEntry field is https://github.com/GoogleCloudPlatform/google-cloud-ruby/commit/f9fb517bdfa8dedfcb477f9188546d62f5c85e6a.

That means we would be taking in changes incorporated from 12/17/16 to 3/4/17 to the gem. Here's a relevant link in case it helps review those changes: https://github.com/GoogleCloudPlatform/google-cloud-ruby/commits/master?before=0a277ce77bb1c0a4a3eeceddd7f1ab3615740dca+70&path%5B%5D=google-cloud-logging

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -109,6 +109,12 @@ module Constants
     config_param :vm_id, :string, :default => nil
     config_param :vm_name, :string, :default => nil
 
+    # Set values from JSON payload with this key to the "trace" LogEntry field
```
draffensperger: Done.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -109,6 +109,12 @@ module Constants
...
+    config_param :trace_key, :string, :default =>
+      'logging.googleapis.com/trace'
+    # Whether to keep the trace LogEntry field/value in the jsonPayload also
```
draffensperger: Done.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -621,15 +627,21 @@ def write(chunk)
           severity = compute_severity(
             entry_resource.type, record, entry_common_labels)
 
+          # Extract "trace" as LogEntry field according to specified config
```
draffensperger: Done. Also rewrote this comment a bit after the rename of the variable to `trace_resource`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -621,15 +627,21 @@ def write(chunk)
...
+          trace = record[@trace_key]
```
draffensperger: I renamed it to `trace_resource` as per the definition of the field in LogEntry, it's actually a resourceful path for the trace, so something with the format "projects/[PROJECT-ID]/traces/[TRACE-ID]" for Stackdriver Trace, or something else for traces from other tracing systems. See: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -621,15 +627,21 @@ def write(chunk)
...
+          record.delete(@trace_key) unless @keep_trace_key
```
draffensperger: Done.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -621,15 +627,21 @@ def write(chunk)
...
+
           if @use_grpc
-            entry = Google::Logging::V2::LogEntry.new(
+            entry_data = {
```
draffensperger: Done.

fluent-plugin-google-cloud.gemspec
igorpeshansky: Let's take the conservative approach and update to the minimum version. We can always pull in the later one if needed.

test/plugin/base_test.rb
```diff
@@ -1065,14 +1126,14 @@ def setup_ec2_metadata_stubs
 
   def setup_auth_stubs
     # Used when loading credentials from a JSON file.
-    stub_request(:post, 'https://www.googleapis.com/oauth2/v3/token')
+    stub_request(:post, 'https://www.googleapis.com/oauth2/v4/token')
```
igorpeshansky: Was this part of the logging gem update? If we use `0.24` instead, is this still needed?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -621,6 +629,13 @@ def write(chunk)
           severity = compute_severity(
             entry_resource.type, record, entry_common_labels)
 
+          # Get trace resource for LogEntry "trace" field per specified config.
```
igorpeshansky: Let's be careful about using the word "resource" -- it means something specific in the Stackdriver context, and users may get confused. Can we use another term, like "trace id" or "span id", or am I completely confusing the terminology?

dhrupadb: +1. We should avoid resource as much as possible unless it has something specific to do with Stackdriver defined resources (i.e. monitored resources).

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -87,6 +87,9 @@ module Constants
     # Address of the metadata service.
     METADATA_SERVICE_ADDR = '169.254.169.254'
 
+    # Default value for trace_key config param to set "trace" LogEntry field.
+    DEFAULT_TRACE_KEY = 'logging.googleapis.com/trace'
```
draffensperger: Done. Yeah, that allows for cleaner referencing of the constant in the test.

lib/fluent/plugin/out_google_cloud.rb
draffensperger: Done. Switched to `fq_traced_id` for fully-qualified trace id and changed the comment accordingly.

test/plugin/base_test.rb
draffensperger: Yes, it has to do with an upgrade of the `google-gax` gem that was done by the `google-cloud-logging` gem on 2/18/17, see: https://github.com/GoogleCloudPlatform/google-cloud-ruby/commit/4e4c7f42e43533d729aa65dd5a8fde47160a8c02

That `google-gax` gem version depends on `googleauth` "~> 0.5", see https://rubygems.org/gems/google-gax/versions/0.8.0

And the `googleauth` gem 0.5.2 on 7/19/17 (so pretty recently) did an upgrade to use the /v4/ auth endpoint, see https://github.com/google/google-auth-library-ruby/blob/master/CHANGELOG.md#052-20170719

What I did was pin the `googleauth` gem to `0.5.1`, and that did allow me to switch this back to `/v3/` while still upgrading the `google-cloud-logging` gem so it gets the new `trace` field.

fluent-plugin-google-cloud.gemspec
draffensperger: OK. This actually can work with exactly version 0.24.1, which is slightly older. I changed the gemspec to depend on that exact version.

test/plugin/base_test.rb
```diff
@@ -1015,6 +1015,67 @@ def test_http_request_skip_setting_latency
     end
   end
 
+  def test_log_with_trace
+    setup_gce_metadata_stubs
+    message = log_entry(0)
+    trace = 'projects/project-1/traces/1234567890abcdef1234567890abcdef'
+    default_trace_key = DEFAULT_TRACE_KEY
```
qingling128: We should be able to just use `DEFAULT_TRACE_KEY` and get rid of the variable `default_trace_key`.

test/plugin/base_test.rb
```diff
@@ -1015,6 +1015,67 @@ def test_http_request_skip_setting_latency
     end
   end
 
+  def test_log_with_trace
```
qingling128: Can we split the tests for `keep_trace_key` and `trace_key` into two separate unit tests?

test/plugin/base_test.rb
draffensperger: Done.

test/plugin/base_test.rb
draffensperger: Done. Yeah, that test was pretty big.

qingling128:
Seems some rebasing is needed to resolve the gem spec conflict.

draffensperger:
I squashed this to a single commit and rebased it on master to fix the gemspec conflict.

igorpeshansky:
Just as an FYI, we generally prefer to not squash PR branches (we use GitHub's "squash and merge"). That way the review history stays mostly intact. Since you've already squashed, I won't make you undo that, but let's not do it in the future.

draffensperger:
Yeah, that makes sense on preserving the history - will avoid that in the future.

draffensperger:
I was able to recover the original commits from my reflog, and then fixed the merge conflict with a new commit of merging in master instead.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -109,6 +112,11 @@ module Constants
     config_param :vm_id, :string, :default => nil
     config_param :vm_name, :string, :default => nil
 
+    # Set values from JSON payload with this key to the "trace" LogEntry field.
+    config_param :trace_key, :string, :default => DEFAULT_TRACE_KEY
+    # Whether to also keep the trace key/value in the payload.
+    config_param :keep_trace_key, :bool, :default => false
```
dhrupadb: Just curious - are there any situations we'd not want this on by default?

qingling128: I thought in most cases we want it `false`. Once we set the `trace` field in `LogEntry`, it's no longer needed in the json payload.

dhrupadb: Ah okay understood, I guess then my question is are there any situations where we'd want it both the Payload and the Entry? I'm trying to assess if this configuration is needed.

draffensperger: I would guess it's mostly useful if a customer were already annotating their logs with traces, and wanted to keep their jsonPayload consistent, but also wanted the trace LogEntry field to be set. In that case, they would also set the `trace_key` config field.

Because the default `trace_key` is specific to the Google cloud logging API, customers would need to annotate their logs intentionally with it with the purpose of having the LogEntry trace field set, and so having it retained in the jsonPayload in that case would just add extra weight to the log entries. So that to me indicates that `false` is a good default.

draffensperger: My reasoning for creating the setting was mostly going off this comment: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/4#issuecomment-245671574

I personally would be fine with removing the `keep_trace_key` setting if you think that's best.

dhrupadb: Nope that makes perfect sense to me! Thanks Dave.

PR #140, opened 2017-07-21
Pin googleauth gem version at 0.5.1 because of https://github.com/google/google-auth-library-ruby/issues/113.

qingling128:
Fixes https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/139 by working around https://github.com/google/google-auth-library-ruby/issues/113.

### DIFF
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -23,7 +23,9 @@ eos
   gem.add_runtime_dependency 'googleapis-common-protos', '~> 1.3'
   gem.add_runtime_dependency 'google-api-client', '~> 0.9.0'
   gem.add_runtime_dependency 'google-cloud-logging', '~> 0.23.2'
-  gem.add_runtime_dependency 'googleauth', '~> 0.4'
+  # Pin the gem at earlier versions because this bug was introduced in v0.5.2:
+  # https://github.com/google/google-auth-library-ruby/issues/113
+  gem.add_runtime_dependency 'googleauth', '~> 0.4', '< 0.5.2'
   gem.add_runtime_dependency 'grpc', '~> 1.0', '< 1.3'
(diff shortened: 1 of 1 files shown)


fluent-plugin-google-cloud.gemspec
```diff
@@ -23,7 +23,9 @@ eos
   gem.add_runtime_dependency 'googleapis-common-protos', '~> 1.3'
   gem.add_runtime_dependency 'google-api-client', '~> 0.9.0'
   gem.add_runtime_dependency 'google-cloud-logging', '~> 0.23.2'
-  gem.add_runtime_dependency 'googleauth', '~> 0.4'
+  # Pin the gem at earlier versions because this bug was introduced in v0.5.2:
```
igorpeshansky: If you have to comment at all, I'd say:
```ruby
# Avoid v0.5.2 because of a bug: https://github.com/google/google-auth-library-ruby/issues/113
```
But you may not even need that, because `git blame` clearly shows the commit that added this constraint, which will refer to this PR (#140), which will then refer to the upstream bug (https://github.com/google/google-auth-library-ruby/issues/113) in the description.

qingling128: Right. Just removed the comment.

igorpeshansky:
Looks like this may not be needed anymore, according to https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/139#issuecomment-317071906 ...

qingling128:
https://github.com/google/google-auth-library-ruby/releases/tag/v0.5.3 fixed the file permission issue. This is no longer needed. Closing it.

igorpeshansky:
We still need this, because 0.5.3 also [switched to a new OAuth URI](https://github.com/google/google-auth-library-ruby/commit/f2b5b8ac98cf08a00c747d42aa937b819b5bcfbd), which breaks our tests.

PR #141, opened 2017-07-24
Add Ruby 2.4 to Travis Build config

dhrupadb:


igorpeshansky:
@dhrupadb Looks like our version of `webmock` is too old. We depend on `1.17`, which pulls in `1.24.6`, but support for Ruby 2.4 has only been added only in `2.3.1`.

PR #144, opened 2017-07-31
Convert timestamp Strings with Integer values to Integers.

igorpeshansky:
Fixes the case when timestamp components are parsed out of logs as strings.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -636,6 +636,16 @@ module Fluent
                           record.delete(@trace_key)
                         end
 
+          ts_secs = begin
+                      Integer ts_secs
+                    rescue ArgumentError, TypeError
+                      ts_secs
+                    end
+          ts_nanos = begin
+                       Integer ts_nanos
+                     rescue ArgumentError, TypeError
+                       ts_nanos
+                     end
           if @use_grpc
             entry = Google::Logging::V2::LogEntry.new(
               labels: entry_common_labels,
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -636,6 +636,16 @@ def write(chunk)
...
 
+          ts_secs = begin
+                      Integer ts_secs
+                    rescue ArgumentError, TypeError
+                      ts_secs
+                    end
+          ts_nanos = begin
+                       Integer ts_nanos
+                     rescue ArgumentError, TypeError
+                       ts_nanos
```
qingling128: For `nanos`, we can set it to `0` in case of an error, and remove the line down below:

```
ts_nanos = 0 unless ts_nanos.is_a?(Integer)
```

igorpeshansky: I've tried that earlier, but it caused the tests to fail. Our tests expect the code to preserve the value if it's not an integer.

qingling128: Ah, I see. That's probably me who added that test to be explicit. The downstream Logging API should convert it to `0` anyway since it's ultimately stored as protos. But we will leave it to the downstream component to decide.

qingling128:
LGTM.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: I've opened a bug to track this internally.

PR #145, opened 2017-08-07
Get source_location, operation and labels from the record if any are available.

### LINKED ISSUE #126: Sending operation object to stackdriver logging

arthurdarcet:
Dispatch the data to the correct fields for `sourceLocation` and `operation`, like it's done for `httpRequest`

I have also merged the `labels` into the record labels, if no conflict is present

fixes #126

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -95,7 +95,7 @@ module Fluent
         .map { |consts| [consts[:resource_type], consts[:metadata_attributes]] }
         .to_h
 
-      # Default values for JSON payload keys to set the "trace "trace",
+      # Default values for JSON payload keys to set the "trace",
       # "sourceLocation", "operation" and "labels" fields in the LogEntry.
       DEFAULT_PAYLOAD_KEY_PREFIX = 'logging.googleapis.com'
       DEFAULT_LABELS_KEY = "#{DEFAULT_PAYLOAD_KEY_PREFIX}/labels"
@@ -1257,11 +1257,8 @@ module Fluent
             begin
               casted_value = send(cast_fn, value)
             rescue TypeError
-              log = <<-eos
-                Failed to #{cast_fn} for {#field_name}.#{original_key}
-                with value #{value.inspect}.
-              eos
-              @log.error log, err
+              @log.error "Failed to #{cast_fn} for #{field_name}." \
+                         "#{original_key} with value #{value.inspect}.", err
               next
             end
             next if casted_value.nil?
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -522,7 +522,7 @@ module Fluent
           fq_trace_id = record.delete(@trace_key)
           entry.trace = fq_trace_id if fq_trace_id
 
-          set_log_entry_field(record, entry)
+          set_log_entry_fields(record, entry)
           set_labels(record, entry)
 
           if @use_grpc
@@ -1242,7 +1242,7 @@ module Fluent
       end
     end
 
-    def set_log_entry_field(record, entry)
+    def set_log_entry_fields(record, entry)
       LOG_ENTRY_FIELDS_MAP.each do |field_name, config|
         payload_key, subfields, grpc_class, non_grpc_class = config
         begin
@@ -1257,6 +1257,11 @@ module Fluent
             begin
               casted_value = send(cast_fn, value)
             rescue TypeError
+              log = <<-eos
+                Failed to #{cast_fn} for {#field_name}.#{original_key}
+                with value #{value.inspect}.
+              eos
+              @log.error log, err
               next
             end
             next if casted_value.nil?
@@ -1274,7 +1279,7 @@ module Fluent
             output.send("#{key}=", value)
           end
 
-          record.delete(payload_key) if record[payload_key].empty?
+          record.delete(payload_key) if fields.empty?
 
           entry.send("#{field_name}=", output)
         rescue StandardError => err
@@ -1284,16 +1289,17 @@ module Fluent
     end
 
     def set_labels(record, entry)
-      return nil unless record[@labels_key].is_a?(Hash)
+      record_labels = record[@labels_key]
+      return nil unless record_labels.is_a?(Hash)
 
-      record[@labels_key].each do |key, value|
+      record_labels.each do |key, value|
         unless entry.labels.key?(key)
-          record[@labels_key].delete(key)
+          record_labels.delete(key)
           entry.labels[key] = value
         end
       end
 
-      record.delete(@labels_key) if record[@labels_key].empty?
+      record.delete(@labels_key) if record_labels.empty?
     end
 
     # Values permitted by the API for 'severity' (which is an enum).
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -446,6 +446,9 @@ module Fluent
               )
             end
             set_http_request(record, entry)
+            set_source_location(record, entry)
+            set_operation(record, entry)
+            set_labels(record, entry)
             set_payload_grpc(entry_resource.type, record, entry, is_json)
           else
             # Remove the labels if we didn't populate them with anything.
@@ -461,6 +464,9 @@ module Fluent
             )
             entry.trace = fq_trace_id if fq_trace_id
             set_http_request(record, entry)
+            set_source_location(record, entry)
+            set_operation(record, entry)
+            set_labels(record, entry)
             set_payload(entry_resource.type, record, entry, is_json)
           end
 
@@ -1240,6 +1246,62 @@ module Fluent
 
       record.delete('httpRequest') if input.empty?
       entry.http_request = output
+    end
+
+    def set_source_location(record, entry)
+      return nil unless record['sourceLocation'].is_a?(Hash)
+      input = record['sourceLocation']
+      if @use_grpc
+        output = Google::Logging::V2::LogEntrySourceLocation.new
+      else
+        output = Google::Apis::LoggingV2beta1::LogEntrySourceLocation.new
+      end
+
+      file = input.delete('file')
+      output.file = file unless file.nil?
+      function = input.delete('function')
+      output.function = function unless function.nil?
+      line = input.delete('line')
+      output.line = line unless line.nil?
+
+      record.delete('sourceLocation') if input.empty?
+      entry.source_location = output
+    end
+
+    def set_operation(record, entry)
+      return nil unless record['operation'].is_a?(Hash)
+
+      input = record['operation']
+      if @use_grpc
+        output = Google::Logging::V2::LogEntryOperation.new
+      else
+        output = Google::Apis::LoggingV2beta1::LogEntryOperation.new
+      end
+
+      id = input.delete('id')
+      output.id = id unless id.nil?
+      producer = input.delete('producer')
+      output.producer = producer unless producer.nil?
+      first = input.delete('first')
+      output.first = first unless first.nil?
+      last = input.delete('last')
+      output.last = last unless last.nil?
+
+      record.delete('operation') if input.empty?
+      entry.operation = output
+    end
+
+    def set_labels(record, entry)
+      return nil unless record['labels'].is_a?(Hash)
+
+      record['labels'].each do |key, value|
+        unless entry.labels.key?(key)
+          record['labels'].delete(key)
+          entry.labels[key] = value
+        end
+      end
+
+      record.delete('labels') if record['labels'].empty?
     end
 
     # Values permitted by the API for 'severity' (which is an enum).
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -95,12 +95,91 @@ module Fluent
         .map { |consts| [consts[:resource_type], consts[:metadata_attributes]] }
         .to_h
 
-      # Default value for trace_key config param to set the "trace",
+      # Default values for JSON payload keys to set the "trace "trace",
       # "sourceLocation", "operation" and "labels" fields in the LogEntry.
       DEFAULT_LABELS_KEY = 'logging.googleapis.com/labels'
+      DEFAULT_HTTP_REQUEST_KEY = 'httpRequest'
       DEFAULT_OPERATION_KEY = 'logging.googleapis.com/operation'
       DEFAULT_SOURCE_LOCATION_KEY = 'logging.googleapis.com/sourceLocation'
       DEFAULT_TRACE_KEY = 'logging.googleapis.com/trace'
+
+      # Map from each field name under LogEntry to corresponding variables
+      # required to perform field value extraction from the log record.
+      LOG_ENTRY_FIELDS_MAP = {
+        'http_request': {
+          # The config to specify label name for field extraction from record.
+          payload_key: '@http_request_key',
+          # Map from subfields' names to their types.
+          fields: [
+            # subfield key in the payload, destination key, cast lambda (opt)
+            %w(requestMethod request_method),
+            %w(requestUrl request_url),
+            %w(requestSize request_size),
+            ['status', 'status', -> (v, _) { return v.to_i }],
+            ['responseSize', 'response_size', -> (v, _) { return v.to_i }],
+            %w(userAgent user_agent),
+            %w(remoteIp remote_ip),
+            %w(referer referer),
+            %w(cacheHit cache_hit),
+            %w(cacheValidatedWithOriginServer
+               cache_validated_with_origin_server),
+            ['latency', 'latency', lambda do |latency, binding|
+              # Parse latency.
+              # If no valid format is detected, skip setting latency.
+              # Format: whitespace (opt.) + integer + point & decimal (opt.)
+              #       + whitespace (opt.) + "s" + whitespace (opt.)
+              # e.g.: "1.42 s"
+              rx = binding.eval('@compiled_http_latency_regexp')
+              match = rx.match(latency)
+              if match
+                # Split the integer and decimal parts in order to calculate
+                # seconds and nanos.
+                seconds = match['seconds'].to_i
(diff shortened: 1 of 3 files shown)


qingling128:
Similar to https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/138, we should use "logging.googleapis.com" as the prefix when detecting these fields in the record. Otherwise it might conflict with users' labels. Also it might break existing clients' integrations if they expect the `operation`, `sourceLocation` and `labels` fields to show up in `jsonPayload`.

qingling128:
BTW, @draffensperger has some plan to work on adding the support for `operation` field as well. (Not sure if the work has been started yet). Might wanna cooperate.

draffensperger:
I've not started the work yet to support the `operation` field, but I have been intending to do some and I did do it for the `trace` field (see https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/138).

One thing I would recommend from looking at the code would be to make this check for `logging.googleapis.com/sourceLocation`, `logging.googleapis.com/operation` and `logging.googleapis.com/labels` to make this more backwards-compatible for customers who may happen to set the more generic `sourceLocation`, `operation` and `labels` in their logs. Also, it might be nice to make config settings like `source_location_key` and `keep_source_location` similar to what I did in the `trace` PR above.

arthurdarcet:
I will update the PR with the naming update, and add config parameters to change them ; as a side note, I think the `httpRequest` key parsing should be consistent with `sourceLocation`/…, so imo it should also be renamed to `logging.googleapis.com/httpRequest`. But I'm not sure how a deprecation warning should be added for the old `httpRequest` key (maybe adding a `_warning: rename httpRequest to logging.…` field in the jsonPayload?)

For the `keep_…` parameters, are they really needed? I'm not really clear on why it was needed for the trace/severity/time keys

draffensperger:
I agree with you that ideally `httpRequest` would be supported in a similar way (perhaps for a while supporting either a prefixed or non-prefixed key), but that adds some extra complexity.

Regarding the `keep_X_key` stuff, here's a conversation about it for the `trace` field:
https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/138/files/9775f49f470cb62d3f379fe892926171465949fc#r130136705

I personally would be fine not having those settings, but I'll leave that up to the other reviewers. Right now only `trace` has that option, so maybe it would be simpler to just remove `keep_trace_field`.

qingling128:
I'm personally leaning towards just removing the `keep_X_key` stuff. There were some complaints from customers about having too many options in our product. It might not be true for all cases. But the idea is worth some thoughts. Sometimes less is more.

draffensperger:
Also, I created a PR for removing the `keep_trace_key` config option: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/147

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1088,6 +1103,62 @@ def set_http_request(record, entry)
       entry.http_request = output
     end
 
+    def set_source_location(record, entry)
```
qingling128: Seems we are doing very similar things for `source_location`, `trace` and `operation`. Can we use some shared utility method to capture the logic and leave the dynamic part separate as a map. I'm thinking of something like this:

Map definition:
```
      # Map from each field name under LogEntry to corresponding variables
      # required to perform field value extraction from the log record.
      LOG_ENTRY_FIELDS_MAP = {
        source_location: [
          # The config to specify label name for field extraction from record.
          '@source_location_key',
          # Map from subfields' names to their types.
          {
            'file' => String.name,
            'function' => String.name,
            'line' => Numeric.name
          },
          # The grpc version class name.
          'Google::Logging::V2::LogEntrySourceLocation',
          # The non-grpc version class name.
          'Google::Apis::LoggingV2beta1::LogEntrySourceLocation'
        ],
        trace: ...,
        operation: ...
      }
```

Set fields in the main method:
```
    [:source_location, :trace, :operation].each do |field_name|
      set_log_entry_field(record, entry, field)
    end
```
Helper method:
```
    def set_log_entry_field(record, entry, field_name)
      field_name_config_variable, subfields_map, grpc_class_name,
        non_grpc_class_name = LOG_ENTRY_FIELDS_MAP[field_name]
      label_name = instance_variable_get(field_name_config_variable)
      subfields = fields_to_labels(record[label_name],
                                                    subfields_map.keys.map { |l| [l, l] }.to_h)
      record.delete(label_name) if record[label_name].empty?

      return unless subfields

      if @use_grpc
        output = Object.const_get(grpc_class_name).new
      else
        output = Object.const_get(non_grpc_class_name).new
      end
      subfields.each do |key, value|
        next unless value
        # Convert the string value to String or Integer objects.
        formatted_value = case subfields_map[key]
          when String.name
            value
          when Numeric.name
            value.to_i
          end
        output.send("#{key}=", formatted_value)
      end
      entry.send("#{field_name}=", output)
    rescue StandardError => err
      @log.error "Failed to set log entry field for #{field_name}.", err
    end
```

test/plugin/base_test.rb
```diff
@@ -1020,6 +1020,134 @@ def test_http_request_skip_setting_latency
     end
   end
 
+  def test_source_location_from_record
```
qingling128: Similar here. Can we move the common logic to verify these fields to shared methods? Then the test itself simply specify the dynamic part (input and output).

e.g.
```
def verify_log_entry_subfield(.., .., ..)
...
end

def test_log_entry_operation_field
   ...
   [<SOME_DATA_TO_TEST_FROM_RECORD>, <SOME_DATA_TO_TEST_PARTIAL>, <SOME_DATA_TO_TEST_NON_HASH>].each do |...
     verify_log_entry_subfield(...
   end
    ...
end

def test_log_entry_source_location_field
   ...
   [<SOME_DATA_TO_TEST_FROM_RECORD>, <SOME_DATA_TO_TEST_PARTIAL>, <SOME_DATA_TO_TEST_NON_HASH>].each do |...
     verify_log_entry_subfield(...
   end
    ...
end

def test_log_entry_trace_field
   ...
   [<SOME_DATA_TO_TEST_FROM_RECORD>, <SOME_DATA_TO_TEST_PARTIAL>, <SOME_DATA_TO_TEST_NON_HASH>].each do |...
     verify_log_entry_subfield(...
   end
    ...
end

**Comment by arthurdarcet** [comment]
Thanks for the reviews!

@draffensperger : I fixed the comment, thanks
@qingling128 : I factorised most of the code, including the httpRequest handling ; but i left trace out of it (it's not really the same structure here, it's a literal at the record level and not a nested Hash of subfields.
Also, I'm not sure what you mean by "We normally just create a feature branch in our general work flow". Would you like me to rebase this against another branch in the repo?

File: test/plugin/base_test.rb [hunk 3]
```diff
@@ -954,11 +962,54 @@ def test_http_request_skip_setting_latency
       end
       verify_log_entries(1, COMPUTE_PARAMS, 'httpRequest') do |entry|
         assert_equal HTTP_REQUEST_MESSAGE, entry['httpRequest'], entry
-        assert_nil get_fields(entry['jsonPayload'])['httpRequest'], entry
+        fields = get_fields(entry['jsonPayload'])
```
qingling128: Let's not change the behavior here. If `latency` comes in as invalid format, we want to throw away the data.

If we change the order of the code in the `set_log_entry_fields` to delete the field first before trying to cast it, this test should pass without being modified. Something like:

```
            value = fields.delete(original_key) # Delete it first. If the casting later returns nil, we just throw away the invalid data.
            next if value.nil?
            casted_value = send(cast_fn, value)
            next if casted_value.nil?
            extracted_fields[destination_key] = casted_value
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1175,71 +1243,55 @@ def compute_severity(resource_type, record, entry_common_labels)
...
+    def set_log_entry_field(record, entry, field_name, cfg)
+      payload_key = instance_variable_get(cfg[:payload_key])
+      return unless record[payload_key].is_a?(Hash)
+      subfields = {}
+      cfg[:fields].each do |key, destination_key, cast_fn|
+        value = record[payload_key][key]
+        next if value.nil?
+        unless cast_fn.nil?
+          begin
+            value = send(cast_fn, value)
```
qingling128: Maybe give this one a new name `casted_value` to improve readability?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1175,71 +1243,55 @@ def compute_severity(resource_type, record, entry_common_labels)
...
-              seconds: latency_seconds,
-              nanos: latency_nanos
-            )
-          else
-            output.latency = {
-              seconds: latency_seconds,
-              nanos: latency_nanos
-            }.delete_if { |_, v| v == 0 }
-          end
+      subfields.each do |k, v|
```
qingling128: We normally try to avoid one-letter variable names if possible:

```
          subfields.each do |key, value|
            output.send("#{key}=", value)
          end
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1337,6 +1389,40 @@ def grpc_severity(severity)
       severity
     end
 
+    def parse_int(v)
```
qingling128: We normally try to avoid one-letter variable names if possible. So can we do this instead? And as mentioned in the other comment, let's add a `parse_string` method as well.

```
    def parse_string(value)
      value.to_s
    end

    def parse_int(value)
      value.to_i
    end

    def parse_bool(v)
      [true, 'true', 1].include?(value)
    end
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1175,71 +1243,55 @@ def compute_severity(resource_type, record, entry_common_labels)
...
+      payload_key = instance_variable_get(cfg[:payload_key])
+      return unless record[payload_key].is_a?(Hash)
+      subfields = {}
+      cfg[:fields].each do |key, destination_key, cast_fn|
+        value = record[payload_key][key]
+        next if value.nil?
+        unless cast_fn.nil?
+          begin
+            value = send(cast_fn, value)
+          rescue TypeError
```
arthurdarcet: actually, no, because the rescue at the end of the method completely stops trying to set anything on the entry, so an invalid latency would drop the entire `httpRequest` part of the payload (and even worse, the `httpRequest` would not even be in the jsonPayload anymore because we already deleted the other fields

arthurdarcet:
@qingling128 Apart from the begin/rescue issue (I left a comment on your review), I fixed all the other points

arthurdarcet:
Looks like protobuf 3.4 breaks the travis build. Everything is passing in local

test/plugin/base_test.rb
```diff
@@ -859,46 +859,57 @@ def test_dataproc_log
...
-    setup_gce_metadata_stubs
-    setup_logging_stubs do
-      d = create_driver
-      d.emit('httpRequest' => HTTP_REQUEST_MESSAGE)
-      d.run
-    end
-    verify_log_entries(1, COMPUTE_PARAMS, 'httpRequest') do |entry|
-      assert_equal HTTP_REQUEST_MESSAGE, entry['httpRequest'], entry
-      assert_nil get_fields(entry['jsonPayload'])['httpRequest'], entry
+  def test_subfields_from_record
```
qingling128: Can we keep these fields in separate unit tests so that when tests fail, it's more obvious what we are testing? Something like:

```
  def test_log_entry_location_field_from_record
    verify_subfields_from_record(DEFAULT_HTTP_REQUEST_KEY)
  end

  def test_log_entry_location_field_from_record
    verify_subfields_from_record(DEFAULT_SOURCE_LOCATION_KEY)
  end

  def test_log_entry_location_field_from_record
    verify_subfields_from_record(DEFAULT_OPERATION_KEY)
  end

  // Below can be in the private section.
  def verify_subfields_from_record(payload_key)
    destination_key, payload_value = LOG_ENTRY_SUBFIELDS_PARAMS[payload_key]
    @logs_sent = []
    setup_gce_metadata_stubs
    setup_logging_stubs do
      d = create_driver
      d.emit(payload_key => payload_value)
      d.run
    end
    verify_log_entries(1, COMPUTE_PARAMS, destination_key) do |entry|
      assert_equal payload_value, entry[destination_key], entry
      fields = get_fields(entry['jsonPayload'])
      assert_nil fields[payload_key], entry
    end
  end

...

  LOG_ENTRY_SUBFIELDS_PARAMS = {
    # payload key, destination key, payload value
    DEFAULT_HTTP_REQUEST_KEY => ['httpRequest', HTTP_REQUEST_MESSAGE],
    DEFAULT_SOURCE_LOCATION_KEY => ['sourceLocation', SOURCE_LOCATION_MESSAGE],
    DEFAULT_OPERATION_KEY => ['operation', OPERATION_MESSAGE]
  }
``

File: lib/fluent/plugin/out_google_cloud.rb [hunk 9]
```diff
@@ -1175,71 +1242,58 @@ def compute_severity(resource_type, record, entry_common_labels)
...
+          record.delete(payload_key) if record[payload_key].empty?
+
+          entry.send("#{field_name}=", output)
+        rescue StandardError => err
+          @log.error "Failed to set log entry field for #{field_name}.", err
         end
       end
+    end
+
+    def set_labels(record, entry)
```
qingling128: Seems like we can use a variable for `record[@labels_key]`:

```
    def set_labels(record, entry)
      record_labels = record[@labels_key]
      return nil unless record_labels.is_a?(Hash)

      record_labels.each do |key, value|
        unless entry.labels.key?(key)
          record_labels.delete(key)
          entry.labels[key] = value
        end
      end

      record.delete(@labels_key) if record_labels.empty?
    end
```

arthurdarcet:
@qingling128 all done

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1175,71 +1242,64 @@ def compute_severity(resource_type, record, entry_common_labels)
...
+          next unless fields.is_a?(Hash)
+
+          extracted_subfields = subfields.each_with_object({}) \
+            do |(original_key, destination_key, cast_fn), extracted_fields|
+            value = fields.delete(original_key)
+            next if value.nil?
+            begin
+              casted_value = send(cast_fn, value)
+            rescue TypeError
+              log = <<-eos
```
qingling128: Seems `eos` will print the extra spaces and new lines.

```
$ cat test.rb
              log = <<-eos
                Failed to cast_fn for field_name.original_key
                with value value.inspect.
              eos
puts log
$ ruby test.rb
                Failed to cast_fn for field_name.original_key
                with value value.inspect.
```
We could instead try:
```
              @log.error "Failed to #{cast_fn} for #{field_name}." \
                         "#{original_key} with value #{value.inspect}.", err
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -95,8 +95,65 @@ module Constants
         .map { |consts| [consts[:resource_type], consts[:metadata_attributes]] }
         .to_h
 
-      # Default value for trace_key config param to set "trace" LogEntry field.
-      DEFAULT_TRACE_KEY = 'logging.googleapis.com/trace'
+      # Default values for JSON payload keys to set the "trace "trace",
```
qingling128: `"trace "trace"` => `"trace"`.

arthurdarcet: done

lib/fluent/plugin/out_google_cloud.rb
arthurdarcet: done

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1262,6 +1262,8 @@ def set_log_entry_fields(record, entry)
                 with value #{value.inspect}.
```
qingling128: Shouldn't we get rid of line 1260 - 1264 then?

arthurdarcet: yes that was a mistake. Thanks!

qingling128:
Please do "Squash and merge" when you merge. Also consolidate all commit messages into one.

arthurdarcet:
@qingling128 was this last message for me? I can squash my commits on my branch, but I don't have any rights on this repo so I can't do the merge

draffensperger:
I remember from my PR we wanted to keep the commit history to give the code review comments context.

@qingling128 could you do the GitHub squash and merge?

qingling128:
Done.

igorpeshansky:
@arthurdarcet The support for `labels` has been reverted in #151, because it requires deeper discussion that the other fields and we wanted to keep it separate. We will reintroduce it as a separate PR and discuss.

qingling128:
The extraction of labels field from the record is kind of a major feature, and should go through product first. So that part of this PR has been reverted, pending for official product review.

PR #146, opened 2017-08-11
Make JSON detection a config option, rather than GKE-only.

igorpeshansky:
Also adds `'msg'` as a supported text payload key for JSON detection.
@meltsufin FYI. Should allow #89 to be fixed on the AppEngine side.
@crassirostris FYI.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -198,6 +198,11 @@ module Fluent
       DEFAULT_SOURCE_LOCATION_KEY
     config_param :trace_key, :string, :default => DEFAULT_TRACE_KEY
 
+    # Whether to try to detect if the record is a text log entry with JSON
+    # content that needs to be parsed.
+    config_param :detect_json, :bool, :default => false
+    # TODO(igorpeshansky): Add a parameter for the text field in the payload.
+
     # Whether to try to detect if the VM is owned by a "subservice" such as App
     # Engine of Kubernetes, rather than just associating the logs with the
     # compute service of the platform. This currently only has any effect when
@@ -443,19 +448,20 @@ module Fluent
           entry_common_labels = \
             group_common_labels.merge(extracted_common_labels)
 
-          if entry_resource.type == CONTAINER_CONSTANTS[:resource_type]
+          is_json = false
+          if @detect_json
             # Save the timestamp if available, then clear it out to allow for
             # determining whether we should parse the log or message field.
-            timestamp = record.key?('time') ? record['time'] : nil
-            record.delete('time')
+            timestamp = record.delete('time')
             # If the log is json, we want to export it as a structured log
             # unless there is additional metadata that would be lost.
-            is_json = false
-            if record.length == 1 && record.key?('log')
-              record_json = parse_json_or_nil(record['log'])
-            end
-            if record.length == 1 && record.key?('message')
-              record_json = parse_json_or_nil(record['message'])
+            record_json = nil
+            if record.length == 1 then
+              ['log', 'message', 'msg'].each do |field|
+                if record.key?(field)
+                  record_json = parse_json_or_nil(record[field])
+                end
+              end
             end
             unless record_json.nil?
               record = record_json
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -456,8 +456,8 @@ module Fluent
             # If the log is json, we want to export it as a structured log
             # unless there is additional metadata that would be lost.
             record_json = nil
-            if record.length == 1 then
-              ['log', 'message', 'msg'].each do |field|
+            if record.length == 1
+              %w(log message msg).each do |field|
                 if record.key?(field)
                   record_json = parse_json_or_nil(record[field])
                 end
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -305,20 +305,83 @@ module BaseTest
   def test_structured_payload_json_log
     setup_gce_metadata_stubs
     setup_logging_stubs do
-      d = create_driver
-      json_string = '{"msg": "test log entry 0", "tag2": "test", "data": 5000}'
+      d = create_driver(APPLICATION_DEFAULT_CONFIG)
+      json_string = '{"msg": "test log entry 0", "tag2": "test", ' \
+                    '"data": 5000, "some_null_field": null}'
       d.emit('message' => 'notJSON ' + json_string)
+      d.emit('log' => 'notJSON ' + json_string)
+      d.emit('msg' => 'notJSON ' + json_string)
       d.emit('message' => json_string)
-      d.emit('message' => "\t" + json_string)
-      d.emit('message' => '  ' + json_string)
+      d.emit('log' => json_string)
+      d.emit('msg' => json_string)
+      d.emit('message' => "  \r\n \t" + json_string)
+      d.emit('log' => "  \r\n \t" + json_string)
+      d.emit('msg' => "  \r\n \t" + json_string)
       d.run
     end
-    verify_log_entries(4, COMPUTE_PARAMS, '') do |entry|
-      assert entry.key?('textPayload'), 'Entry did not have textPayload'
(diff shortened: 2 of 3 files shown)


test/plugin/base_test.rb
```diff
@@ -305,20 +305,83 @@ def test_structured_payload_malformatted_log
...
+      d.emit('msg' => json_string)
+      d.emit('message' => "  \r\n \t" + json_string)
+      d.emit('log' => "  \r\n \t" + json_string)
+      d.emit('msg' => "  \r\n \t" + json_string)
+      d.run
+    end
+    log_index = 0
+    verify_log_entries(9, COMPUTE_PARAMS, '') do |entry|
+      log_index += 1
+      if log_index % 3 == 1
```
crassirostris: Maybe better keep test configuration and test logic separate? I.e. instead of this assertion have a bool flag in the test configuration

test/plugin/base_test.rb
```diff
@@ -305,20 +305,83 @@ def test_structured_payload_malformatted_log
   def test_structured_payload_json_log
     setup_gce_metadata_stubs
     setup_logging_stubs do
-      d = create_driver
-      json_string = '{"msg": "test log entry 0", "tag2": "test", "data": 5000}'
+      d = create_driver(APPLICATION_DEFAULT_CONFIG)
+      json_string = '{"msg": "test log entry 0", "tag2": "test", ' \
+                    '"data": 5000, "some_null_field": null}'
+      d.emit('message' => 'notJSON ' + json_string)
```
crassirostris: I think the test is missing for the condition that asserts the existence of one field only

test/plugin/base_test.rb
```diff
@@ -305,20 +305,83 @@ def test_structured_payload_malformatted_log
...
     end
-    verify_log_entries(4, COMPUTE_PARAMS, '') do |entry|
-      assert entry.key?('textPayload'), 'Entry did not have textPayload'
+    log_index = 0
+    verify_log_entries(9, COMPUTE_PARAMS, '') do |entry|
+      log_index += 1
+      if log_index == 1
+        assert entry.key?('textPayload'),
+               "Entry ##{log_index} #{entry} did not have textPayload"
+      elsif log_index < 4
```
crassirostris: Same comment as above. I think it's better to split the test configuration and the logic, to make adding tests in future easier

test/plugin/base_test.rb
```diff
@@ -305,20 +305,83 @@ def test_structured_payload_malformatted_log
...
+        assert_equal 'test log entry 0', get_string(fields['msg']), entry
+        assert_equal 'test', get_string(fields['tag2']), entry
+        assert_equal 5000, get_number(fields['data']), entry
+        assert_equal null_value, fields['some_null_field'], entry
+      end
     end
   end
 
-  def test_structured_payload_json_container_log
+  def test_structured_payload_json_log_container
```
crassirostris: Do we really need this test, since the logic is generic now?

test/plugin/base_test.rb
```diff
@@ -331,13 +394,34 @@ def test_structured_payload_json_container_log
...
-    verify_log_entries(
-      3, CONTAINER_FROM_METADATA_PARAMS, '') do |entry|
+    verify_log_entries(3, CONTAINER_FROM_METADATA_PARAMS, '') do |entry|
+      log_index += 1
+      assert entry.key?('textPayload'),
+             "Entry ##{log_index} #{entry} did not have textPayload"
+    end
+  end
+
+  def test_structured_payload_json_log_detect_json_container
```
crassirostris: Same comment as above

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -600,19 +605,20 @@ def write(chunk)
...
             group_common_labels.merge(extracted_common_labels)
 
-          if entry_resource.type == CONTAINER_CONSTANTS[:resource_type]
+          is_json = false
+          if @detect_json
             # Save the timestamp if available, then clear it out to allow for
             # determining whether we should parse the log or message field.
-            timestamp = record.key?('time') ? record['time'] : nil
-            record.delete('time')
+            timestamp = record.delete('time')
```
dhrupadb: Just confirming - this a more concise way of doing the 2 lines you replaced?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -600,19 +605,20 @@ def write(chunk)
...
             # unless there is additional metadata that would be lost.
-            is_json = false
-            if record.length == 1 && record.key?('log')
-              record_json = parse_json_or_nil(record['log'])
-            end
-            if record.length == 1 && record.key?('message')
-              record_json = parse_json_or_nil(record['message'])
+            record_json = nil
+            if record.length == 1
+              %w(log message msg).each do |field|
```
dhrupadb: what's the difference between `msg` and `message`. Trying to understand why we need both.

igorpeshansky: They are just two of the three different field names that may contain the log text. Because different systems use different names by default, we should be checking all three.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Exactly: [Hash#delete](https://docs.ruby-lang.org/en/2.0.0/Hash.html#method-i-delete).

test/plugin/base_test.rb
igorpeshansky: @crassirostris Can you please clarify? We expect the whole JSON to end up in `textPayload`.

test/plugin/base_test.rb
igorpeshansky: Good idea. It was getting pretty unwieldy. Done.

test/plugin/base_test.rb
igorpeshansky: Done.

test/plugin/base_test.rb
igorpeshansky: I wanted to make absolutely sure that we don't break the container use case. I'll defer to your opinion -- if you feel that these tests are redundant, I'll remove them.

test/plugin/base_test.rb
igorpeshansky: Same answer as above.

test/plugin/base_test.rb
```diff
@@ -302,50 +302,150 @@ def test_structured_payload_malformatted_log
...
+    end
+  end
+
+  def test_structured_payload_json_log_default_not_parsed_json
+    setup_gce_metadata_stubs
+    json_string = '{"msg": "test log entry 0", "tag2": "test", ' \
+                  '"data": 5000, "some_null_field": null}'
+    setup_logging_stubs do
+      d = create_driver(APPLICATION_DEFAULT_CONFIG)
+      %w(log msg).each do |field|
```
qingling128: `%w(message log msg)`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -117,6 +117,11 @@ module Constants
     # Whether to also keep the trace key/value in the payload.
     config_param :keep_trace_key, :bool, :default => false
 
+    # Whether to try to detect if the record is a text log entry with JSON
+    # content that needs to be parsed.
+    config_param :detect_json, :bool, :default => false
```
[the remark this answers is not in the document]
igorpeshansky: Correct. I've already discussed this with @crassirostris -- the Kubernetes fluentd config will be updated simultaneously with picking up this change.

test/plugin/base_test.rb
igorpeshansky: No, `message` is special, because it ends up in `textPayload`, so it has a separate test above (`test_structured_payload_json_log_default_not_parsed_text`).

igorpeshansky:
Conflicts resolved. PTAL.

test/plugin/base_test.rb
qingling128: Cool cool. Thanks for clarifiying.

igorpeshansky:
Thanks, @crassirostris. I propose we discuss the right structure for the tests and refactor them in a follow-up PR.

PR #148, opened 2017-08-21
Docker and GKE container support.

qingling128:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -38,12 +38,14 @@ class Map
 module Fluent
   # fluentd output plugin for the Stackdriver Logging API
   class GoogleCloudOutput < BufferedOutput
-    # Constants for service names and resource types.
+    # Constants.
     module Constants
+      # Service names and resource types.
```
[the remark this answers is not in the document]
qingling128: Makes sense. Changed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -11,9 +11,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+require 'excon'
 require 'grpc'
 require 'json'
 require 'open-uri'
+require 'rubygems'
```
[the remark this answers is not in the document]
qingling128: Ah I see. Good to know.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -58,7 +60,11 @@ module Constants
         resource_type: 'container',
         extra_resource_labels: %w(namespace_id pod_id container_name),
         extra_common_labels: %w(namespace_name pod_name),
-        metadata_attributes: %w(kube-env).to_set
+        metadata_attributes: %w(kube-env)
+      }
+      DOCKER_CONSTANTS = {
+        service: 'dockercontainer.googleapis.com',
```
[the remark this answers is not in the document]
qingling128: There is no existing service with this name. Logging API V2 no longer replies `XXX.googleapis.com` type of services. So the purpose of this is just to provide a unique namespace for the labels. I've just changed it to `docker.googleapis.com`. BTW I had sent the backend folks an email regarding this naming a while ago. Will resume conversation on that thread as well.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -154,6 +160,26 @@ module Constants
...
       }
+
+      # Used as a map key for entries with no locally unique id in the payload.
+      NO_UNIQUE_ID_PRESENT = 'no_locally_unique_id'
+
+      # Metadata Agent support.
+
+      # Use empty string as request path when locally-unique key of monitored
+      # resource can be implicitly inferred by Metadata Agent.
+      IMPLICIT_MONITORED_RESOURCE_UNIQUE_KEY = ''
```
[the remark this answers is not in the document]
qingling128: Renamed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -154,6 +160,26 @@ module Constants
...
+      DEFAULT_METADATA_AGENT_URL =
+        'http://local-metadata-agent.stackdriver.com:8000'
+
+      # The label name of locally unique id in the json payload. When a record
+      # has this field in the payload, we will use the value to retrieve
+      # monitored resource from Stackdriver Metadata agent.
+      DEFAULT_UNIQUE_ID_KEY = "#{DEFAULT_PAYLOAD_KEY_PREFIX}/locally_unique_id"
+
+      # Docker container support.
+      DEFAULT_DOCKER_API_SOCKET_PATH = '/var/run/docker.sock'
```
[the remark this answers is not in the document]
qingling128: Oops. Removed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -401,7 +440,7 @@ def configure(conf)
 
       # Log an informational message containing the Logs viewer URL
       @log.info 'Logs viewer address: https://console.cloud.google.com/logs/',
-                "viewer?project=#{@project_id}&resource=#{@resource_type}/",
+                "viewer?project=#{@project_id}&resource=#{@resource.type}/",
```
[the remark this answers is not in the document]
qingling128: I changed it to:
```
      @log.info 'Logs viewer address: https://console.cloud.google.com/logs/',
                "viewer?project=#{@project_id}"
```

This directs users to the project main logging page. They will need to select the resource type and filters themselves. Is this what we want?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -417,245 +456,243 @@ def shutdown
...
+            determine_group_level_monitored_resource_and_labels(
+              locally_unique_id, tag_level_resource, tag_level_common_labels)
+
+          records.each do |time, record|
+            next unless record.is_a?(Hash)
+            entry_level_resource, entry_level_common_labels =
+              determine_entry_level_monitored_resource_and_labels(
+                group_level_resource, group_level_common_labels, record)
+
+            if [CONTAINER_CONSTANTS[:resource_type],
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -417,245 +456,243 @@ def shutdown
...
+                DOCKER_CONSTANTS[:resource_type]].include?(
+                  entry_level_resource.type)
+              # Save the timestamp if available, then clear it out to allow for
```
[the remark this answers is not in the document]
qingling128: Sounds good.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1112,14 +1218,57 @@ def determine_entry_level_labels(group_resource, record)
...
+      # TODO(qingling128): Use Google::Api::MonitoredResource directly after we
+      # upgrade gRPC version to include the fix for the protobuf map
+      # corruption issue.
+      Google::Apis::LoggingV2beta1::MonitoredResource.new(
+        type: resource.type,
+        labels: resource.labels.to_h
+      )
+    end
+
+    # Call Metadata Agent and parse response to json. Return nil in case of any
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1112,14 +1218,57 @@ def determine_entry_level_labels(group_resource, record)
...
+    # error / failure.
+    def call_metadata_agent(path)
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1112,14 +1218,57 @@ def determine_entry_level_labels(group_resource, record)
...
+          @log.error 'Response from Metadata Agent is not in valid json ' \
+                     "format: '#{response.inspect}'."
+          return nil
+        end
+        @log.debug "Response from Metadata Agent: #{parsed_hash}"
+        return parsed_hash
+      end
+    rescue StandardError => e
+      @log.error 'Error calling Metadata Agent.', error: e
+      return nil
```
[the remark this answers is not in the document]
qingling128: Removed `return nil`.

test/plugin/base_test.rb
```diff
@@ -1160,11 +1334,20 @@ def setup_prometheus
     Prometheus::Client.registry.instance_variable_set('@metrics', {})
   end
 
+  def setup_metadata_agent_stubs(resource_type)
```
igorpeshansky: As I understand the way webmock works, we can stub out all the possible URLs at once (in fact, we could even have a static unique_id->monitored_resource map, and use [dynamic response generation](https://github.com/bblimke/webmock#responses-dynamically-evaluated-from-lambda)). We can then set expectations for specific URLs being requested in individual tests.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -154,6 +160,26 @@ module Constants
...
+      # resource can be implicitly inferred by Metadata Agent.
+      IMPLICIT_MONITORED_RESOURCE_UNIQUE_KEY = ''
+
+      DEFAULT_METADATA_AGENT_URL =
+        'http://local-metadata-agent.stackdriver.com:8000'
+
+      # The label name of locally unique id in the json payload. When a record
+      # has this field in the payload, we will use the value to retrieve
+      # monitored resource from Stackdriver Metadata agent.
+      DEFAULT_UNIQUE_ID_KEY = "#{DEFAULT_PAYLOAD_KEY_PREFIX}/locally_unique_id"
```
igorpeshansky: This would probably problematic to change at this point, but I just realized that this string does not mention the word "resource" anywhere. It would make more sense to call this "#{DEFAULT_PAYLOAD_KEY_PREFIX}/local_resource_id" instead. WDYT?

igorpeshansky: Why do we call this `DEFAULT_UNIQUE_ID_KEY`? Can this key be overridden? If not, let's call it something like `LOCAL_RESOURCE_ID_KEY` or something...

test/plugin/base_test.rb
```diff
@@ -1045,6 +1045,180 @@ def test_log_entry_trace_field
...
+
+  # Test the enable_metadata_agent set to true.
+  def test_configure_enable_metadata_agent_true
+    setup_gce_metadata_stubs
+    setup_metadata_agent_stubs(IMPLICIT_MONITORED_RESOURCE_UNIQUE_KEY)
+    d = create_driver(ENABLE_METADATA_AGENT_CONFIG)
+    assert_true d.instance.instance_variable_get(:@enable_metadata_agent)
+  end
+
+  # Test an implicit monitored resource can be retrieved from Metadata Agent
```
igorpeshansky: Test *that* an implicit ...

test/plugin/base_test.rb
qingling128: Refactored. PTAL.

test/plugin/base_test.rb

test/plugin/base_test.rb
```diff
@@ -1045,6 +1045,180 @@ def test_log_entry_trace_field
...
+
+  # Test the enable_metadata_agent config not set or set to false.
+  def test_configure_enable_metadata_agent_default_and_false
+    setup_gce_metadata_stubs
+    [create_driver, create_driver(DISABLE_METADATA_AGENT_CONFIG)].each do |d|
+      assert_false d.instance.instance_variable_get(:@enable_metadata_agent)
+    end
+  end
+
+  # Test the enable_metadata_agent set to true.
```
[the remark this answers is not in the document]

test/plugin/base_test.rb
```diff
@@ -1045,6 +1045,180 @@ def test_log_entry_trace_field
     end
   end
 
+  # Metadata Agent related tests.
+
+  # Test the enable_metadata_agent config not set or set to false.
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1227,9 +1376,14 @@ def compute_severity(resource_type, record, entry_common_labels)
...
         return parse_severity(record.delete('severity'))
-      elsif resource_type == CONTAINER_CONSTANTS[:resource_type] &&
-            entry_common_labels.key?("#{CONTAINER_CONSTANTS[:service]}/stream")
-        stream = entry_common_labels["#{CONTAINER_CONSTANTS[:service]}/stream"]
+      elsif [CONTAINER_CONSTANTS[:resource_type],
+             DOCKER_CONSTANTS[:resource_type]].include?(resource_type)
+        stream = entry_level_common_labels[
+          "#{CONTAINER_CONSTANTS[:service]}/stream"] if
+           resource_type == CONTAINER_CONSTANTS[:resource_type]
+        stream = entry_level_common_labels[
```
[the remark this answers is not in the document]
qingling128: Right. I've removed this logic from the code base. For now the configuration for docker container stderr/stdout logs looks like this:
```
# For Docker container stderr / stdout logs.
<filter container.**>
  @type record_transformer
  enable_ruby
  <record>
    "logging.googleapis.com/locally_unique_id" ${tag}
    "severity" ${record['source']}
  </record>
  remove_keys container_id,container_name
</filter>
```

While the command to run docker looks like:
```
sudo docker run -it --log-driver=fluentd --log-opt fluentd-address=tcp://127.0.0.1/24224 --log-opt tag="container.{{.FullID}}"
```

Note that to support the additional `severity` field for docker container, we will need to save then restore `severity` field just like what we did to `time` when parsing json. (This was not an issue for GKE.)

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1100,6 +1197,15 @@ def determine_entry_level_labels(group_resource, record)
...
         end
+
+      # Docker containers.
+      when DOCKER_CONSTANTS[:resource_type]
+        # For stderr / stdout logs coming from Docker Fluentd Logging Driver,
+        # the log record has two fields: 'source' and 'log'. Extract and remove
+        # 'source' from the json record and set it as a common label.
+        common_labels.merge!(
+          delete_and_extract_labels(
+            record, 'source' => "#{DOCKER_CONSTANTS[:service]}/stream"))
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,81 +993,90 @@ def determine_agent_level_common_labels
...
+    # "logging.googleapis.com/locally_unique_id". This way we can avoid
+    # redundant calls to Metadata Agent to retrieve the monitored resource
+    # corresponding to the locally_unique_id.
+    def group_log_entries_by_unique_id(entries)
+      groups = {}
+      # Log entries with no locally unique id go into a group with key
+      # NO_UNIQUE_ID_PRESENT.
+      groups[NO_UNIQUE_ID_PRESENT] = []
+      entries.each do |time, record|
+        if record.is_a?(Hash) && record.key?(DEFAULT_UNIQUE_ID_KEY)
```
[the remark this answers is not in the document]
qingling128: It *should* always be a `Hash`. User could mess with it though. For example they can send non-hash values via `in_forward` plugin. We used to have this check. So I kept it. We should add some logs when it's not a hash indicating we are dropping invalid log entries though.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,81 +993,90 @@ def determine_agent_level_common_labels
...
+    # Once log entries are grouped by tags, we want to further group them by
+    # locally_unique_id, which is specified in the payload under the key
+    # "logging.googleapis.com/locally_unique_id". This way we can avoid
+    # redundant calls to Metadata Agent to retrieve the monitored resource
+    # corresponding to the locally_unique_id.
+    def group_log_entries_by_unique_id(entries)
+      groups = {}
+      # Log entries with no locally unique id go into a group with key
+      # NO_UNIQUE_ID_PRESENT.
+      groups[NO_UNIQUE_ID_PRESENT] = []
```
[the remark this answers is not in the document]
qingling128: This no longer exists after commit e0cdf75caeece1a9e80ae4274280d6d73f629cda. Instead we are hashing by the tag & local_resource_id pairs.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -417,245 +456,243 @@ def shutdown
...
-            record.delete('time')
-            # If the log is json, we want to export it as a structured log
-            # unless there is additional metadata that would be lost.
-            is_json = false
-            if record.length == 1 && record.key?('log')
-              record_json = parse_json_or_nil(record['log'])
-            end
-            if record.length == 1 && record.key?('message')
-              record_json = parse_json_or_nil(record['message'])
+      entries_grouped_by_tag = group_log_entries_by_tag(chunk)
```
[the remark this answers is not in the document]
qingling128: This has been changed in commit e0cdf75caeece1a9e80ae4274280d6d73f629cda.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Nope, this can not be overridden. I just changed to `LOCAL_RESOURCE_ID_KEY`.

qingling128: I think `local_resource_id` makes more sense. And it's not too late to change it at this point. So I just changed it along with all the comments. I'll change the configs in the EAP doc once I push a new gem.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -154,6 +160,26 @@ module Constants
           'Google::Apis::LoggingV2beta1::LogEntryOperation'
         ]
       }
+
+      # Used as a map key for entries with no locally unique id in the payload.
+      NO_UNIQUE_ID_PRESENT = 'no_locally_unique_id'
```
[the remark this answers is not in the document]
qingling128: Yup, this has been removed as in commit e0cdf75caeece1a9e80ae4274280d6d73f629cda.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -351,22 +382,30 @@ def configure(conf)
...
-      # no longer need to require all these metadata in logging agent. But for
-      # now, they are still required.
+      # Set agent-level monitored resource. This monitored resource is initiated
+      # as the logging agent starts up. It will be inherited by all log entries
+      # processed by this agent. First try to retrieve it via Metadata Agent.
+      if @enable_metadata_agent
+        # The locally-unique key for this should be the instance id. Since this
+        # can be implicitly inferred by Metadata Agent, we do not need to
+        # explicitly send the key.
+        @resource = call_metadata_agent_for_monitored_resource(
```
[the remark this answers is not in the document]
qingling128: We've been doing this for a while. (See details in `determine_agent_level_monitored_resource_via_legacy`). I've created a separate ticket for potential refactoring: b/65175256.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1485,7 +1638,8 @@ def set_payload(resource_type, record, entry, is_json)
         entry.text_payload = record['log']
       elsif is_json
         entry.json_payload = record
-      elsif resource_type == CONTAINER_CONSTANTS[:resource_type] &&
+      elsif [CONTAINER_CONSTANTS[:resource_type],
+             DOCKER_CONSTANTS[:resource_type]].include?(resource_type) &&
```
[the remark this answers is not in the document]
qingling128: Yeah, that sounds good to me.

qingling128:
All feedback addressed. PTAL. (I noticed there are some conflicts with master. I'll hold for rebasing at this moment. Will wait for https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/146 to be merged first).

test/plugin/base_test.rb
```diff
@@ -1045,6 +1045,188 @@ def test_log_entry_trace_field
...
+        d.emit(log_entry)
+        d.run
+      end
+      verify_log_entries(1, expected_params) do |entry|
+        assert_equal_with_default entry['severity'], severity, 'DEFAULT', entry
+      end
+    end
+  end
+
+  # Test logs from applications running in GKE. These logs has the label
```
igorpeshansky: Nit: `These logs have the label`... But I wonder why you chose to call it out for GKE, but not for Docker...

test/plugin/base_test.rb
```diff
@@ -1160,11 +1342,27 @@ def setup_prometheus
     Prometheus::Client.registry.instance_variable_set('@metrics', {})
   end
 
+  def url_with_local_resource_id(local_resource_id)
```
igorpeshansky: `metadata_request_url`?

test/plugin/base_test.rb
```diff
@@ -1160,11 +1342,27 @@ def setup_prometheus
...
+  def assert_requested_metadata_agent_stub(local_resource_id)
+    assert_requested :get, url_with_local_resource_id(local_resource_id)
+  end
+
   def container_tag_with_container_name(container_name)
     "kubernetes.#{CONTAINER_POD_NAME}_#{CONTAINER_NAMESPACE_NAME}_" \
       "#{container_name}"
   end
 
+  # GKE Container
```
igorpeshansky: (a) Should this be above `container_tag_with_container_name`?
(b) Missing trailing '.'.

test/plugin/base_test.rb
igorpeshansky: Do we want to put a `# Metadata Agent.` comment above these functions?

test/plugin/base_test.rb
```diff
@@ -1192,6 +1390,34 @@ def container_log_entry(log, stream = CONTAINER_STREAM)
...
+  def gke_application_container_log_entry(log)
+    {
+      log: log,
+      LOCAL_RESOURCE_ID_KEY =>
+        "gke_containerName.#{CONTAINER_NAMESPACE_ID}" \
+        ".#{CONTAINER_POD_NAME}.#{CONTAINER_CONTAINER_NAME}"
+    }
+  end
+
+  # Docker Container
```
igorpeshansky: Nit: Missing trailing '.'.

test/plugin/base_test.rb
```diff
@@ -1192,6 +1390,34 @@ def container_log_entry(log, stream = CONTAINER_STREAM)
...
+  def docker_container_stdout_stderr_log_entry(
```
igorpeshansky: [Optional] `docker_container_stream_log_entry`...

test/plugin/constants.rb
```diff
@@ -23,8 +24,10 @@ module Constants
   # Attributes used for the GCE metadata service.
   PROJECT_ID = 'test-project-id'
   ZONE = 'us-central1-b'
+  ZONE2 = 'us-central1-c'
```
igorpeshansky: Let's call these `METADATA_ZONE` and `METADATA_VM_ID`, to make it consistent with `CUSTOM_ZONE` and `CUSTOM_VM_ID` below...

test/plugin/constants.rb
```diff
@@ -477,4 +562,50 @@ module Constants
     '' => '_'
   }
   ALL_TAGS = VALID_TAGS.merge(INVALID_TAGS)
+
+  # Stub value for Monitored resources from Metadata Agent.
+  # Map from the local_resource_id to the retrieved monitored resource.
+  MONITORED_RESOURCE_STUBS = {
+    # Implicit GCE instance.
+    IMPLICIT_LOCAL_RESOURCE_ID =>
+      '{' \
```
igorpeshansky: How about:
```ruby
  MONITORED_RESOURCE_STUBS = {
    # Implicit GCE instance.
    IMPLICIT_LOCAL_RESOURCE_ID =>
      {
        'type' => COMPUTE_CONSTANTS[:resource_type],
        'labels' => {
          'zone': ZONE2,
          'instance_id': VM_ID2
        }
      }.to_json,
    # Docker container stderr / stdout logs.
    "container.#{DOCKER_CONTAINER_ID}" =>
      {
        ...
      }.to_json,
      ...
  }
```
?

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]
qingling128: I created b/65257345 for this follow up.

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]
igorpeshansky: As I said elsewhere, can't we just set the severity explicitly, e.g.:
```
    "severity" ${if record['source']=='STDOUT' then 'INFO' else 'ERROR' end}
```
?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -105,6 +112,23 @@ module Constants
         "#{DEFAULT_PAYLOAD_KEY_PREFIX}/sourceLocation"
       DEFAULT_TRACE_KEY = "#{DEFAULT_PAYLOAD_KEY_PREFIX}/trace"
 
+      # Metadata Agent support.
+
+      # Use empty string as request path when locally-unique id of monitored
+      # resource can be implicitly inferred by Metadata Agent.
+      IMPLICIT_MONITORED_RESOURCE_UNIQUE_ID = ''
```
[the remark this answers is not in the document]
qingling128: Moved.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -105,6 +112,23 @@ module Constants
...
+        'http://local-metadata-agent.stackdriver.com:8000'
+
+      # The label name of locally unique id in the json payload. When a record
+      # has this field in the payload, we will use the value to retrieve
+      # monitored resource from Stackdriver Metadata agent.
+      DEFAULT_UNIQUE_ID_KEY = "#{DEFAULT_PAYLOAD_KEY_PREFIX}/locally_unique_id"
+    end
+
+    # Constants for log entry field extraction.
+    module LogEntryFieldsConstants
```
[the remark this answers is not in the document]
qingling128: Renamed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -463,16 +490,21 @@ def write(chunk)
...
               is_json = true
             end
-            # Restore timestamp if necessary.
+            # Restore timestamp and severity if necessary. Note that the nested
+            # json might also has 'time' and 'severity' fields. If that is the
+            # case, we do not want to override the value.
             unless record.key?('time') || timestamp.nil?
               record['time'] = timestamp
             end
+            unless record.key?('severity') || severity.nil?
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,138 +986,166 @@ def determine_agent_level_common_labels
...
+      groups = {}
+      chunk.msgpack_each do |tag, time, record|
+        sanitized_tag = sanitize_tag(tag)
+        if sanitized_tag.nil?
+          @log.warn "Dropping log entries with invalid tag: '#{tag}'. " \
+                    'A tag should be a string with utf8 characters.'
+          next
+        end
+        locally_unique_id = record.delete(DEFAULT_UNIQUE_ID_KEY) if
+          record.is_a?(Hash)
```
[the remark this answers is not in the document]
qingling128: For logs from `in_forward` plugin, the record can be anything. We even have a test `test_malformed_log` to test this case. I added some logging for it.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,138 +986,166 @@ def determine_agent_level_common_labels
...
+        hash_key = [sanitized_tag, locally_unique_id].freeze
```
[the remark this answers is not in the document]
qingling128: Ah, I removed it in order to leave it as `nil` so that later on we can evaluate `if local_resource_id` as false. If I convert it to an empty string, it becomes confusing as we also have the implicit local resource id as an empty string.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,138 +986,166 @@ def determine_agent_level_common_labels
...
-    # there is no match.
-    def determine_group_level_monitored_resource_type(tag)
+    # Determine the group level monitored resource and common labels shared by a
+    # collection of entries.
+    def determine_group_level_monitored_resource_and_labels(tag,
+                                                            locally_unique_id)
+      resource = @resource.dup
+      resource.labels = @resource.labels.dup
+      common_labels = @common_labels.dup
+      matched_regexp_group = nil
```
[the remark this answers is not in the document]
qingling128: Yes. if it's not `GKE`, the `@tag_regexp_list` will be empty. I added this line because some tests were failing.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,138 +986,166 @@ def determine_agent_level_common_labels
...
-      [@resource.type, nil]
-    end
 
-    # Determine group level monitored resource labels and common labels. These
-    # labels will be shared by a collection of entries.
-    def determine_group_level_labels(group_resource_type, matched_regex_group)
-      group_resource_labels = @resource.labels.dup
-      group_common_labels = @common_labels.dup
+      # Determine the monitored resource based on the locally unique id.
+      # Different monitored resource type has unique ids in different format. We
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,138 +986,166 @@ def determine_agent_level_common_labels
...
+      # will query Metadata Agent for the monitored resource. Return the
+      # original monitored resource (either the instance resource or the
+      # resource inferred by tag) if failing to get a monitored resource from
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,138 +986,166 @@ def determine_agent_level_common_labels
...
+      #   "gke_containerName.<namespace_id>.<pod_name>.<container_name>"
+      if @enable_metadata_agent && locally_unique_id
+        @log.debug 'Calling metadata agent with locally unique id: ' \
+                  "#{locally_unique_id}."
+        retrieved_resource = call_metadata_agent_for_monitored_resource(
+          locally_unique_id)
+        @log.debug 'Retrieved monitored resource from metadata agent: ' \
+                  "#{retrieved_resource.inspect}."
+        unless retrieved_resource.nil?
+          # Temporarily renaming 'gke_container' to 'container'.
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,138 +986,166 @@ def determine_agent_level_common_labels
...
             delete_and_extract_labels(
               common_labels_candidates,
-              CONTAINER_CONSTANTS[:extra_common_labels]
-                .map { |l| [l, "#{CONTAINER_CONSTANTS[:service]}/#{l}"] }.to_h))
+              GKE_CONSTANTS[:extra_common_labels]
+                .map { |l| [l, "#{GKE_CONSTANTS[:service]}/#{l}"] }.to_h))
         end
+
+      # Docker container.
+      when DOCKER_CONSTANTS[:resource_type]
```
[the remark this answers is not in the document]
qingling128: Per discussion with Summit, we should not be sending this metadata label for docker container since it's been ingested in the Logging Backend. For other existing resource types we have been doing it for a while and could not easily deprecate it.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1324,6 +1430,9 @@ def set_labels(record, entry)
       'E' => 'ERROR',
       'C' => 'CRITICAL',
       'A' => 'ALERT',
+      # Container stderr / stdout logs.
+      'STDOUT' => 'INFO',
```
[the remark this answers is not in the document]
qingling128: I've removed this map. We will address the `severity` value in the config.

test/plugin/base_test.rb
```diff
@@ -1045,6 +1045,188 @@ def test_log_entry_trace_field
...
+        d = create_driver(DOCKER_CONTAINER_CONFIG)
+        n.times do |i|
+          d.emit(docker_container_stdout_stderr_log_entry(log_entry(i)))
+        end
+        d.run
+      end
+      verify_log_entries(n, DOCKER_CONTAINER_PARAMS)
+      assert_requested_metadata_agent_stub(IMPLICIT_LOCAL_RESOURCE_ID)
+      assert_requested_metadata_agent_stub("container.#{DOCKER_CONTAINER_ID}")
+      WebMock.reset!
```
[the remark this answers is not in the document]
qingling128: I changed `setup_metadata_agent_stubs` to take a block. The naming or the method is left as it is, similar to `setup_logging_stubs`.

test/plugin/base_test.rb
```diff
@@ -1045,6 +1045,188 @@ def test_log_entry_trace_field
...
+    setup_logging_stubs do
+      d = create_driver(ENABLE_METADATA_AGENT_CONFIG)
+      d.emit(docker_container_application_log_entry(log_entry(0)))
+      d.run
+    end
+    verify_log_entries(1, DOCKER_CONTAINER_PARAMS_WITH_NO_STREAM) do |entry|
+      # Timestamp in 'time' field from log entry should be set properly.
+      assert_equal DOCKER_CONTAINER_SECONDS_EPOCH,
+                   entry['timestamp']['seconds'], entry
+      assert_equal DOCKER_CONTAINER_NANOS, entry['timestamp']['nanos'],
```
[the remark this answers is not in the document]
qingling128: Done.

test/plugin/base_test.rb
```diff
@@ -1045,6 +1045,188 @@ def test_log_entry_trace_field
...
+    end
+    assert_requested_metadata_agent_stub(IMPLICIT_LOCAL_RESOURCE_ID)
+    assert_requested_metadata_agent_stub(
+      "containerName.#{DOCKER_CONTAINER_NAME}")
+  end
+
+  # Test that the 'source' field is properly extracted from the record json and
+  # set as a common label 'stream'. Also entry.severity is set accordingly for
+  # Docker container logs.
+  def test_source_and_severity_for_docker_container_logs
```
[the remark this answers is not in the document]
qingling128: Right. I've removed the `severity` check.

test/plugin/base_test.rb
qingling128: Changed `has` to `have` and called it out for Docker container as well.

test/plugin/base_test.rb

test/plugin/base_test.rb
qingling128: Moved this above and added `.`.

test/plugin/base_test.rb

test/plugin/base_test.rb
qingling128: Hmm... I think `stderr` and `stdout` is more straight forward.

test/plugin/constants.rb
qingling128: Changed.

test/plugin/constants.rb
qingling128: Good idea. Changed.

qingling128:
PTAL.

The requested changes that requires a `git rebase` from `master` will be postponed till everything else seems fine. Because I don't want to confuse GitHub and mess up the comment history (learnt that the hard way).

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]
qingling128: Right. I've removed the logic to map `severity` for `stderr` and `stdout`.

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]
qingling128: There is a test `test_malformed_log` that makes sure we throw away non-hash logs instead of letting the agent crash. For logs coming from `in_forward` plugin, the `record` can be anything.

qingling128:
I replied to the last two comments. But for some reason it does not show up in the `conversation` tab... So let me answer them here.

[The `record.is_a?(Hash)` check]
There is a test `test_malformed_log` that makes sure we throw away `non-hash` logs instead of letting the agent crash. For logs coming from `in_forward` plugin, the record can be anything. Previously we are checking it in two places. I just modified it to check it once.

[`severity` mapping logic]
I've removed the logic to map `severity` for `stderr` and `stdout`. We will rely on configurations instead.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,138 +983,172 @@ def determine_agent_level_common_labels
...
+          next
+        end
+        sanitized_tag = sanitize_tag(tag)
+        if sanitized_tag.nil?
+          @log.warn "Dropping log entries with invalid tag: '#{tag}'. " \
+                    'A tag should be a string with utf8 characters.'
+          next
+        end
+        local_resource_id = record.delete(LOCAL_RESOURCE_ID_KEY)
+        # If local_resource_id is nil, use that as the hash key anyway.
```
igorpeshansky: I'd put it another way: `# A nil local_resource_id means "fall back to legacy".`

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,138 +983,172 @@ def determine_agent_level_common_labels
...
-      group_common_labels.freeze
-
-      [group_resource, group_common_labels]
+    # Group the log entries by tag and local_resource_id pairs.
+    def group_log_entries_by_tag_and_local_resource_id(chunk)
+      groups = {}
+      chunk.msgpack_each do |tag, time, record|
+        unless record.is_a?(Hash)
+          @log.warn "Dropping log entries with malformed record: '#{record}'." \
+                    ' A log record should be in json format.'
```
igorpeshansky: Nit: JSON.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,138 +983,172 @@ def determine_agent_level_common_labels
...
-      group_resource.labels.freeze
-      group_common_labels.freeze
-
-      [group_resource, group_common_labels]
+    # Group the log entries by tag and local_resource_id pairs.
+    def group_log_entries_by_tag_and_local_resource_id(chunk)
+      groups = {}
+      chunk.msgpack_each do |tag, time, record|
+        unless record.is_a?(Hash)
+          @log.warn "Dropping log entries with malformed record: '#{record}'." \
```
igorpeshansky: [Optional] Do we want `#{record.inspect}`? Or is a stringified version of the record sufficient to figure out what it was?

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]
igorpeshansky: How would you feel about calling it out explicitly with a "hideously awesome" recipe I [found somewhere](https://stackoverflow.com/a/13299030/6752892):
```ruby
      if @tag_regexp_list.each do |derived_type, tag_regexp|
        matched_regexp_group = tag_regexp.match(tag)
        if matched_regexp_group
          resource.type = derived_type
          break
        end
      end.empty?
        matched_regexp_group = nil
      end
```
? Or with something less aw{esome,ful} and more conventional, e.g.:
```ruby
      unless @tag_regexp_list.empty?
        @tag_regexp_list.each do |derived_type, tag_regexp|
          matched_regexp_group = tag_regexp.match(tag)
          if matched_regexp_group
            resource.type = derived_type
            break
          end
        end
      else
        matched_regexp_group = nil
      end
```
?

test/plugin/constants.rb
```diff
@@ -215,7 +244,16 @@ module Constants
       "#{COMPUTE_CONSTANTS[:service]}/resource_name" => HOSTNAME
     }
   }
+  COMPUTE_PARAMS_WITH_DIFFERENT_VM_ID_AND_ZONE = COMPUTE_PARAMS.merge(
```
igorpeshansky: `COMPUTE_PARAMS_WITH_METADATA_VM_ID_AND_ZONE`

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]
igorpeshansky: > *Aside*: I feel we're introducing more legacy stuff to get rid of by propagating `DOCKER_CONSTANTS` throughout the code. Ideally, we shouldn't need to reference it at all -- the code should not care what the resource looks like.

That said, in this particular case we shouldn't need to remove the `resource_name` field because it would never be added (lines 975-976 should only be executed on raw GCE or on GKE). So this seems like dead code to me, unless I'm missing something.

test/plugin/constants.rb
```diff
@@ -477,4 +563,50 @@ module Constants
...
+    # Implicit GCE instance.
+    IMPLICIT_LOCAL_RESOURCE_ID =>
+      {
+        'type' => COMPUTE_CONSTANTS[:resource_type],
+        'labels' => {
+          'zone' => METADATA_ZONE,
+          'instance_id' => METADATA_VM_ID
+        }
+      }.to_json,
+    # Docker container stderr g stdout logs.
```
igorpeshansky: I think you meant to keep the `/` here instead of `g`.

test/plugin/base_test.rb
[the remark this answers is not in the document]
igorpeshansky: Careful, `WebMock` covers all stubs, not just the metadata agent stubs. In fact, ideally, this would be part of the `setup_logging_stubs` context, because it also changes `@logs_sent`, but, unfortunately, we want to stop mocking the logging API early, to ensure `d.emit` is only called within that block... I wish Ruby allowed passing multiple blocks (one to execute the code and one to verify the results), but, alas, this seems to be [woefully inefficient](https://mudge.name/2011/01/26/passing-blocks-in-ruby-without-block.html)...
So, my suggestion was to create a top-level context object to wrap all stubs. Does that sound reasonable?

test/plugin/base_test.rb
igorpeshansky: Fine, but can we at least be consistent about the ordering of `stdout_stderr` vs `stderr_stdout`?

test/plugin/constants.rb
```diff
@@ -23,8 +25,10 @@ module Constants
   # Attributes used for the GCE metadata service.
   PROJECT_ID = 'test-project-id'
   ZONE = 'us-central1-b'
+  METADATA_ZONE = 'us-central1-c'
```
igorpeshansky: I would pull these out into their own section, e.g.:
```ruby
  # Attributes used for the Metadata Agent resources.
  METADATA_ZONE = 'us-central1-c'
  METADATA_VM_ID = '0123456789'
```

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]
qingling128: The line 975-976 gets executed on agent startup. At that point we can't tell `a raw GCE instance` from `a GCE instance with docker installed and potentially want to send docker logs`. The we will be adding this `resource_name` label regardless.

`Docker container` logs only get identified when the monitored resource (specified by `logging.googleapis.com/local_resource_id`) returned by metadata agent has `resource_type` as `docker_container`. This process happens kind of late when logs start coming in.

In a word, we still need to remove `resource_name` for `docker_container` unless we change the behavior dramatically as proposed in b/65175256.

test/plugin/base_test.rb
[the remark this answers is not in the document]
qingling128: Yes, `WebMock.reset!` will reset all stubs. I had to switch the order of a few lines to make sure that did not break out tests. If we are doing a top-level context object, we will be setting up **all** stubs for **all** tests that need any stub, right? That would make our test a bit cleaner. I wonder if it is a bit overkill though.

On the other hand, if we decide to do that, we should probably add `assert_expected_stub` checks for each case when a stub is used.

WDYT?

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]
qingling128: The first one kind of scared me...

We would need to initiate `matched_regexp_group` in the beginning because of how ruby decides its variable scope. As shown below `if` works, but `for each loop` does not.

```
$ irb
2.1.10 :002 > if true
2.1.10 :003?>     foo = 'a'
2.1.10 :004?>   else
2.1.10 :005 >       foo = 'b'
2.1.10 :006?>   end
 => "a"
2.1.10 :007 > foo
 => "a"
2.1.10 :008 > {'a' => 'a'}.each do |k, v|
2.1.10 :009 >       val = 'val'
2.1.10 :010?>   end
 => {"a"=>"a"}
2.1.10 :011 > val
NameError: undefined local variable or method `val' for main:Object
        from (irb):11
        from /usr/local/google/home/lingshi/.rvm/rubies/ruby-2.1.10/bin/irb:11:in `<main>'
```

I changed the code to be
```
      matched_regexp_group = nil
      # @tag_regexp_list can be an empty list.
      @tag_regexp_list.each do |derived_type, tag_regexp|
        matched_regexp_group = tag_regexp.match(tag)
        if matched_regexp_group
          resource.type = derived_type
          break
        end
      end
```
if that helps.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Guess it does not hurt using `#{record.inspect}` (only better). Just changed it.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

test/plugin/constants.rb
qingling128: Changed.

test/plugin/constants.rb
qingling128: Changed.

test/plugin/constants.rb
qingling128: LOL. Oops. Changed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,138 +983,174 @@ def determine_agent_level_common_labels
...
-      group_common_labels.freeze
-
-      [group_resource, group_common_labels]
+    # Group the log entries by tag and local_resource_id pairs.
+    def group_log_entries_by_tag_and_local_resource_id(chunk)
+      groups = {}
+      chunk.msgpack_each do |tag, time, record|
+        unless record.is_a?(Hash)
+          @log.warn 'Dropping log entries with malformed record: ' \
+                    "#{record.inspect}'. " \
```
igorpeshansky: I think you've lost the opening "`'`".

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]

test/plugin/base_test.rb
qingling128: Makes sense. I changed all occurrences to `stdout_stderr`.

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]
igorpeshansky: Fine, makes sense. Can we add a TODO to remove this once the resource is retrieved at a proper time?

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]
igorpeshansky: Ah, I didn't think about the block scoping issue. What you have now works, unless you also want to move the comment to the end of the previous line, i.e.:
```ruby
      matched_regexp_group = nil  # @tag_regexp_list can be an empty list.
```

test/plugin/base_test.rb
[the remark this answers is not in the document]
igorpeshansky: Oh, sorry I wasn't clear. I was simply suggesting a stub context object that would initialize `@logs_sent` and reset `WebMock` at the end. We could still call `setup_*_stubs` in the block passed into the context to set up only the stubs we need. The `assert_expected_stub` calls you have now would also have to appear within the block (before the reset). See my earlier example... Does this make more sense?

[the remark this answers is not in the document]
qingling128: AH! I see what you meant now. Sorry I misunderstood. Changed in a82a36334b933299a81a058eaf91cddd32d0e2e7.

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
qingling128: Oops. Added.

qingling128:
I cherry picked the other commits. Seems it still shows as `there is conflicts`. I'll rebase it after the code review is done.

PTAL.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -955,138 +980,175 @@ def determine_agent_level_common_labels
...
+    # collection of entries.
+    def determine_group_level_monitored_resource_and_labels(tag,
+                                                            local_resource_id)
+      resource = @resource.dup
+      resource.labels = @resource.labels.dup
+      common_labels = @common_labels.dup
+
+      # Change the resource type and set matched_regexp_group if the tag matches
+      # certain regexp.
+      matched_regexp_group = nil # @tag_regexp_list can be an empty list.
```
igorpeshansky: Two spaces before the comment, please.

test/plugin/base_test.rb
```diff
@@ -1160,6 +1412,33 @@ def setup_prometheus
     Prometheus::Client.registry.instance_variable_set('@metrics', {})
   end
 
+  # Metadata Agent.
+
+  def metadata_request_url(local_resource_id)
+    "#{DEFAULT_METADATA_AGENT_URL}/monitoredResource/#{local_resource_id}"
+  end
+
+  # Provide a stub context that initialize @logs_sent, execute the block and
```
igorpeshansky: Nit: initializes, executes, resets...

lib/fluent/plugin/out_google_cloud.rb

I tried `matched_regexp_group = nil<two_spaces_here># @tag_regexp_list` and Rubocop complains. So I guess you meant `matched_regexp_group = nil #<two_spaces_here> @tag_regexp_list`?

test/plugin/base_test.rb

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: I actually did mean two spaces before the `#`, but didn't realize Rubocop would complain. So let's go back to one space on each side of the `#`.

qingling128: Hahahaha, I see. Just changed it back.

qingling128:
Rebased. PTAL.

BTW, I fixed a few minor comments while rebasing. Also reverted my change to the Log Viewer URL since there is a separate bug to addresss that. Here are changes between the latest and the code before rebasing:

```
git diff origin/lingshi-container-rebase-backup..origin/lingshi-container

diff --git a/lib/fluent/plugin/out_google_cloud.rb b/lib/fluent/plugin/out_google_cloud.rb
index 3b81bc9..795c22c 100644
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -393,12 +393,6 @@ module Fluent
       end

       # Set required variables: @project_id, @vm_id, @vm_name and @zone.
-      # If any info above is included in the response from Metadata Agent, make
-      # use of that. Otherwise make some additional requests to metadata server.
-      #
-      # Note: Once we support metadata injection on the Logging API side, we
-      # might no longer need to require all these metadata in logging agent. But
-      # for now, they are still required.
       set_required_metadata_variables

       # Retrieve monitored resource.
@@ -439,7 +433,8 @@ module Fluent

       # Log an informational message containing the Logs viewer URL
       @log.info 'Logs viewer address: https://console.cloud.google.com/logs/',
-                "viewer?project=#{@project_id}"
+                "viewer?project=#{@project_id}&resource=#{@resource.type}/",
+                "instance_id/#{@vm_id}"
     end

     def start

@@ -592,8 +585,7 @@ module Fluent
             end

           rescue GRPC::Cancelled => error
-            increment_failed_requests_count(
-              GRPC::Core::StatusCodes::CANCELLED)
+            increment_failed_requests_count(GRPC::Core::StatusCodes::CANCELLED)
             # RPC cancelled, so retry via re-raising the error.
             raise error

@@ -1125,7 +1117,7 @@ module Fluent
           @cloudfunctions_log_match =
             @compiled_cloudfunctions_log_regexp.match(record['log'])
           common_labels['execution_id'] =
-            @cloudfunctions_log_match['execution_id'] if \
+            @cloudfunctions_log_match['execution_id'] if
               @cloudfunctions_log_match &&
               @cloudfunctions_log_match['execution_id']
         end

diff --git a/test/plugin/base_test.rb b/test/plugin/base_test.rb
index 46fd6bd..5e570c7 100644
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -1272,10 +1272,10 @@ module BaseTest
     end
   end

-  # Test logs from applications running in GKE. These logs have the label
+  # Test GKE container logs. These logs have the label
   # "logging.googleapis.com/local_resource_id" set in the format of
   # "gke_containerName.<namespace_id>.<pod_name>.<container_name>".
-  def test_gke_container_application_logs
+  def test_gke_container_logs
     [1, 2, 3, 5, 11, 50].each do |n|
       new_stub_context do
         setup_gce_metadata_stubs
@@ -1284,7 +1284,7 @@ module BaseTest
         setup_logging_stubs do
           d = create_driver(ENABLE_METADATA_AGENT_CONFIG)
           n.times do |i|
-            d.emit(gke_application_container_log_entry(log_entry(i)))
+            d.emit(gke_container_log_entry(log_entry(i)))
           end
           d.run
         end
@@ -1471,7 +1471,7 @@ module BaseTest
     }
   end

-  def gke_application_container_log_entry(log)
+  def gke_container_log_entry(log)
     {
       log: log,
       LOCAL_RESOURCE_ID_KEY =>
diff --git a/test/plugin/constants.rb b/test/plugin/constants.rb
index f354b65..8297486 100644
--- a/test/plugin/constants.rb
+++ b/test/plugin/constants.rb
@@ -601,7 +601,7 @@ module Constants
           'container_id' => DOCKER_CONTAINER_ID
         }
       }.to_json,
-    # GKE container application logs.
+    # GKE container logs.
     "gke_containerName.#{CONTAINER_NAMESPACE_ID}.#{CONTAINER_POD_NAME}." \
     "#{CONTAINER_CONTAINER_NAME}" =>
       {
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -420,39 +449,27 @@ def shutdown
...
-        grouped_entries[sanitized_tag].push(arr)
-      end
+      grouped_entries = group_log_entries_by_tag_and_local_resource_id(chunk)
 
-      grouped_entries.each do |tag, arr|
+      grouped_entries.each do |(tag, local_resource_id), arr|
         entries = []
-        group_resource, group_common_labels =
-          determine_group_level_monitored_resource_and_labels(tag)
+        group_level_resource, group_level_common_labels =
```
igorpeshansky: [Optional] I'm fine with keeping this as-is, but this renaming (`group_resource`->`group_level_resource`, `entry_resource`->`entry_level_resource`, etc) seems somewhat gratuitous to me.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -954,138 +974,175 @@ def determine_agent_level_common_labels
...
-      )
-
-      # Freeze the per-request state. Any further changes must be made on a
-      # per-entry basis.
-      group_resource.freeze
-      group_resource.labels.freeze
-      group_common_labels.freeze
-
-      [group_resource, group_common_labels]
+    # Group the log entries by tag and local_resource_id pairs.
```
igorpeshansky: `# Also filter out invalid non-Hash entries.`

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -954,138 +974,175 @@ def determine_agent_level_common_labels
...
+        @log.debug 'Calling metadata agent with local_resource_id: ' \
+                  "#{local_resource_id}."
+        retrieved_resource = call_metadata_agent_for_monitored_resource(
+          local_resource_id)
+        @log.debug 'Retrieved monitored resource from metadata agent: ' \
+                  "#{retrieved_resource.inspect}."
+        unless retrieved_resource.nil?
+          # TODO(qingling128): Fix this temporary renaming from 'gke_container'
+          # to 'container'.
+          retrieved_resource.type = 'container' if
```
igorpeshansky: [Optional] If you move this line below the `resource` assignment, you could use `resource` instead of `retrieved_resource` and fit this on one line.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -954,138 +974,175 @@ def determine_agent_level_common_labels
...
+      # GKE container:
+      #   "gke_containerName.<namespace_id>.<pod_name>.<container_name>"
+      if @enable_metadata_agent && local_resource_id
+        @log.debug 'Calling metadata agent with local_resource_id: ' \
+                  "#{local_resource_id}."
+        retrieved_resource = call_metadata_agent_for_monitored_resource(
+          local_resource_id)
+        @log.debug 'Retrieved monitored resource from metadata agent: ' \
+                  "#{retrieved_resource.inspect}."
+        unless retrieved_resource.nil?
```
igorpeshansky: `if retrieved_resource`

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -354,28 +381,30 @@ def configure(conf)
...
-      # TODO(qingling128): After Metadata Agent support is added, try extracting
-      # these info from responses from Metadata Agent first.
+      # Set agent-level monitored resource. This monitored resource is initiated
+      # as the logging agent starts up. It will be inherited by all log entries
+      # processed by this agent. First try to retrieve it via Metadata Agent.
+      if @enable_metadata_agent
+        # The local_resource_id for this should be the instance id. Since this
+        # can be implicitly inferred by Metadata Agent, we do not need to
+        # explicitly send the key.
+        @resource = call_metadata_agent_for_monitored_resource(
```
igorpeshansky: Do we want add a TODO here to retrieve the resource on demand (similar to the one you have in line 1094)?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1111,14 +1168,56 @@ def determine_entry_level_labels(group_resource, record)
...
               .map { |l| ["#{service_constants[:service]}/#{l}", l] }.to_h))
       end
 
-      [resource_labels, common_labels]
+      [resource, common_labels]
+    end
+
+    # Call Metadata Agent to get monitored resource information and parse
+    # response to Google::Api::MonitoredResource.
+    def call_metadata_agent_for_monitored_resource(local_resource_id)
```
igorpeshansky: Same here: s/`call`/`query`/?

test/plugin/base_test.rb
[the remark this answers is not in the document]
igorpeshansky: Should we not rename the test as well?

test/plugin/base_test.rb
```diff
@@ -1106,6 +1106,197 @@ def test_log_entry_trace_field
...
+  def test_time_field_extraction_for_docker_container_logs
+    new_stub_context do
+      setup_gce_metadata_stubs
+      setup_metadata_agent_stubs
+      setup_logging_stubs do
+        d = create_driver(ENABLE_METADATA_AGENT_CONFIG)
+        d.emit(docker_container_application_log_entry(log_entry(0)))
+        d.run
+      end
+      verify_log_entries(1, DOCKER_CONTAINER_PARAMS_WITH_NO_STREAM) do |entry|
```
igorpeshansky: Careful: since you've merged #146, this now simply verifies that there's 1 entry with a 'textPayload' field, but no longer verifies the content of that field. See, e.g., `test_timestamps`, `test_one_container_log_from_tag_stderr`, and `verify_container_logs`.

test/plugin/base_test.rb
```diff
@@ -1106,6 +1106,197 @@ def test_log_entry_trace_field
...
+        # Timestamp in 'time' field from log entry should be set properly.
+        assert_equal DOCKER_CONTAINER_SECONDS_EPOCH,
```
igorpeshansky: Other similar tests use `assert_equal_with_default` -- does it make sense to be consistent?

test/plugin/base_test.rb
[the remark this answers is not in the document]
qingling128: I guess you meant `Should we rename the test as well?` Renamed it to `test_source_for_docker_container_logs` and also fixed the comments.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -83,26 +86,43 @@ module Constants
...
       # Default back to GCE if invalid value is detected.
       SUBSERVICE_MAP.default = COMPUTE_CONSTANTS[:resource_type]
 
       # The map between a resource type and expected subservice attributes.
       SUBSERVICE_METADATA_ATTRIBUTES = \
-        [APPENGINE_CONSTANTS, CONTAINER_CONSTANTS, DATAPROC_CONSTANTS]
-        .map { |consts| [consts[:resource_type], consts[:metadata_attributes]] }
-        .to_h
+        [APPENGINE_CONSTANTS, GKE_CONSTANTS, DATAPROC_CONSTANTS]
+        .map do |consts|
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Ah, originally I renamed them because I added `tag_resource` and `tag_labels`. It's kind of ambiguous what that means. So i named them in an more explicit way showing that these are just regular monitored resource and metadata labels except they are at different levels.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -467,16 +484,17 @@ def write(chunk)
               record = record_json
               is_json = true
             end
-            # Restore timestamp if necessary.
-            unless record.key?('time') || timestamp.nil?
-              record['time'] = timestamp
-            end
+            # Restore timestamp and severity if necessary. Note that the nested
+            # json might also has 'time' and 'severity' fields. If that is the
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

test/plugin/base_test.rb
qingling128: Ah, i see. Just added the corresponding `verify_default_log_entry_text` checks.

test/plugin/base_test.rb
qingling128: Neither `DOCKER_CONTAINER_SECONDS_EPOCH` nor `DOCKER_CONTAINER_NANOS` is `0`, so we should be fine asserting it this way.

qingling128:
PTAL.

test/plugin/base_test.rb
igorpeshansky: Oops, sorry, I should have been clearer. This check happens automatically when you don't pass a block, so you only needed to change it here.

test/plugin/base_test.rb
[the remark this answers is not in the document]
igorpeshansky: I meant "shouldn't we". :-) But thanks.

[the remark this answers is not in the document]
qingling128: You're welcome.

test/plugin/base_test.rb
qingling128: Ah, I see. I should have taken a look at the change in #146. Just changed it back.

qingling128:
Do we want to squash and merge everything in one commit? Or to keep refactoring, docker support and gke support in separate commits?

igorpeshansky:
Let's just squash and merge. The history will remain in this PR.

PR #151, opened 2017-08-28
Revert support for labels in log entry.

qingling128:
The extraction of `labels` field from the record is kind of a major feature, and should go through product first. Reverting this part of the changes that was originally added in https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/145.


arthurdarcet:
@igorpeshansky ok, I have opened #152 re-introducing this. Let me know how the product review goes and if you need any changes

PR #152, opened 2017-08-29
Support for labels in log entry

arthurdarcet:
Reverts GoogleCloudPlatform/fluent-plugin-google-cloud#151

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -98,6 +98,7 @@ module Fluent
       # Default values for JSON payload keys to set the "trace",
       # "sourceLocation", "operation" and "labels" fields in the LogEntry.
       DEFAULT_PAYLOAD_KEY_PREFIX = 'logging.googleapis.com'
+      DEFAULT_LABELS_KEY = "#{DEFAULT_PAYLOAD_KEY_PREFIX}/labels"
       DEFAULT_HTTP_REQUEST_KEY = 'httpRequest'
       DEFAULT_OPERATION_KEY = "#{DEFAULT_PAYLOAD_KEY_PREFIX}/operation"
       DEFAULT_SOURCE_LOCATION_KEY =
@@ -191,6 +192,7 @@ module Fluent
     config_param :vm_name, :string, :default => nil
 
     # Map keys from a JSON payload to corresponding LogEntry fields.
+    config_param :labels_key, :string, :default => DEFAULT_LABELS_KEY
     config_param :http_request_key, :string, :default =>
       DEFAULT_HTTP_REQUEST_KEY
     config_param :operation_key, :string, :default => DEFAULT_OPERATION_KEY
@@ -521,6 +523,7 @@ module Fluent
           entry.trace = fq_trace_id if fq_trace_id
 
           set_log_entry_fields(record, entry)
+          set_labels(record, entry)
 
           if @use_grpc
             set_payload_grpc(entry_resource.type, record, entry, is_json)
@@ -1280,6 +1283,20 @@ module Fluent
           @log.error "Failed to set log entry field for #{field_name}.", err
         end
       end
+    end
+
+    def set_labels(record, entry)
+      record_labels = record[@labels_key]
+      return nil unless record_labels.is_a?(Hash)
+
+      record_labels.each do |key, value|
+        unless entry.labels.key?(key)
+          record_labels.delete(key)
+          entry.labels[key] = value
+        end
+      end
+
+      record.delete(@labels_key) if record_labels.empty?
     end
 
     # Values permitted by the API for 'severity' (which is an enum).
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -952,6 +952,47 @@ module BaseTest
     end
   end
 
+  def test_labels_from_record
+    setup_gce_metadata_stubs
+    setup_logging_stubs do
+      d = create_driver
+      d.emit(DEFAULT_LABELS_KEY => CUSTOM_LABELS_MESSAGE)
+      d.run
+    end
+    labels = COMPUTE_PARAMS[:labels].merge(CUSTOM_LABELS_MESSAGE)
+    params = COMPUTE_PARAMS.merge(labels: labels)
+    verify_log_entries(1, params, 'labels') do |entry|
+      assert_nil get_fields(entry['jsonPayload'])[DEFAULT_LABELS_KEY], entry
+    end
+  end
+
+  def test_labels_from_record_conflict
+    setup_gce_metadata_stubs
+    setup_logging_stubs do
+      d = create_driver
+      d.emit(DEFAULT_LABELS_KEY => { CONFLICTING_LABEL_KEY => 'a_string' })
+      d.run
+    end
+    verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
+      fields = get_fields(entry['jsonPayload'])
+      labels = get_fields(get_struct(fields[DEFAULT_LABELS_KEY]))
+      assert_equal('a_string', get_string(labels[CONFLICTING_LABEL_KEY]), entry)
+    end
+  end
+
+  def test_labels_from_record_when_not_hash
+    setup_gce_metadata_stubs
+    setup_logging_stubs do
+      d = create_driver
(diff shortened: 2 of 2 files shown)


igorpeshansky:
@arthurdarcet It would help if you outlined your use case in the PR description above (just edit) and explained what parts of it aren't addressed by the existing `labels` and `label_map` configuration options.

qingling128:
From my understanding, below is how `label_map` and `adhoc-labels` (implemented in this PR) are used.

[How label_map is used]

In configuration
```
label_map {
  some_label: user_customized_label
}
```

User's json:
```
{
  "name": "Apache application logs",
  "code": "APACHE",
  "message": "error failing...",
  "some_label": "front-end"
}
```

Resulting json
```
{
  "name": "Apache application logs",
  "code": "APACHE",
  "message": "error failing..."
}
```
Resulting labels
```
labels: {
  user_customized_label: front-end
}
```

[How ad-hoc labels are used]

User's json:
```
{
  "name": "Apache application logs",
  "code": "APACHE",
  "message": "error failing...",
  "logging.googleapis.com/labels": {
    "user_customized_label": "front-end"
  }
}
```

Resulting json
```
{
  "name": "Apache application logs",
  "code": "APACHE",
  "message": "error failing..."
}
```
Resulting labels
```
labels: {
  user_customized_label: front-end
}
```

arthurdarcet:
@igorpeshansky We are running a Kubernetes cluster managed by Google Container Engine. So GKE manages for us the fluentd daemon, and tweaking the configuration is really tricky (to say the least)

Beside, this fluentd plugin might be collecting logs from a very high number of apps ; using a configuration option means that you have to add configuration entries for each app-specific label, something that should instead be managed by the app itself. It seems much cleaner to have the app output "logging.googleapis.com/labels: {my_label: val}" instead so that this daemon picks it up on its own

qingling128:
Hi @arthurdarcet, I've pinged our product owner to let her know the feedback.

As for the possibility to customize Fluentd config in Kubernetes, I believe there is some on-going work (not sure if it's released yet). @crassirostris, would you mind to chime in? The question is whether customers can customize Fluentd config in Kubernetes. If yes, which version do they need to upgrade to? If not, any ETA?

crassirostris:
> As for the possibility to customize Fluentd config in Kubernetes, I believe there is some on-going work (not sure if it's released yet). @crassirostris, would you mind to chime in? The question is whether customers can customize Fluentd config in Kubernetes. If yes, which version do they need to upgrade to? If not, any ETA?

It's possible to customize the fluentd logging agent, here's the documentation for that: https://kubernetes.io/docs/tasks/debug-application-cluster/logging-stackdriver#configuring-stackdriver-logging-agents. I think it's not that complicated after all: you should turn off the default logging solution and deploy the same DaemonSet + ConfigMap you had before, with your changes in the ConfigMap. I'm ready to assist with that and improve the docs if they're less helpful that desired

However, unfortunately, customizing the configuration means it's not possible for GKE to support and predictably update it and you have to do it yourself

arthurdarcet:
@crassirostris The page you linked warns:
> If you’re on GKE and Stackdriver Logging is enabled in your cluster, you cannot change its parameters. Likewise, if you’re not on GKE, but Stackdriver Logging is installed as an addon, you won’t be able to change deployment parameters using Kubernetes API

Which is what I mentioned earlier: for a Kubernetes cluster deployed by Container Engine (GKE), the fluentd DaemonSet is managed by Google. Any attempt to modify it (for instance to update the config with a new label map), will be reverted to the default after a few seconds. The only way we found that let us update the fluentd config was to add a `nodeSelector` that matches nothing to the default DeamonSet. This (for now) does not get overwritten after a few seconds and we end up with a 0/0 DeamonSet. Then we can deploy a new DeamonSet with our config.

I think there are major issues with this:
 - It's not "simple" at all, finding a way to disable the default DeamonSet is at best a "hack", probably won't work long, and is documented nowhere
 - Loosing the GKE-managed updates is a huge pain, GKE customers are paying for exactly this auto-update feature and having to manually diff the config and patch your manual deployment is an issue
 - And this seems the wrong approach to me anyway: These label look like application-level logic to me, and should be managed by each application, not in a global config file. The `logging.googleapi.com/labels` approach fixes this, and does so while leaving the default GKE config alone.

crassirostris:
> It's not "simple" at all, finding a way to disable the default DeamonSet is at best a "hack", probably won't work long, and is documented nowhere

Sorry, that's my bad, documentation is stale, `gcloud` now allows to disable Stackdriver Logging in the running cluster (the command is `gcloud beta container clusters update CLUSTER_NAME --logging-service=none`). I'll fix the docs shortly

> Loosing the GKE-managed updates is a huge pain, GKE customers are paying for exactly this auto-update feature and having to manually diff the config and patch your manual deployment is an issue

Agree, but the problem is that when the custom configuration is broken, there's no clear understanding what "fix" exactly means and what behavior is expected in the end

> And this seems the wrong approach to me anyway: These label look like application-level logic to me, and should be managed by each application, not in a global config file. The logging.googleapi.com/labels approach fixes this, and does so while leaving the default GKE config alone.

That actually sounds reasonable

@igorpeshansky The case is: running the same configuration with different application, each of them setting the labels, operation and so on independently. That'll also help the logging client libraries to adapt the stdout/stderr approach without loosing the ability to set those fields: instead of sending the logging entry directly to the API, they can just write specially formatted json to the stdio

crassirostris:
Another use case: labels can have the information about Docker splitting the log line (details in https://github.com/moby/moby/issues/34855) without changing the payload type

JeanMertz:
I came here searching for this exact same functionality. I too would prefer not to have to change the default logging agent on our GKE cluster, and found this page to help us understand how to set `operation`, `sourceLocation` and `httpRequest` fields:

https://cloud.google.com/logging/docs/agent/configuration#special_fields_in_structured_payloads

Unfortunately, it didn't mention anything about labels, which is why I ended up in this repository, and found out this is currently not supported.

It would be great if `logging.googleapis.com/labels` was supported, as explained by @arthurdarcet.

as a side-note: at our company, we wrote a small Go library for easier Stackdriver logging from our GKE cluster; [zapdriver](https://github.com/blendle/zapdriver), it currently supports setting the `labels` field, but I noticed that it wasn't being picked up by Stackdriver, which is why I started searching for a solution. Unfortunately, Stackdriver + Fluentd + GKE documentation is very hard to find, or at least hard to figure out what is what.

selslack:
@qingling128 @crassirostris what actually prevents this PR from being reviewed and merged?

qingling128:
This PR did not pass PM review the last time since the functionality is considered duplicate from https://cloud.google.com/logging/docs/agent/configuration#label-setup.

The instruction to customize gke is available at https://cloud.google.com/monitoring/kubernetes-engine/customizing.

If there is a use case that is not covered, we could revive the feature request.

JeanMertz:
I'd like to add my voice that it's strange that this isn't supported out of the box by GKE. Sure, we can start fiddling with the logging agent configuration, but the reason we pay for GKE combined with Stackdriver is that it's supposed to "just work". I don't consider this duplicate functionality, as one requires manual configuration on GKE, the other (this PR) doesn't.

See also: https://github.com/blendle/zapdriver/issues/4

qingling128:
@arthurdarcet & @JeanMertz - Thanks for your inputs. These are valuable to us. I've added this feature request to our next product sync. I've summarize the issue as below.

>Users want a way to set labels by including them directly in the log entries.
e.g. by specifying "logging.googleapis.com/labels: {my_label: val}", the agent can pick up the labels automatically.
>All the [exisiting solutions](https://cloud.google.com/logging/docs/agent/configuration#label-setup) we provide require modifying the configuration. One agent could be collecting logs from many applications. Labels could be application specific. So them should be managed by the applications themselves and be isolated from the agent configuration. Besides, reconfigurating and restarting all agents for every label change is not trivial work and hard to maintain.
>This is especially hard in the GKE environment, because manually configuring the agent is feasible but not ideal. An out-of-the-box experience is preferred.

Feel free to add or emphasize any other use cases if I missed any.

qingling128:
Due to the holiday schedule, our next product sync would be on Jan 8.

qingling128:
Updates on this feature request: It has been approved by product. The related development and documentation work will be prioritized.

qingling128:
This is being implemented in https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/292.

PR #154, opened 2017-08-30
Add severity translations for java.util.logging.

jmikula:
Added severity translation from "[SEVERE](https://docs.oracle.com/javase/8/docs/api/java/util/logging/Level.html#SEVERE)" to "[ERROR](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity)."

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1285,6 +1285,7 @@ module Fluent
     # Translates other severity strings to one of the valid values above.
     SEVERITY_TRANSLATIONS = {
       # log4j levels (both current and obsolete).
+      'SEVERE' => 'ERROR',
       'WARN' => 'WARNING',
       'FATAL' => 'CRITICAL',
       'TRACE' => 'DEBUG',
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1285,6 +1285,7 @@ def set_log_entry_fields(record, entry)
     # Translates other severity strings to one of the valid values above.
     SEVERITY_TRANSLATIONS = {
       # log4j levels (both current and obsolete).
+      'SEVERE' => 'ERROR',
```
qingling128: `SEVERE` does not seem to be a standard level for log4j: https://www.tutorialspoint.com/log4j/log4j_logging_levels.htm

Maybe it could belong to "other misc".

Also, why is `SEVERE` mapped to `ERROR` instead of `CRITICAL`?

igorpeshansky: This is from [`java.util.logging`](https://docs.oracle.com/javase/8/docs/api/java/util/logging/Level.html#SEVERE), but +1 on moving it into a separate section below (right before `# nginx levels ...`).

dhrupadb: Agreed. We could add a special case for this so as to not confuse them with log4j levels.

jmikula:
I added a section for java.util.logging and a mapping from CONFIG -> DEBUG.
An argument could be made for other mappings other than what I have in the PR. Open to feedback.

qingling128:
LGTM.

PR #158, opened 2017-09-20
Added stackdriver_retried_entries_count counter

jkohen:
Igor, do you have any opinion on using singular nouns for this counter, when we use plural for the others? Singular is in line with Stackdriver best practices.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -583,7 +583,7 @@ module Fluent
             entries_count = entries.length
             client.write_log_entries(write_request)
             increment_successful_requests_count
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -359,6 +359,11 @@ def configure(conf)
         @dropped_entries_count = registry.counter(
           :stackdriver_dropped_entries_count,
           'A number of log entries dropped by the Stackdriver output plugin')
+        @log_entry_retry_count = registry.counter(
```
igorpeshansky: I would call it `@retried_entries_count` for consistency... But what you have is fine too.

PR #159, opened 2017-09-20
Add logging for retries.

qingling128:
Should we log retries as `info` or `warn`?

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -575,6 +575,7 @@ module Fluent
               entries: entries
             )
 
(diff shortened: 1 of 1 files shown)


igorpeshansky:
I would log them as `debug`, actually. We can always turn it on with `-v`, and it doesn't spam the logs by default.

qingling128:
Changed to `debug`. PTAL.

jkohen:
You could have asked me to add the logging in PR #158, which touches the same code for the same purpose.

qingling128:
Ah, this is for debugging some customer issue. We kind of want it asap.

jkohen:
I also want PR 158 ASAP :)

Issue #131, opened 2017-05-31
LogEntry timestamps truncated to seconds


vdm:
New GCE VM, just installed latest version Stackdriver Logging agent. This reproduces across several different VMs, agents, Debian and Centos.

Sub-second timestamps were included in fluentd <0.14, which google-fluentd appears to use. Why are timestamps still truncated in Stackdriver Logging?

qingling128:
Hi @vdm, sub-second timestamps should be supported in the latest version Stackdriver Logging agent. Can you give some example of the issue you saw? Sample entries and resulting logs for example.

ahprez:
I seem to have the same problem.  My app runs on a GCE VM which has the latest google-fluentd 1.5.17 installed. Here is a log entry sample from our app, followed by the <source> configuration: 

2017-08-09 21:11:37,849 thread-1  INFO [MyApp] Hello World!
 
<source>
  type tail
  format multiline
  format_firstline /\d{4}-\d{1,2}-\d{1,2}/
  format1 /^(?<time>\d{4}-\d{1,2}-\d{1,2} \d{1,2}:\d{1,2}:\d{1,2},\d{3}) (?<thread>.*) (?<severity>[^\s]+) \[(?<serviceName>.*)\] (?<message>.*)/ 
  time_format %F %T,%L
  path /foo/bar/myservice.log
  pos_file /var/lib/google-fluentd/pos/myservice.pos
  read_from_head true
  tag myservice
</source>

In Stackdriver, the time for this log entry shows as 21:11:37.000.   Any ideas why the milliseconds aren't appearing?

sayap:
Same here. With google-fluentd 1.5.17-1, the timestamps are truncated to seconds.

sayap:
Managed to get sub-second resolution after making this change to fluentd:
```diff
--- a/lib/fluent/time.rb
+++ b/lib/fluent/time.rb
@@ -73,7 +73,7 @@ module Fluent
     end
 
     def to_msgpack(io = nil)
-      @sec.to_msgpack(io)
+      to_f.to_msgpack(io)
     end
 
     def to_msgpack_ext
```

qingling128:
Hi @sayap, which Fluentd version is this? You can check it with:
```
$ /opt/google-fluentd/embedded/bin/gem list | grep fluentd
fluentd (0.14.19)
```

[...]

PR #161, opened 2017-09-27
solution to #131 truncated timestamps

vdm:
#131 by @mururu and @sayap

Tested by editing installed agent plugin code. Saw correct timestamps start to show up for fluentd log entries in Google Cloud Logging webapp.


vdm:
CLA signed.

PR #163, opened 2017-10-09
Upgrade google-api-client to 0.14.5 and google-cloud-logging to 1.2.3.

qingling128:
These upgrades are required for the coming partial success support.

Changes in these upgrades that affect us:
- `erb` is no longer bundled in Ruby core as of Ruby 2.2, which is required by these new gems. So we have to require it explicitly instead.
- The types of the following fields have been changed from `integer` to `String`: `cache_fill_bytes`, `response_size` and `request_size` in `HttpRequest`, `line` in `SourceLocation`. So we had to change our tests accordingly. https://github.com/google/google-api-ruby-client/issues/619 is created to double check these changes are intended. It feels weird.

### DIFF
--- a/test/plugin/test_out_google_cloud.rb
+++ b/test/plugin/test_out_google_cloud.rb
@@ -251,7 +251,7 @@ class GoogleCloudOutputTest < Test::Unit::TestCase
 
   private
 
-  WRITE_LOG_ENTRIES_URI = 'https://logging.googleapis.com/v2beta1/entries:write'
+  WRITE_LOG_ENTRIES_URI = 'https://logging.googleapis.com/v2/entries:write'
 
   def rename_key(hash, old_key, new_key)
     hash.merge(new_key => hash[old_key]).reject { |k, _| k == old_key }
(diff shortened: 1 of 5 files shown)


qingling128:
BTW, we should not be blocked by the question in google/google-api-ruby-client#619 or any potential changes coming from this. Because the change in this PR is not visible to clients and won't affect them anyway. It's the format in which Logging Agent talks to Logging API.

test/plugin/test_out_google_cloud.rb
```diff
@@ -316,4 +316,20 @@ def get_number(field)
   def null_value
     nil
   end
+
+  # 'responseSize' and 'requestSize' are Integers in the gRPC proto, yet Strings
+  # in REST API client.
```
igorpeshansky: Want to reference the issue you've opened (https://github.com/google/google-api-ruby-client/issues/619) in a comment here and before `source_location_message`?

qingling128: Sure thing. Just added this:

```
  # TODO(lingshi): Address this accordingly once the following  question is
  # answered: https://github.com/google/google-api-ruby-client/issues/619.
  # If this discrepency is legit, add some comments to explain the reason.
  # Otherwise once the discrepency is fixed, we need to upgrade to that version
  # and change our tests accordingly.
```

qingling128: Changed `lingshi` to `qingling128`.....

test/plugin/test_out_google_cloud.rb
```diff
@@ -316,4 +316,25 @@ def get_number(field)
...
+  # TODO(qingling128): Address this accordingly once the following  question is
```
igorpeshansky: Uber-nit: two spaces between "following" and "question".

test/plugin/test_out_google_cloud.rb
```diff
@@ -316,4 +316,25 @@ def get_number(field)
...
+  # answered: https://github.com/google/google-api-ruby-client/issues/619.
+  # If this discrepency is legit, add some comments to explain the reason.
```
igorpeshansky: Typo: discrepency. Also below.

test/plugin/test_out_google_cloud.rb
qingling128: Fixed.

test/plugin/test_out_google_cloud.rb
qingling128: Fixed.

PR #165, opened 2017-10-10
Partial success support for REST path.

qingling128:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -169,6 +169,10 @@ module InternalConstants
           'Google::Apis::LoggingV2::LogEntryOperation'
         ]
       }
+
+      # The name of the WriteLogEntriesPartialErrors field in the error details.
+      PARTIAL_ERROR_KEY = 'type.googleapis.com/' \
```
igorpeshansky: All of the `*_KEY` constants represent fields in the log record. This is a field in the response, so maybe `PARTIAL_ERROR_FIELD`?

igorpeshansky: Let's break right after the `=`, so the string fits on one line.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -281,6 +285,11 @@ module InternalConstants
     # Cloud Logging API.
     config_param :use_grpc, :bool, :default => false
 
+    # Whether valid entries should be written even if some other entries fail
+    # due to INVALID_ARGUMENT or PERMISSION_DENIED errors when communicating to
+    # the Cloud Logging API. This is highly recommended.
+    config_param :partial_success, :bool, :default => false
```
igorpeshansky: This currently only works for the non-gRPC path. Should we add some sort of validation that prints a warning that this key will be ignored if `use_grpc` is on?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -627,12 +636,16 @@ def write(chunk)
                 log_name: log_name,
                 resource: group_level_resource,
                 labels: group_level_common_labels,
+                partial_success: @partial_success,
                 entries: entries)
             entries_count = entries.length
 
             # TODO: RequestOptions
```
igorpeshansky: Is this TODO covered by the call below?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
...
+    #           ogging.v2.WriteLogEntriesPartialErrors\" value: \"\\n\\034\\010\
+    #           \000\\022\\030\\010\\007\\022\\024User not authorized.\\n-\\010\
+    #           \001\\022)\\010\\003\\022%Log name contains illegal character :\
+    #           " } }"
+    #       }
+    #     ]
+    #   }
+    # }
+    #
+    # In this exxample, the logEntryErrors we are extracting should be:
```
igorpeshansky: Typo: exxample.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
...
 
+    # Extract a map of error details from an potentially partially successful
+    # request. The keys in this map are the indexes of the failed log entries in
+    # the original json payload. The values in this map are the error code and
+    # error message.
+    #
+    # A sample error.body looks like:
+    # {
+    #   "error": {
+    #     "code": 403,
```
igorpeshansky: Is this really 403? I thought that translated to `Google::Apis::AuthorizationError`, which would have been caught earlier...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
...
+                          error: error_message
+              end
+            end
           end
         end
       end
     end
 
+    # Extract a map of error details from an potentially partially successful
+    # request. The keys in this map are the indexes of the failed log entries in
```
igorpeshansky: s/the indexes/stringified indexes/...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
...
+    # A sample error.body looks like:
+    # {
+    #   "error": {
+    #     "code": 403,
+    #     "message": "User not authorized.",
+    #     "status": "PERMISSION_DENIED",
+    #     "details": [
+    #       {
+    #         "@type": "type.googleapis.com/google.logging.v2.WriteLogEntriesPar
+    #           tialErrors",
```
igorpeshansky: Are you breaking up this string because of rubocop?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
...
+    #     "code": 7,
+    #     "message": "User not authorized."
+    #   },
+    #   "1": {
+    #     "code": 3,
+    #     "message": "Log name contains illegal character :"
+    #   }
+    # }
+    def extract_log_entry_errors(error)
+      error_details = parse_json_or_nil(error.body)['error']['details']
```
igorpeshansky: I wouldn't use `parse_json_or_nil` -- that function is used for JSON detection, and here we know the body is actually valid JSON. Plus, you lose error information. Just use `JSON.parse` directly, and rescue `JSON::ParserError` instead of a broad catch-all.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
...
+    rescue => e
+      @log.warn 'Failed to extract log entry errors from the error details' \
+                " json: #{error.body}.", error: e
+      {}
+    end
+
+    # Given the logEntryErrors, constuct a map from errors to a list of indexes
+    # of log entries that failed for this specific error.
+    #
+    # The logEntryErrors should be in the format of:
```
igorpeshansky: I wouldn't bother repeating this here -- maybe construct one example that shows the full details, the extracted errors, and the resulting map, and put it above these two methods?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
...
+    # }
+    def extract_log_entry_errors(error)
+      error_details = parse_json_or_nil(error.body)['error']['details']
+      partial_errors = error_details.detect do |error_detail|
+        error_detail['@type'] == PARTIAL_ERROR_KEY
+      end
+      partial_errors['logEntryErrors']
+    rescue => e
+      @log.warn 'Failed to extract log entry errors from the error details' \
+                " json: #{error.body}.", error: e
```
igorpeshansky: s/json/JSON/.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
           rescue Google::Apis::ClientError => error
             # Most ClientErrors indicate a problem with the request itself and
             # should not be retried.
-            increment_dropped_entries_count(entries_count)
-            @log.warn "Dropping #{entries_count} log message(s)",
-                      error_class: error.class.to_s, error: error.to_s
+            error_details_map = {}
+            if @partial_success
+              error_details_map = construct_error_details_map(
+                extract_log_entry_errors(error))
```
igorpeshansky: The log entry errors in the gRPC case will be different -- do you want to abstract that difference away? Or name this function `extract_log_entry_errors_rest` or something?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
...
+    #     "message": "Log name contains illegal character :"
+    #   },
+    #   "2": {
+    #     "code": 3,
+    #     "message": "Log name contains illegal character :"
+    #   }
+    # }
+    #
+    # The constructed map would be:
+    # // Keys are [error_code, error_message]. Values are a list of indexes of
```
igorpeshansky: Why do we care about error codes? They're gRPC errors, so will be confusing in the REST case, won't they?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
...
+            if error_details_map.empty?
+              increment_dropped_entries_count(entries_count)
+              @log.warn "Dropping #{entries_count} log message(s)",
+                        error_class: error.class.to_s, error: error.to_s
+            else
+              error_details_map.each do |(error_code, error_message), indexes|
+                partial_error_count = indexes.length
+                increment_dropped_entries_count(partial_error_count)
+                @log.warn "Dropping #{partial_error_count} log message(s)",
+                          error_code: "google.rpc.Code.#{error_code}",
```
igorpeshansky: Careful -- these are numeric codes, not symbolic names.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
...
+    #   ]: ["0"],
+    #   [
+    #     3,
+    #     'Log name contains illegal character :'
+    #   ]: ["1", "2"]
+    # }
+    def construct_error_details_map(log_entry_errors)
+      error_details_map = {}
+      log_entry_errors.each do |index, error|
+        error_key = [error['code'], error['message']]
```
igorpeshansky: `.freeze()`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
...
+    #     7,
+    #     'User not authorized.'
+    #   ]: ["0"],
+    #   [
+    #     3,
+    #     'Log name contains illegal character :'
+    #   ]: ["1", "2"]
+    # }
+    def construct_error_details_map(log_entry_errors)
+      error_details_map = {}
```
igorpeshansky:
```ruby
error_details_map = Hash.new{|h,k| h[k] = []}
```
Then you can avoid the `.key?` conditional below.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +678,140 @@ def write(chunk)
...
+      log_entry_errors.each do |index, error|
+        error_key = [error['code'], error['message']]
+        if error_details_map.key?(error_key)
+          error_details_map[error_key] << index
```
igorpeshansky: Technically, it would eventually be useful to get indexes as ints, not strings. But for now you're only using the length of the array anyway, so let's just add a TODO to convert them to ints?

lib/fluent/plugin/out_google_cloud.rb
qingling128: Changed to `PARTIAL_ERROR_FIELD`.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Ah, this `TODO` is really old. Just removed it.

lib/fluent/plugin/out_google_cloud.rb
qingling128: I was going to name the other one `XX_grpc`, which is consistent with how we name the other functions. WDYT?

lib/fluent/plugin/out_google_cloud.rb
qingling128: Yeah, I'm aware of that. Do we want to do the translation here? Or change this to something like `"google.rpc.Code.[#{error_code}]"`

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: `401` maps to [Google::Apis::AuthorizationError](http://www.rubydoc.info/github/google/google-api-ruby-client/Google/Apis/AuthorizationError). Every other `4XX` maps to [Google::Apis::ClientError](http://www.rubydoc.info/github/google/google-api-ruby-client/Google/Apis/ClientError).

This sample is the output from this piece of code:

```ruby
$ cat non_grpc.rb
require 'google/apis/logging_v2'
require 'googleauth'

client = Google::Apis::LoggingV2::LoggingService.new
client.authorization = Google::Auth.get_application_default(
  'https://www.googleapis.com/auth/logging.write')

resource = Google::Apis::LoggingV2::MonitoredResource.new(
  type: 'gce_instance',
  labels: {}
)

entries = [
  Google::Apis::LoggingV2::LogEntry.new(
    {
      resource: resource,
      log_name: 'projects/gapic-test/logs/my-log',
      text_payload: 'pass'
    }
  ),
  Google::Apis::LoggingV2::LogEntry.new(
    {
      resource: resource,
      log_name: 'projects/gapic-test/logs/bad:log',
      text_payload: 'fail'
    }
  )
]

write_request = \
  Google::Apis::LoggingV2::WriteLogEntriesRequest.new(
    log_name: 'projects/gapic-test/logs/my-log',
    resource: resource,
    labels: {},
    partial_success: true,
    entries: entries)

begin
  client.write_entry_log_entries(
    write_request,
    options: {
      api_format_version: '2'
    }
  )
rescue Google::Apis::ClientError => error
  puts "error:\n#{error.inspect}\n"
  puts "error body =\n#{error.body}\n"
end
```

```
$ ruby non_grpc.rb
error:
#<Google::Apis::ClientError: PERMISSION_DENIED: User not authorized.>
error body =
{
  "error": {
    "code": 403,
    "message": "User not authorized.",
    "status": "PERMISSION_DENIED",
    "details": [
      {
        "@type": "type.googleapis.com/google.logging.v2.WriteLogEntriesPartialErrors",
        "logEntryErrors": {
          "0": {
            "code": 7,
            "message": "User not authorized."
          },
          "1": {
            "code": 3,
            "message": "Log name contains illegal character :"
          }
        }
      },
      {
        "@type": "type.googleapis.com/google.rpc.DebugInfo",
        "detail": "[ORIGINAL ERROR] generic::permission_denied: User not authorized. [google.rpc.error_details_ext] { message: \"User not authorized.\" details { type_url: \"type.googleapis.com/google.logging.v2.WriteLogEntriesPartialErrors\" value: \"\\n\\034\\010\\000\\022\\030\\010\\007\\022\\024User not authorized.\\n-\\010\\001\\022)\\010\\003\\022%Log name contains illegal character :\" } }"
      }
    ]
  }
}
```

lib/fluent/plugin/out_google_cloud.rb
qingling128: Ah.... fat fingers. Changed.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Removed the duplicated example and changed the phrasing to:

```
    # For logEntryErrors as in the example above (in the comment section of the
    # function extract_log_entry_errors), the constructed map would be:
    # {
    #   [
    #     7,
    #     'User not authorized.'
    #   ]: ["0"],
    #   [
    #     3,
    #     'Log name contains illegal character :'
    #   ]: ["1", "2"]
    # }
```

lib/fluent/plugin/out_google_cloud.rb
qingling128: If users make direct calls to the Cloud Logging API, this is what they get back. The idea is to have a consistent error code I think. Do we want to hide it? It kind of helps with debugging. Or do we want to translate it to something REST-ish?

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Yes. Rubocop complains about it.

qingling128:
PTAL.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: We should eventually standardize on the naming, but I guess that's fine for now.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Let's go with `"google.rpc.Code[#{error_code}]"`.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Will we always get a 403 with partial errors, or is this just an example?

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Sigh.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Careful, the example above only had 2 error indexes, and in your map you have 3. Should we update the example to have two illegal character errors?

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: I guess just dumping the numeric error codes is fine. We can't translate them without introducing a dependency on gRPC, can we?

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Let's also add a note in the comment that this only works without `use_grpc`. We can remove that note whenever we remove the check.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Cool cool. I'll do that in a separate PR.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: This is just an example. I just added some comment to clarify:

```
    # Note that the root level "code", "message" and "status" simply matches the
    # root cause of the first failed log entry. For example, if we switch the
    # order of the log entries, then we will get:
    # {
    #    "error" : {
    #       "code" : 400,
    #       "message" : "Log name contains illegal character :",
    #       "status" : "INVALID_ARGUMENT",
    #       "details": ......
    #    }
    # }
    # We will ignore it anyway and look at the details instead which includes
    # info for all failed log entries.
```

lib/fluent/plugin/out_google_cloud.rb
qingling128: Hahahhahaha.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Meant to do that. Thanks for catching this.

lib/fluent/plugin/out_google_cloud.rb
qingling128: I think we should be fine with `"google.rpc.Code[#{error_code}]"` for now.

qingling128:
PTAL.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Let's just open a tracking bug so we don't forget, but not waste time on it now.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -281,6 +285,14 @@ module InternalConstants
     # Cloud Logging API.
     config_param :use_grpc, :bool, :default => false
 
+    # Whether valid entries should be written even if some other entries fail
+    # due to INVALID_ARGUMENT or PERMISSION_DENIED errors when communicating to
+    # the Cloud Logging API. This is highly recommended.
+    # TODO(qingling128): Remove the comment below when the support is added for
```
igorpeshansky: This TODO would appear in the generated docs. I would instead mention this in the TODO on the warning.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -627,12 +647,14 @@ def write(chunk)
                 log_name: log_name,
                 resource: group_level_resource,
                 labels: group_level_common_labels,
+                partial_success: @partial_success,
                 entries: entries)
             entries_count = entries.length
-
```
igorpeshansky: Let's keep this blank line.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -627,12 +647,14 @@ def write(chunk)
...
-            # TODO: RequestOptions
             begin
-              client.write_entry_log_entries(write_request)
+              client.write_entry_log_entries(
+                write_request,
+                options: { api_format_version: '2' }
+              )
             rescue Google::Apis::Error => error
```
igorpeshansky: Let's add a blank line before this `rescue`, for consistency.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: I just confirmed that it's not possible to get a 401 on only one of the entries -- it always applies to the whole request. So we don't have to worry about parsing partial errors when catching an `AuthorizationError`. But what about `ServerError`? What if the first entry has a 500?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +687,147 @@ def write(chunk)
...
+    #       },
+    #       {
+    #         "@type": "type.googleapis.com/google.rpc.DebugInfo",
+    #         "detail": ......
+    #       }
+    #     ]
+    #   }
+    # }
+    #
+    # Note that the root level "code", "message" and "status" simply matches the
```
igorpeshansky: s/`Note that the`/`The`/.

igorpeshansky: Oxford comma before `and`, please.

igorpeshansky: s/`matches`/`match`/ (this refers to all three of `code`, `message`, and `status`).

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +687,147 @@ def write(chunk)
...
+    # root cause of the first failed log entry. For example, if we switch the
```
igorpeshansky: s/`switch` ... `will`/`switched` ... `would`/.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +687,147 @@ def write(chunk)
...
+    end
+
+    # Given the logEntryErrors, constuct a map from errors to a list of indexes
+    # of log entries that failed for this specific error.
+    #
+    # The keys in this map are an array of [error_code, error_message]. The
+    # values in this map are a list of indexes of log entries that failed due to
+    # this error.
+    #
+    # For logEntryErrors as in the example above (in the comment section of the
```
igorpeshansky: s/`comment section of the function`/`header comment for`/

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +687,147 @@ def write(chunk)
...
+    rescue => e
+      @log.warn 'Failed to extract log entry errors from the error details' \
+                " JSON: #{error.body}.", error: e
+      {}
+    end
+
+    # Given the logEntryErrors, constuct a map from errors to a list of indexes
+    # of log entries that failed for this specific error.
+    #
+    # The keys in this map are an array of [error_code, error_message]. The
```
igorpeshansky: s/`an array of [error_code, error_message]`/`[error_code, error_message] pairs`/

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +687,147 @@ def write(chunk)
...
+    # values in this map are a list of indexes of log entries that failed due to
```
igorpeshansky: s/`. The values in this map are`/`, and the values are`/

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +687,147 @@ def write(chunk)
...
+    # this error.
+    #
+    # For logEntryErrors as in the example above (in the comment section of the
+    # function extract_log_entry_errors), the constructed map would be:
+    # {
+    #   [
+    #     7,
+    #     'User not authorized.'
+    #   ]: ["0"],
```
igorpeshansky: This is a Ruby hash, not a JSON object, so let's use single quotes around the strings. In fact, the whole thing would fit on a couple of lines:
```ruby
    # {
    #   [7, 'User not authorized.']: ['0'],
    #   [3, 'Log name contains illegal character :']: ['1', '3']
    # }
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +687,147 @@ def write(chunk)
...
+    #           },
+    #           "3": {
+    #             "code": 3,
+    #             "message": "Log name contains illegal character :"
+    #           }
+    #         }
+    #       },
+    #       {
+    #         "@type": "type.googleapis.com/google.rpc.DebugInfo",
+    #         "detail": ......
```
igorpeshansky: Too many `.`s. Just `...` will do. Also below.

lib/fluent/plugin/out_google_cloud.rb
qingling128: b/67788132 created.

lib/fluent/plugin/out_google_cloud.rb
qingling128: From the documentation of `partial_success` in the `Stackdriver Logging API` site:

>partial success: Optional. Whether valid entries should be written even if some other entries fail due to INVALID_ARGUMENT or PERMISSION_DENIED errors. If any entry is not written, then the response status is the error associated with one of the failed entries and the response includes error details keyed by the entries' zero-based index in the entries.write method.

It sounds like only `INVALID_ARGUMENT` or `PERMISSION_DENIED` will trigger partial success. I took a look at the extra implementation and unit tests of the Logging API write request handler, it also suggests so.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: This one and the rest are not at the same level though. It might be more confusing that way.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

qingling128:
PTAL.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,141 @@ def write(chunk)
...
+      error_details = JSON.parse(error.body)['error']['details']
+      partial_errors = error_details.detect do |error_detail|
+        error_detail['@type'] == PARTIAL_ERROR_FIELD
+      end
+      partial_errors['logEntryErrors']
+    rescue JSON::ParserError => e
+      @log.warn 'Failed to extract log entry errors from the error details' \
+                " because of JSON parsing error: #{error.body}.", error: e
+      {}
+    rescue => e
```
igorpeshansky: Why do you need the second `rescue`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,141 @@ def write(chunk)
...
+      @log.warn 'Failed to extract log entry errors from the error details' \
+                " JSON: #{error.body}.", error: e
+      {}
+    end
+
+    # Given the logEntryErrors, constuct a map from errors to a list of indexes
```
igorpeshansky: Typo: constuct.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,141 @@ def write(chunk)
...
+    # }
+    def construct_error_details_map(log_entry_errors)
+      error_details_map = Hash.new { |h, k| h[k] = [] }
+      log_entry_errors.each do |index, error|
+        error_key = [error['code'], error['message']].freeze
+        # TODO(qingling128): Convert indexes to integers.
+        error_details_map[error_key] << index
+      end
+      error_details_map
+    rescue => e
```
igorpeshansky: What exception are we expecting here?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,141 @@ def write(chunk)
...
+      @log.warn 'Failed to extract log entry errors from the error details' \
+                " JSON: #{error.body}.", error: e
+      {}
+    end
+
+    # Given the logEntryErrors, constuct a map from errors to a list of indexes
+    # of log entries that failed for this specific error.
+    #
+    # The keys in this map are [error_code, error_message] pairs. The values in
+    # this map are a list of indexes of log entries that failed due to this
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
qingling128: I added the warning to be:

```
      @log.warn 'Failed to parse the log entry errors because logEntryErrors ' \
                "is not a hash or it does not contain 'code' and 'message' " \
                'fields.', error: e
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,141 @@ def write(chunk)
...
+    rescue => e
+      @log.warn 'Failed to extract log entry errors from the error details' \
+                " JSON: #{error.body}.", error: e
+      {}
+    end
+
+    # Given the logEntryErrors, constuct a map from errors to a list of indexes
+    # of log entries that failed for this specific error.
+    #
+    # The keys in this map are [error_code, error_message] pairs. The values in
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Because the `error`, `details` or `logEntryErrors` fields might not exist. Or there is no error details with `@type` equal to `PARTIAL_ERROR_FIELD`.

lib/fluent/plugin/out_google_cloud.rb

qingling128:
PTAL.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: In those cases, you'd get a `NoMethodError` on `NilClass` (or a `TypeError` if the parsed error is not a Hash), which is not going to be very useful to the user. But we could prevent those by returning an empty hash (`{}`) in all those cases, e.g.:
```ruby
    def ensure_hash(value)
      if value.is_a?(Hash) then value else {} end
    end

    def extract_log_entry_errors(error)
      error_body = ensure_hash(JSON.parse(error.body))
      error = ensure_hash(error_body['error'])
      error_details = ensure_hash(error['details'])
      partial_errors = error_details.detect do |error_detail|
        error_detail['@type'] == PARTIAL_ERROR_FIELD
      end
      ensure_hash(partial_errors['logEntryErrors'])
    rescue JSON::ParserError => e
      @log.warn 'Failed to extract log entry errors from the error details' \
                " because of JSON parsing error: #{error.body}.", error: e
      {}
    end
```
This will also make sure that the value going into `construct_error_details_map` is a Hash.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: As I said above, the exception you'd get would not be very useful to the user. You could use the same `ensure_hash` method, e.g.:
```ruby
    def construct_error_details_map(log_entry_errors)
      error_details_map = Hash.new { |h, k| h[k] = [] }
      log_entry_errors.each do |index, error|
        error_hash = ensure_hash(error)
        unless error_hash['code'] and error_hash['message'] then
          @log.warn "Entry with index #{index} is missing 'code' or 'message'"
                    ' field.'
          return {} # Can also be 'next' or 'break'...
        end
        error_key = [error_hash['code'], error_hash['message']].freeze
        # TODO(qingling128): Convert indexes to integers.
        error_details_map[error_key] << index
      end
      error_details_map
    end
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,143 @@ def write(chunk)
...
+    def construct_error_details_map(log_entry_errors)
+      error_details_map = Hash.new { |h, k| h[k] = [] }
+      log_entry_errors.each do |index, error|
+        error_key = [error['code'], error['message']].freeze
+        # TODO(qingling128): Convert indexes to integers.
+        error_details_map[error_key] << index
+      end
+      error_details_map
+    rescue => e
+      @log.warn 'Failed to parse the log entry errors because logEntryErrors ' \
```
igorpeshansky: If you decide to keep this message, please put the spaces at the start of the lines.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,171 @@ def write(chunk)
...
+    #     "code": 3,
+    #     "message": "Log name contains illegal character :"
+    #   }
+    # }
+    def extract_log_entry_errors(error)
+      error_body = ensure_hash(JSON.parse(error.body), 'error body')
+      error = ensure_hash(error_body['error'], 'error')
+      error_details = ensure_array(error['details'], 'error details')
+      partial_errors = error_details.detect do |error_detail|
+        ensure_hash(error_detail, 'error detail')
```
igorpeshansky: Simpler:
```ruby
        ensure_hash(error_detail, 'error detail')['@type'] == PARTIAL_ERROR_FIELD
```

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: You still have the unqualified `rescue` block...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,159 @@ def write(chunk)
...
+                " is not a hash or it does not contain 'code' and 'message'" \
+                ' fields.', error: e
+      {}
+    end
+
+    def ensure_hash(value, error_name)
+      if value.is_a?(Hash)
+        value
+      else
+        @log.warn "Non-hash value detected for #{error_name}: #{value.inspect}."
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,159 @@ def write(chunk)
...
+        end
+        error_key = [error_hash['code'], error_hash['message']].freeze
+        # TODO(qingling128): Convert indexes to integers.
+        error_details_map[error_key] << index
+      end
+      error_details_map
+    end
+
+    def ensure_array(value)
+      ensure_non_empty(Array.try_convert(value))
```
igorpeshansky:
```ruby
Array.try_convert(value) or fail JSON::ParserError
```
Then you don't need `ensure_non_empty`. Same for `ensure_hash`:
```ruby
Hash.try_convert(value) or fail JSON::ParserError
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,159 @@ def write(chunk)
...
+    #     "message": "Log name contains illegal character :"
+    #   },
+    #   "3": {
+    #     "code": 3,
+    #     "message": "Log name contains illegal character :"
+    #   }
+    # }
+    def extract_log_entry_errors(error)
+      error_body = ensure_hash(JSON.parse(error.body))
+      error = ensure_hash(error_body['error'])
```
igorpeshansky:
```ruby
error = ensure_hash(ensure_hash(JSON.parse(error.body))['error'])
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,159 @@ def write(chunk)
...
+      error_details = ensure_array(error['details'])
+      partial_errors = error_details.detect do |error_detail|
```
igorpeshansky:
```ruby
partial_errors = ensure_array(error['details']).detect do |error_detail|
  ensure_hash(error_detail)['@type'] == PARTIAL_ERROR_FIELD
end
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,159 @@ def write(chunk)
...
+        ensure_hash(error_detail)['@type'] == PARTIAL_ERROR_FIELD
+      end
+      unless partial_errors
+        @log.warn 'Failed to detect an error detail with type' \
```
igorpeshansky: This will warn twice -- once here, and once when catching the exception from the later `ensure_hash`.
My suggestion is to throw the exception from the _ifnone_ function of `Array#detect`:
```ruby
partial_errors = ensure_array(error['details']).detect(
  lambda { fail JSON::ParserError.new("No type #{PARTIAL_ERROR_FIELD}") }
) do |error_detail|
  ensure_hash(error_detail)['@type'] == PARTIAL_ERROR_FIELD
end
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,159 @@ def write(chunk)
...
+          " #{PARTIAL_ERROR_FIELD}."
+      end
+      ensure_hash(ensure_hash(partial_errors)['logEntryErrors'])
+    rescue JSON::ParserError => e
```
igorpeshansky: We can put this `rescue` around the `construct_error_details_map(extract_log_entry_errors(error))` call instead. That way it'll also catch the errors thrown by the `ensure_hash` in `construct_error_details_map`...
We just need to remove the "`JSON parsing error`" bit from the error message, because it'll be apparent from the exception object.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Ah. I tried this already. Rubocop does not like this type of syntax.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,14 +688,155 @@ def write(chunk)
...
+    #   },
+    #   "3": {
+    #     "code": 3,
+    #     "message": "Log name contains illegal character :"
+    #   }
+    # }
+    def extract_log_entry_errors(error)
+      error = ensure_hash(ensure_hash(JSON.parse(error.body))['error'])
+      partial_errors = ensure_array(error['details']).detect(
+        -> { fail JSON::ParserError, "No type #{PARTIAL_ERROR_FIELD}." }
```
igorpeshansky: Neat, I didn't know about this lambda syntax...

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Sorry for being unclear -- I meant putting it around the *call*, in line 696... You can just do that in the conditional block:
```ruby
            if @partial_success
              @log.debug 'Encountered error when talking to Stackdriver' \
                         ' Logging API.',
                         error: error.class, error_body: error.body
              error_details_map = construct_error_details_map(
                extract_log_entry_errors(error))
            rescue JSON::ParserError => e
              @log.warn 'Failed to extract log entry errors from the error' \
                        " details: #{error.body}.", error: e
            end
```

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: This seems to fix it:
```diff
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -825,16 +825,11 @@ module Fluent
     end
 
     def ensure_array(value)
-      ensure_non_empty(Array.try_convert(value))
+      Array.try_convert(value) || (fail JSON::ParserError, "#{value.class}")
     end
 
     def ensure_hash(value)
-      ensure_non_empty(Hash.try_convert(value))
-    end
-
-    def ensure_non_empty(converted_value)
-      fail JSON::ParserError unless converted_value
-      converted_value
+      Hash.try_convert(value) || (fail JSON::ParserError, "#{value.class}")
     end
 
     private
```

lib/fluent/plugin/out_google_cloud.rb
qingling128: I did try this. Yet I got `Avoid more than 3 levels of block nesting` warning from `Rubocop`.

On a second thought, there is no real reason to keep the two functions separate. I just combined them so I can have one `rescue` that covers the whole function.

PTAL.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1711,6 +1747,118 @@ def convert_to_utf8(input)
...
+          return {}
+        end
+        error_key = [error_hash['code'], error_hash['message']].freeze
+        # TODO(qingling128): Convert indexes to integers.
+        error_details_map[error_key] << index
+      end
+      error_details_map
+    rescue JSON::ParserError => e
+      @log.warn 'Failed to extract log entry errors from the error details:' \
+                " #{error.body}.", error: e
```
igorpeshansky: Shouldn't we return `{}` here?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1711,6 +1747,118 @@ def convert_to_utf8(input)
...
+      partial_errors = error_details.detect(
+        -> { fail JSON::ParserError, "No type #{PARTIAL_ERROR_FIELD}." }
+      ) do |error_detail|
+        ensure_hash(error_detail)['@type'] == PARTIAL_ERROR_FIELD
+      end
+      log_entry_errors = ensure_hash(
+        ensure_hash(partial_errors)['logEntryErrors'])
+      log_entry_errors.each do |index, log_entry_error|
+        error_hash = ensure_hash(log_entry_error)
+        unless error_hash['code'] && error_hash['message']
```
igorpeshansky:
```ruby
fail JSON::ParserError, "Entry #{index} is missing 'code' or 'message'." \
  unless error_hash['code'] && error_hash['message']
```

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Good point. If you only have one function, this works fine.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1711,6 +1747,118 @@ def convert_to_utf8(input)
...
+        error_details_map[error_key] << index
+      end
+      error_details_map
+    rescue JSON::ParserError => e
+      @log.warn 'Failed to extract log entry errors from the error details:' \
+                " #{error.body}.", error: e
+    end
+
+    def ensure_array(value)
+      Array.try_convert(value) || (fail JSON::ParserError, "#{value.class}")
```
igorpeshansky: [Really optional] This could also be:
```ruby
def check(cls, value)
  cls.try_convert(value) || (fail JSON::ParserError, "#{value.class}")
end
```
and then used as:
```ruby
      ...
      error_details = check(Array,
        check(Hash, check(Hash, JSON.parse(error.body))['error'])['details'])
      partial_errors = error_details.detect(
        -> { fail JSON::ParserError, "No type #{PARTIAL_ERROR_FIELD}." }
      ) do |error_detail|
        check(Hash, error_detail)['@type'] == PARTIAL_ERROR_FIELD
      end
      log_entry_errors = check(Hash,
        check(Hash, partial_errors)['logEntryErrors'])
      ...
```

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Hmm... I think this reduces readability in this case where we already have a lot of parentheses. Will keep the `ensure_array` and `ensure_hash` functions instead.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -665,9 +688,22 @@ def write(chunk)
           rescue Google::Apis::ClientError => error
             # Most ClientErrors indicate a problem with the request itself and
             # should not be retried.
-            increment_dropped_entries_count(entries_count)
-            @log.warn "Dropping #{entries_count} log message(s)",
-                      error_class: error.class.to_s, error: error.to_s
+            error_details_map = {}
+            error_details_map = construct_error_details_map(error) if
```
igorpeshansky: This can now become:
```ruby
error_details_map = if @partial_success then construct_error_details_map(error) else {} end
```
Or you could even add `return {} unless @partial_success` at the start of `construct_error_details_map` and make this an unconditional assignment.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: The code I proposed has the exact same number of parentheses -- the only difference is that the `Array` and `Hash` bits are moved inside the parentheses. :-) And another advantage is that something like `check(String, index)` is also possible. But you're probably right that it's a bit too magical, so let's stay with what you have.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1711,6 +1745,118 @@ def convert_to_utf8(input)
...
+    #   }
+    # }
+    #
+    # The ultimate map that is constructed is:
+    # {
+    #   [7, 'User not authorized.']: ['0'],
+    #   [3, 'Log name contains illegal character :']: ['1', '3']
+    # }
+    def construct_error_details_map(error)
+      return {} unless @partial_success
```
igorpeshansky: Let's document this in the header comment too?

PR #170, opened 2017-10-17
Fix common label extraction for Cloud ML and Dataflow.

qingling128:
Turned out that we never had Cloud ML and CLoud Dataflow tests.

The constants were added during the Logging API V2 Migration: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/110/commits/6eebbc8cb5ff5c632b6e6e688f8db32e7ac1a0e6.

Yet there are no actual tests.

The root cause for the issue is that we should extract certain labels from `common metadata labels` to `resource labels` at the group level. (Previously we only did it at the entry level).

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1091,6 +1091,18 @@ module Fluent
         common_labels.delete("#{COMPUTE_CONSTANTS[:service]}/resource_name")
       end
 
+      # Cloud Dataflow and Cloud ML.
+      # These labels can be set via configuring 'labels'.
+      # Report them as monitored resource labels instead of common labels.
+      # e.g. "dataflow.googleapis.com/job_id" => "job_id"
+      [DATAFLOW_CONSTANTS, ML_CONSTANTS].each do |service_constants|
+        next unless resource.type == service_constants[:resource_type]
+        resource.labels.merge!(
+          delete_and_extract_labels(
+            common_labels, service_constants[:extra_common_labels]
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1091,6 +1091,18 @@ def determine_group_level_monitored_resource_and_labels(tag,
         common_labels.delete("#{COMPUTE_CONSTANTS[:service]}/resource_name")
       end
 
+      # Cloud Dataflow and Cloud ML.
+      # These labels can be set via configuring 'labels'.
```
igorpeshansky: s/`configuring 'labels'`/`the 'labels' option`/.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1091,6 +1091,18 @@ def determine_group_level_monitored_resource_and_labels(tag,
...
+      # Report them as monitored resource labels instead of common labels.
+      # e.g. "dataflow.googleapis.com/job_id" => "job_id"
+      [DATAFLOW_CONSTANTS, ML_CONSTANTS].each do |service_constants|
+        next unless resource.type == service_constants[:resource_type]
+        resource.labels.merge!(
+          delete_and_extract_labels(
+            common_labels, service_constants[:extra_common_labels]
```
igorpeshansky: `:extra_resource_labels`?

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

PR #171, opened 2017-10-19
Remove gRPC version pin.

qingling128:
There is an open load balancing bug with `gRPC` at https://github.com/grpc/grpc/issues/12506.

Previously we use the same port for all mocks in our test, which triggers a ton of `Call dropped by load balancing policy` errors.

The short term workaround would be to use a different port for each test run. I've tested this change on a GCP VM. It seems to work fine. All logs are ingested. No errors in the Fluentd log.

The issue is still a bit concerning. Yet at least we can start the developing work for partial success which requires `gRPC` `1.6.6`.

### DIFF
--- a/test/plugin/test_out_google_cloud_grpc.rb
+++ b/test/plugin/test_out_google_cloud_grpc.rb
@@ -212,8 +212,13 @@ class GoogleCloudOutputGRPCTest < Test::Unit::TestCase
     use_grpc true
   )
 
+  @@mock_port = 0 # rubocop:disable Style/ClassVars
+
   def generate_mock_host
-    "localhost:#{Random.new.rand(50_000..60_000)}"
+    # rubocop:disable Style/ClassVars
+    @@mock_port = (@@mock_port + 1) % 10_000 + 50_000
+    "localhost:#{@@mock_port}"
+    # rubocop:enable Style/ClassVars
   end
 
   # Create a Fluentd output test driver with the Google Cloud Output plugin with
@@ -319,6 +324,10 @@ class GoogleCloudOutputGRPCTest < Test::Unit::TestCase
   end
 
   # Set up grpc stubs to mock the external calls.
+  # TODO(qingling128): Remove this comment after grpc/12506 is resolved.
+  # Due to a gRPC load balancing issue (grpc/12506), we have to use a different
+  # port each time we create a gRPC mock as a temporary workaround. Thus we can
+  # only create one driver in each of setup_logging_stubs context.
   def setup_logging_stubs(should_fail = false, code = 0, message = 'Ok')
     # Save the mock host in an instance variable, so later on when creating the
     # logging driver, we can refer to this host with exactly the same port.
--- a/test/plugin/test_out_google_cloud_grpc.rb
+++ b/test/plugin/test_out_google_cloud_grpc.rb
@@ -91,8 +91,8 @@ class GoogleCloudOutputGRPCTest < Test::Unit::TestCase
       [true, 13, 1, 2, [0, 1, 0, 0, 2]]
     ].each do |should_fail, code, request_count, entry_count, metric_values|
       setup_prometheus
-      setup_logging_stubs(should_fail, code, 'SomeMessage') do
-        (1..request_count).each do
+      (1..request_count).each do
+        setup_logging_stubs(should_fail, code, 'SomeMessage') do
           d = create_driver(USE_GRPC_CONFIG + PROMETHEUS_ENABLE_CONFIG, 'test',
                             GRPCLoggingMockFailingService.rpc_stub_class)
           (1..entry_count).each do |i|
@@ -205,8 +205,6 @@ class GoogleCloudOutputGRPCTest < Test::Unit::TestCase
 
   private
 
-  GRPC_MOCK_HOST = 'localhost:56789'
-
   WriteLogEntriesRequest = Google::Logging::V2::WriteLogEntriesRequest
   WriteLogEntriesResponse = Google::Logging::V2::WriteLogEntriesResponse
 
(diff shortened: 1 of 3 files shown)


test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -205,31 +205,41 @@ def test_non_integer_timestamp
...
-
   WriteLogEntriesRequest = Google::Logging::V2::WriteLogEntriesRequest
   WriteLogEntriesResponse = Google::Logging::V2::WriteLogEntriesResponse
 
   USE_GRPC_CONFIG = %(
     use_grpc true
   )
 
+  def generate_mock_host
+    "localhost:#{Random.new.rand(50_000..60_000)}"
```
igorpeshansky: Why random and not sequential?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -205,31 +205,41 @@ def test_non_integer_timestamp
...
   # grpc enabled. The signature of this method is different between the grpc
   # path and the non-grpc path. For grpc, an additional grpc stub class can be
   # passed in to construct the mock used by the test driver.
   def create_driver(conf = APPLICATION_DEFAULT_CONFIG, tag = 'test',
                     grpc_stub = GRPCLoggingMockService.rpc_stub_class)
+    if @mock_host
+      mock_host = @mock_host
+      @mock_host = nil
+    else
+      mock_host = generate_mock_host
```
igorpeshansky: Is this for cases where there are no stubs set up or this is the second driver within the stub context? Wouldn't we want to fail here instead?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -310,6 +320,9 @@ def _undefined
 
   # Set up grpc stubs to mock the external calls.
```
igorpeshansky: Let's document the (hopefully temporary) constraint that a we should create exactly one driver within this context.

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -310,6 +320,9 @@ def _undefined
...
   def setup_logging_stubs(should_fail = false, code = 0, message = 'Ok')
+    # Save the mock host in an instance variable, so later on when creating the
+    # logging driver, we can refer to this host with exactly the same port.
+    @mock_host = generate_mock_host
```
igorpeshansky: We could make it a bit less magic by yielding the host into the block, so we can pass it into `create_driver`, but that would require changing every use of `setup_logging_stubs`, and this is a temporary change, so up to you.
Let's add this as a separate commit, though, so we can revert just that one without reverting the whole PR.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: Some test creates the driver without setting up the logging stub.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: Documented.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: I thought about it, but decided not to make changes to every test for a temporary workaround like this.

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -309,7 +324,14 @@ def _undefined
   end
 
   # Set up grpc stubs to mock the external calls.
+  # TODO(qingling128): Remove this comment after grpc/12506 is resolved.
+  # Due to a gRPC load balancing issue (grpc/12506), we have to use a different
+  # port each time we create a gRPC mock as a temporary workaround. Thus we can
+  # only create one driver in each of setup_logging_stubs context.
```
igorpeshansky: s/`each of`/`each`/

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -309,7 +324,14 @@ def _undefined
   end
 
   # Set up grpc stubs to mock the external calls.
+  # TODO(qingling128): Remove this comment after grpc/12506 is resolved.
```
igorpeshansky: Let's reference this properly, i.e., s/`grpc/12506`/`grpc/grpc#12506`/. Also below.

qingling128: Changed.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: Changed.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: I added some comment to clarify this.

`@mock_host` only gets set up when `setup_logging_stubs` is called. Some tests create drivers without creating a logging stub context. For these cases just use a new mock port. 

Another option is to change these test cases. Yet this is a temporary workaround. I'd like to keep the size of the things we need to revert as small as possible.

igorpeshansky: I'm confused — how do the tests succeed if the expectations are not set up? Does that mean that the expectations are irrelevant, because the tests never invoke the driver? Then there should be no need to set up the driver — it might as well be `nil`. What am I missing?

qingling128: For tests as simple as

```
  def test_configure_service_account_application_default
    setup_gce_metadata_stubs
    d = create_driver
    assert_equal HOSTNAME, d.instance.vm_name
  end
```

We never reach the point where a RPC call needs to be made. So logging stubs is never set up.

On a second thought, as long as we don't actually call the api_client in the driver, we don't require a mock host. So I just removed the logic to create a mock host for the driver when no logging stub was set up. It should be fine for it to stay `nil`.

PTAL.

igorpeshansky: My point exactly.

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -356,7 +347,6 @@ def setup_logging_stubs(should_fail = false, code = 0, message = 'Ok')
     end
     srv.stop
     t.join
-    @mock_host = nil
```
igorpeshansky: We probably want to leave this here, so that `@mock_host` from the previous test doesn't leak into subsequent ones.

qingling128: Since we no longer check whether `@mock_host` is set before using it, we should not need to worry about setting it to `nil`, right? Each logging stub will use a new port anyway. We just need to make sure no more than 1 driver gets created in one logging stub context.

igorpeshansky: If a driver gets created outside of the logging stubs context, we don't want it to reuse the port of the previous test, which is what would happen if we don't reset `@mock_host`, unless I'm missing something.
The second driver within the same stub context will fail during creation, because it'll try to bind to the same port, so I wouldn't worry about that.

qingling128: For a driver that gets created outside of the logging stubs context, we don't expect it to call the `api_client` method to use the port. If it does, it will fail, whether `mock_host` is `nil` or the port of the previous test. Maybe it's better to fail with `nil`. I just added the `@mock_host = nil` back.

PR #176, opened 2017-11-09
Make Stackdriver Logging API endpoint configurable.

qingling128:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -321,6 +323,9 @@ module InternalConstants
                  :default => nil,
                  :secret => true
 
+    # The URL of Stackdriver Logging API.
```
igorpeshansky: This theoretically appears in some generated parameter description somewhere. Add something like "`Right now this only works with the gRPC path (use_grpc = true).`"?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -369,6 +374,12 @@ def configure(conf)
                   ' partial_success flag will be ignored.'
       end
 
+      unless @logging_api_url == DEFAULT_LOGGING_API_URL || @use_grpc
```
igorpeshansky: Add a TODO, along the lines of:
```
      # TODO(qingling128): Remove this warning after the support is added. Also
      # remove the comment in the description of this configuration.
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -369,6 +374,12 @@ def configure(conf)
...
+        @log.warn 'Detected customized logging_api_url while use_grpc is not' \
+                  ' enabled. Customized logging_api_url for the non-gRPC path' \
+                  ' is not supported. The config will be ignored.'
```
igorpeshansky: `The logging_api_url option will be ignored.`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1718,7 +1729,7 @@ def api_client
         creds = GRPC::Core::CallCredentials.new(authentication.updater_proc)
         creds = ssl_creds.compose(creds)
         @client = Google::Logging::V2::LoggingServiceV2::Stub.new(
-          'logging.googleapis.com', creds)
+          @logging_api_url, creds)
```
igorpeshansky: Technically, in the case of an RPC stub, it's a _host_ (a standalone host name), not a _URL_ (`protocol://host:port/path`). In fact, we would need to reconcile the need to specify the host on the gRPC path and the full URL on the non-gRPC path.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Maybe we can let them always specify the full URL like `https://logging.googleapis.com`. Then for the gRPC path, we throw away the `protocal` part?

WDYT?

qingling128:
Friendly ping.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: That could work. Will gRPC take a `host:port` spec? Also, what happens if the URL has a path? We should probably specify what the expected format is.

qingling128:
Rebased and added a protocal section to the url. Also added some example of what the url should be. Yes, `gRPC` takes `port` and `path` just fine.

I also added a flag `use_secure_channel` to skip authentication for Logging API mocks.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -323,6 +325,18 @@ module InternalConstants
...
+    # gRPC path (use_grpc = true).
+    # The default value is: https://logging.googleapis.com.
+    # One common usage of this config is to provide a mocked / stubbed Logging
+    # API. e.g. http://localhost:52000.
+    config_param :logging_api_url, :string, :default => DEFAULT_LOGGING_API_URL
+
+    # Whether to use a secure channel when talking to the Logging API. This
+    # should only be false if the Stackdriver Logging Agent is set up to talk
+    # to a mocked / stubbed Logging API. Otherwise the requests will fail.
+    config_param :use_secure_channel, :bool, :default => true
```
igorpeshansky: Isn't this equivalent to having the `https://` protocol on the `logging_api_url` (as opposed to the `http://` protocol)?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -226,21 +226,29 @@ def test_non_integer_timestamp
...
-      authentication = Google::Auth.get_application_default
-      creds = GRPC::Core::CallCredentials.new(authentication.updater_proc)
-      ssl_creds.compose(creds)
+      if @use_secure_channel
+        ssl_creds = GRPC::Core::ChannelCredentials.new
+        authentication = Google::Auth.get_application_default
+        creds = GRPC::Core::CallCredentials.new(authentication.updater_proc)
+        ssl_creds.compose(creds)
+      else
+        :this_channel_is_insecure
```
igorpeshansky: Shouldn't you be assigning this to `creds`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -323,6 +325,18 @@ module InternalConstants
                  :default => nil,
                  :secret => true
 
+    # The URL of Stackdriver Logging API. Right now this only works with the
+    # gRPC path (use_grpc = true).
+    # The default value is: https://logging.googleapis.com.
```
igorpeshansky: This sentence is redundant, as it's obvious from the `:default` item below.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -323,6 +325,18 @@ module InternalConstants
...
+    # One common usage of this config is to provide a mocked / stubbed Logging
+    # API. e.g. http://localhost:52000.
```
igorpeshansky: Nit: comma after `API` and after `e.g.`.

igorpeshansky: Careful -- your current code won't work with the `http://` prefix (it should, though).
Also, just wondering: do we really want to encode a port here?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1768,13 +1791,17 @@ def init_api_client
...
+          authentication = Google::Auth.get_application_default
+          creds = GRPC::Core::CallCredentials.new(authentication.updater_proc)
+          creds = ssl_creds.compose(creds)
+        else
+          creds = :this_channel_is_insecure
+        end
         @client = Google::Cloud::Logging::V2::LoggingServiceV2Client.new(
           channel: GRPC::Core::Channel.new(
-            'logging.googleapis.com', nil, creds))
+            @logging_api_url.sub(%r{^https?\://}, ''), nil, creds))
```
igorpeshansky: Would it make sense to use `URI.parse` instead of string substitution?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1768,13 +1791,17 @@ def init_api_client
...
-        creds = ssl_creds.compose(creds)
+        if @use_secure_channel
+          ssl_creds = GRPC::Core::ChannelCredentials.new
+          authentication = Google::Auth.get_application_default
+          creds = GRPC::Core::CallCredentials.new(authentication.updater_proc)
+          creds = ssl_creds.compose(creds)
+        else
+          creds = :this_channel_is_insecure
+        end
         @client = Google::Cloud::Logging::V2::LoggingServiceV2Client.new(
```
igorpeshansky: Does this need to be rebased again?

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: What I mean is, should we use the `http://` vs. `https://` prefix to infer this value, rather than adding a separate config option?

test/plugin/test_out_google_cloud_grpc.rb
igorpeshansky: I'm confused. The value by itself is useless, isn't it? What is this method even doing? Isn't it a no-op?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1866,6 +1891,18 @@ def api_client
       @client
     end
 
+    def grpc_channel_name
+      uri = URI.parse(@logging_api_url)
+      host = uri.host
+      unless host
+        raise Fluent::ConfigError,
+              'The URL specified by config logging_api_url is invalid:' \
```
igorpeshansky: `The logging_api_url option specifies an invalid URL: #{@logging_api_url}`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1866,6 +1891,18 @@ def api_client
...
+              " #{logging_api_url}."
```
igorpeshansky: Are you missing an `@` here?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -228,29 +228,28 @@ def create_driver(conf = APPLICATION_DEFAULT_CONFIG,
                     tag = 'test',
                     multi_tags = false)
     conf += USE_GRPC_CONFIG
+    use_secure_channel = !conf.include?('use_secure_channel false')
```
igorpeshansky: Isn't this stale now?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1866,6 +1886,22 @@ def api_client
       @client
     end
 
+    def use_secure_channel
```
igorpeshansky: I think having this as a separate method detracts from readability. Would rather see this test inline.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1866,6 +1886,22 @@ def api_client
...
+      @logging_api_url.start_with?('https')
```
igorpeshansky: Shouldn't this be `starts_with?('https://')`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1866,6 +1886,22 @@ def api_client
...
+    end
+
+    def grpc_channel_name
```
igorpeshansky: Probably makes sense to inline this method into `init_api_client` as well, allowing you to use `uri.scheme` instead of the `starts_with` string test.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Inlined.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

test/plugin/test_out_google_cloud_grpc.rb
qingling128: Right. Removed this and all other occurrences.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -325,6 +327,12 @@ module InternalConstants
                  :default => nil,
                  :secret => true
 
+    # The URL of Stackdriver Logging API. Right now this only works with the
+    # gRPC path (use_grpc = true).
+    # One common usage of this config is to provide a mocked / stubbed Logging
+    # API, e.g., http://localhost:52000.
```
igorpeshansky: Probably worth mentioning that using `http://` will create an insecure channel.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -325,6 +327,12 @@ module InternalConstants
                  :default => nil,
                  :secret => true
 
+    # The URL of Stackdriver Logging API. Right now this only works with the
+    # gRPC path (use_grpc = true). An unsecured channel is used if the url
```
igorpeshansky: Nit: `URL`.

qingling128:
PTAL

PR #179, opened 2017-11-21
Request the logging scope from signet when fetching an access token.

pkern:
The plugin already sets the logging scope to be requested for newly
acquired access token. However signet ignores this unless it obtains the
refresh token itself. google-fluentd usually uses a service account to
do so. If someone passes in an already created refresh token using the
GOOGLE_REFRESH_TOKEN environment variable, the access token request will
not request the logging scope. This fixes the latter.

More context about the change in signet (0.8+) can be found in
https://github.com/google/signet/pull/89

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1714,8 +1714,11 @@ module Fluent
     def api_client
       if @use_grpc
         ssl_creds = GRPC::Core::ChannelCredentials.new
-        authentication = Google::Auth.get_application_default
-        creds = GRPC::Core::CallCredentials.new(authentication.updater_proc)
+        authentication = Google::Auth.get_application_default(LOGGING_SCOPE)
+        updater_proc = lambda do |a_hash|
+          authentication.apply(a_hash, use_configured_scope: true)
+        end
+        creds = GRPC::Core::CallCredentials.new(updater_proc)
         creds = ssl_creds.compose(creds)
         @client = Google::Logging::V2::LoggingServiceV2::Stub.new(
           'logging.googleapis.com', creds)
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1722,7 +1722,8 @@ module Fluent
       else
         unless @client.authorization.expired?
           begin
-            @client.authorization.fetch_access_token!
+            @client.authorization.fetch_access_token!(
+              use_configured_scope: true)
           rescue MultiJson::ParseError
             # Workaround an issue in the API client; just re-raise a more
             # descriptive error for the user (which will still cause a retry).
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1715,10 +1715,9 @@ module Fluent
       if @use_grpc
         ssl_creds = GRPC::Core::ChannelCredentials.new
         authentication = Google::Auth.get_application_default(LOGGING_SCOPE)
-        updater_proc = lambda do |a_hash|
+        creds = GRPC::Core::CallCredentials.new(lambda do |a_hash|
           authentication.apply(a_hash, use_configured_scope: true)
-        end
-        creds = GRPC::Core::CallCredentials.new(updater_proc)
+        end)
         creds = ssl_creds.compose(creds)
         @client = Google::Logging::V2::LoggingServiceV2::Stub.new(
           'logging.googleapis.com', creds)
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1722,7 +1722,7 @@ def api_client
       else
         unless @client.authorization.expired?
           begin
-            @client.authorization.fetch_access_token!
+            @client.authorization.fetch_access_token!(:use_configured_scope => true)
```
igorpeshansky: Have you tested with the gRPC path? Do you need an equivalent change there (lines 1717-1718)?

igorpeshansky: Also, please fix the `rubocop` errors.

pkern: Where would I need to put the use_grpc = true to test with it? (Assuming a regular google-fluentd package.)

pkern: `rubocop` now passes.

igorpeshansky: You can add that option (`use_grpc true`, no `=`) in the `google_cloud` output plugin stanza.

igorpeshansky:
Also, you don't have to force-push and squash commits during the review — we'll squash-and-merge in the end anyway.

lib/fluent/plugin/out_google_cloud.rb
pkern: So with that it doesn't work again. That provisions the updater function provided by googleauth. Unfortunately the one it monkey patches into Signed::OAuth2::Client does not take any options. I pushed a workaround in this library as it doesn't seem to be too odd to do the request mangling here. I guess rubocop will tell me if the coding style matches.

pkern:
Please review the gRPC carefully. I think something is still off but I'd assume it's not my change (and our setup is pretty peculiar). For the original change I have been more confident that it's actually ok as-is.

(Right now in my setup it submits successfully with gRPC and then hangs. OTOH we also don't really use gRPC so I have no idea if this is due to auth or something else.)

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1714,15 +1714,19 @@ def init_api_client
     def api_client
       if @use_grpc
         ssl_creds = GRPC::Core::ChannelCredentials.new
-        authentication = Google::Auth.get_application_default
-        creds = GRPC::Core::CallCredentials.new(authentication.updater_proc)
+        authentication = Google::Auth.get_application_default(LOGGING_SCOPE)
+        updater_proc = lambda do |a_hash|
+          authentication.apply(a_hash, use_configured_scope: true)
+        end
+        creds = GRPC::Core::CallCredentials.new(updater_proc)
```
igorpeshansky: You could inline the `lambda` expression directly into the `GRPC::Core::CallCredentials.new` call, i.e.,
```ruby
        creds = GRPC::Core::CallCredentials.new(lambda do |a_hash|
          authentication.apply(a_hash, use_configured_scope: true)
        end)

File: lib/fluent/plugin/out_google_cloud.rb [hunk 3]
```diff
@@ -1714,15 +1714,19 @@ def init_api_client
     def api_client
       if @use_grpc
         ssl_creds = GRPC::Core::ChannelCredentials.new
-        authentication = Google::Auth.get_application_default
-        creds = GRPC::Core::CallCredentials.new(authentication.updater_proc)
+        authentication = Google::Auth.get_application_default(LOGGING_SCOPE)
```
igorpeshansky: I'll let @qingling128 comment on whether we need an equivalent change to the tests ([`GoogleCloudOutputGRPCTest::GoogleCloudOutputWithGRPCMock.api_client`](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/master/test/plugin/test_out_google_cloud_grpc.rb#L246)).

lib/fluent/plugin/out_google_cloud.rb
pkern: Done.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Since we have added the scope in the plugin, we should change it accordingly in `GoogleCloudOutputGRPCTest::GoogleCloudOutputWithGRPCMock.api_client` as well. 

The way credentials are constructed in the mock should be identical to:
```
        ssl_creds = GRPC::Core::ChannelCredentials.new
        authentication = Google::Auth.get_application_default(LOGGING_SCOPE)
        creds = GRPC::Core::CallCredentials.new(lambda do |a_hash|
          authentication.apply(a_hash, use_configured_scope: true)
        end)
        creds = ssl_creds.compose(creds)
```

pkern:
PTAL. One of the Travis builds flaked out with an OSError that the address it tries to listen on is already in use and I can't retry the build. I don't think it's the fault of the updated test.

qingling128:
I restarted Travis and the following build passed.

LGTM.

PR #180, opened 2017-11-22
Add spanId field support, bump google-api-client version to 0.17 and google-cloud-logging version to 1.3.

draffensperger:
This will enable logging agent users to associate log entries with particular trace spans.

This requires an update of `google-api-client` and `google-cloud-logging` to support the new spanId field for the REST and gRPC call paths (See https://github.com/google/google-api-ruby-client/commit/35a23f126cde4723059b57954e90db4cf50ce7e9 and https://github.com/GoogleCloudPlatform/google-cloud-ruby/commit/48464a68fcdcc68fdd57a4872be2c02588b789dd).

### DIFF
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -1088,38 +1088,39 @@ module BaseTest
     end
   end
 
-  def test_log_entry_trace_field
+  def verify_log_entry_field_key(log_entry_field, default_key, custom_key,
+                                 custom_key_config)
     setup_gce_metadata_stubs
     message = log_entry(0)
-    trace = 'projects/project-1/traces/1234567890abcdef1234567890abcdef'
+    sample_value = "sample #{log_entry_field} value"
     [
       {
-        # It leaves trace entry field nil if no trace value sent.
+        # It leaves log entry field nil if no keyed value sent.
         driver_config: APPLICATION_DEFAULT_CONFIG,
         emitted_log: { 'msg' => message },
-        expected_fields: { 'msg' => message },
-        expected_trace_value: nil
+        expected_payload: { 'msg' => message },
+        expected_field_value: nil
       },
       {
-        # By default, it sets trace via Google-specific key.
+        # By default, it sets log entry field via a default key.
         driver_config: APPLICATION_DEFAULT_CONFIG,
-        emitted_log: { 'msg' => message, DEFAULT_TRACE_KEY => trace },
-        expected_fields: { 'msg' => message },
-        expected_trace_value: trace
+        emitted_log: { 'msg' => message, default_key => sample_value },
+        expected_payload: { 'msg' => message },
+        expected_field_value: sample_value
       },
       {
-        # It allows setting the trace via a custom configured key.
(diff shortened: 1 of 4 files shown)


test/plugin/base_test.rb
```diff
@@ -1088,38 +1088,39 @@ def test_http_request_skip_setting_latency
     end
   end
 
-  def test_log_entry_trace_field
+  def verify_log_entry_field_key(log_entry_field, default_key, custom_key,
+                                 custom_key_config)
     setup_gce_metadata_stubs
     message = log_entry(0)
-    trace = 'projects/project-1/traces/1234567890abcdef1234567890abcdef'
+    sample_value = "sample #{log_entry_field} value"
```
igorpeshansky: Seems like we should be doing this kind of test for all of the other fields (instead of `verify_subfields_from_record`, `verify_subfields_partial_from_record`, etc). Can we make the sample value a parameter as well, so we can eventualy apply this to `operation_key`, `source_location_key`, and `http_request_key`?

test/plugin/base_test.rb
```diff
@@ -1088,38 +1088,39 @@ def test_http_request_skip_setting_latency
     end
   end
 
-  def test_log_entry_trace_field
+  def verify_log_entry_field_key(log_entry_field, default_key, custom_key,
```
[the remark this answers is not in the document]
draffensperger: Done.

test/plugin/base_test.rb
draffensperger: Makes sense - I moved `sample_value` to be an argument to the method.

igorpeshansky:
@draffensperger It's unfortunate that you've done both the move and the extraction of the `sample_value` parameter in the same commit -- that makes it hard to see what's changed. That's why I asked for the separate commit that just moves the test...
Is there a chance you can split that commit into two?

draffensperger:
@igorpeshansky I changed the function move and making the sample value a new parameter into two separate commits.

Sorry - I misunderstood your comment as asking that I be sure to make a new commit rather than doing a squash onto my initial commit.

test/plugin/base_test.rb
```diff
@@ -1089,15 +1089,15 @@ def test_http_request_skip_setting_latency
   end
 
   def test_log_entry_trace_field
-    verify_log_entry_field_key('trace', DEFAULT_TRACE_KEY,
-                               'custom_trace_key',
-                               CONFIG_CUSTOM_TRACE_KEY_SPECIFIED)
+    sample_value = 'projects/project-1/traces/1234567890abcdef1234567890abcdef'
+    verify_field_key('trace', DEFAULT_TRACE_KEY, 'custom_trace_key',
+                     CONFIG_CUSTOM_TRACE_KEY_SPECIFIED, sample_value)
```
igorpeshansky: [Optional] I'd just inline the sample value into the call...

draffensperger: Done here and for span ID also.

draffensperger:
@igorpeshansky and @qingling128 if this is good to go, could one of you merge it in?

Issue #182, opened 2017-11-24
Detecting the platform based on METADATA_SERVICE_ADDR API response header is not working on some AWS instancetype


baskaran-md:
**Error:**
```
2017-11-24 21:27:40 +0000 [info]: plugin/out_google_cloud.rb:737:detect_platform: Unable to determine platform
2017-11-24 21:27:40 +0000 [error]: fluent/supervisor.rb:540:rescue in dry_run: config error file="/opt/apigee/data/apigee-logger/google-fluentd.conf" error_class=Fluent::ConfigError error="Unable to obtain metadata parameters: zone vm_id"
```

Some of the new instance families like **c5.9xlarge**, doesn't return the same response headers like the other instance types. 

```
[root@node baskaran.md]# curl -s http://169.254.169.254 -I
HTTP/1.1 200 OK
Connection: close
Content-Type: text/plain
```

However, on other instances like **c3.large,** 
```
[root@node baskaran.md]# curl http://169.254.169.254/ -I
HTTP/1.0 200 OK
Content-Type: text/plain
Accept-Ranges: bytes
ETag: "2990503662"
Last-Modified: Thu, 09 Nov 2017 04:08:37 GMT
Content-Length: 197
Connection: close
Date: Fri, 24 Nov 2017 04:05:13 GMT
Server: EC2ws
```

So relying on the `Server` response header to identify the platform is not foolproof. Instead, can we fallback to another check like below?
**Ref:** http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-data-categories

```
[root@node baskaran.md]# curl http://169.254.169.254/latest/meta-data/services/domain -i
HTTP/1.1 200 OK
Connection: close
Content-Type: text/plain

amazonaws.com
```

Meanwhile, I am following up with Amazon Support to understand more on the response headers of those APIs.

igorpeshansky:
Thank you, please keep us posted on what you find out from AWS support.

baskaran-md:
AWS Support was able to reproduce the issue, and they have assigned the ticket to internal team to investigate more. 

I am also checking with them on the reliability of `services/domain` resource. Will post back their response here.

baskaran-md:
Amazon has deployed a fix on their server side to include the default headers in the response of metadata.

[...]

PR #183, opened 2017-11-24
Check the domain in METADATA_SERVICE_ADDR to confirm on EC2 platform

baskaran-md:
Fixes https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/182

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -768,6 +768,13 @@ module Fluent
             return Platform::EC2
           end
         end
+        open('http://' + METADATA_SERVICE_ADDR +
+                '/latest/meta-data/services/domain') do |f|
+          if f.read == 'amazonaws.com'
+            @log.info 'Detected EC2 platform'
+            return Platform::EC2
+          end
+        end
       rescue StandardError => e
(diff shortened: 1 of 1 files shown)


baskaran-md:
I signed it!

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -768,6 +768,13 @@ def detect_platform
             return Platform::EC2
           end
         end
+        open('http://' + METADATA_SERVICE_ADDR +
+                '/latest/meta-data/services/domain') do |f|
```
igorpeshansky: Should we just be checking that the `/latest` path doesn't return a 404?

baskaran-md: Checking just the `/latest` seems so generic to confirm the EC2 platform.
Even Google / Azure can add that path in future.

baskaran-md:
@igorpeshansky  Yes, the "Server" response header is not present in the latest instance type. 
More details @ https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/182

baskaran-md:
Closing the request as per https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/182#issuecomment-354341343

PR #184, opened 2017-11-28
Automatically adjust timestamps from the future.

qingling128:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1344,48 @@ def compute_timestamp(resource_type, record, time)
...
+                end
+      ts_nanos = begin
+                   Integer ts_nanos
+                 rescue ArgumentError, TypeError
+                   ts_nanos
+                 end
+
+      # Adjust timestamps from the future.
+      if ts_secs.is_a?(Integer) && ts_nanos.is_a?(Integer)
+        parsed_timestamp = Time.at(ts_secs + ts_nanos / 1_000_000_000.0)
```
igorpeshansky: This is already computed as `timestamp` in many of the cases above. Would it make sense to assign `timestamp` in the remaining cases based on `ts_secs` and `ts_nanos` and just use `timestamp` here?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1344,48 @@ def compute_timestamp(resource_type, record, time)
...
+        end_of_year = Time.utc(Time.now.year, 12, 31, 23, 59, 59.999)
```
igorpeshansky: Careful of timezones. Is the timestamp you're comparing with local or UTC? If the former, shouldn't you be computing the local end of the year?

igorpeshansky: How about:
```ruby
next_year = Time.utc(Time.now.year)
...
if parsed_timestamp >= next_year
  ...
end
```
?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1344,48 @@ def compute_timestamp(resource_type, record, time)
...
+        if parsed_timestamp > end_of_year
+          # This isn't the issue with system logs missing a year, because we are
+          # further than the current year. It is unlikely that users wanted logs
+          # more than a year in the future, This could be logs from a VM just
+          # starting up and the clock hasn't synced, or there was no reasonable
+          # time in the payload but we parsed some random number. Leave time as
+          # the default value so downstream backend will handle it.
+          ts_secs = 0
+          ts_nanos = 0
+        elsif parsed_timestamp > one_day_later
```
igorpeshansky: `>=`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1344,48 @@ def compute_timestamp(resource_type, record, time)
...
+                rescue ArgumentError, TypeError
+                  ts_secs
+                end
+      ts_nanos = begin
+                   Integer ts_nanos
+                 rescue ArgumentError, TypeError
+                   ts_nanos
+                 end
+
+      # Adjust timestamps from the future.
```
igorpeshansky: I would describe the heuristics here, rather than making the reader filter for comments in the code below.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1344,48 @@ def compute_timestamp(resource_type, record, time)
...
+      if ts_secs.is_a?(Integer) && ts_nanos.is_a?(Integer)
+        parsed_timestamp = Time.at(ts_secs + ts_nanos / 1_000_000_000.0)
+        end_of_year = Time.utc(Time.now.year, 12, 31, 23, 59, 59.999)
+        one_day_later = Time.now + 60 * 60 * 24
+        if parsed_timestamp > end_of_year
+          # This isn't the issue with system logs missing a year, because we are
+          # further than the current year. It is unlikely that users wanted logs
+          # more than a year in the future, This could be logs from a VM just
```
igorpeshansky: Did you mean s/`,`/`.`/?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1344,48 @@ def compute_timestamp(resource_type, record, time)
...
+          # starting up and the clock hasn't synced, or there was no reasonable
+          # time in the payload but we parsed some random number. Leave time as
+          # the default value so downstream backend will handle it.
```
igorpeshansky: Nit: `the downstream backend`. In fact, I'd say: `Reset time to the default value and let the downstream API handle it.`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1344,48 @@ def compute_timestamp(resource_type, record, time)
...
+          ts_secs = 0
+          ts_nanos = 0
+        elsif parsed_timestamp > one_day_later
+          # This is the current year, but in the future. Most likely this is
+          # from some previous year. We assume it is last year. This could
+          # incorrectly label 2 year old logs, but they probably aren't super
```
igorpeshansky: Nit: `2-year-old`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1344,48 @@ def compute_timestamp(resource_type, record, time)
...
+          # important and implies that fluentd is broken and resending unneeded
```
igorpeshansky: > implies that fluentd is broken and resending unneeded logs.

Not clear. Clients could be ingesting old unrotated logs for the first time. There's no way for fluentd to know that these are old. I would remove this phrase.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1344,48 @@ def compute_timestamp(resource_type, record, time)
...
+          # logs. Adjust the timestamp to 1 year ago from the parsed value.
```
igorpeshansky: Since you already know that the timestamp is in this year: `Adjust the timestamp to last year.`

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1344,48 @@ def compute_timestamp(resource_type, record, time)
...
+          adjusted_timestamp = Time.mktime(
```
igorpeshansky: I can't help but think we should be able to use the [`DateTime`](https://ruby-doc.org/stdlib-2.2.1/libdoc/date/rdoc/DateTime.html) class, e.g.:
```ruby
adjusted_timestamp = parsed_timestamp.to_datetime.prev_year.to_time
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1344,48 @@ def compute_timestamp(resource_type, record, time)
...
+                   Integer ts_nanos
+                 rescue ArgumentError, TypeError
+                   ts_nanos
+                 end
+
+      # Adjust timestamps from the future.
+      if ts_secs.is_a?(Integer) && ts_nanos.is_a?(Integer)
+        parsed_timestamp = Time.at(ts_secs + ts_nanos / 1_000_000_000.0)
+        end_of_year = Time.utc(Time.now.year, 12, 31, 23, 59, 59.999)
+        one_day_later = Time.now + 60 * 60 * 24
```
igorpeshansky: I can't help but think we should be able to use the [`DateTime`](https://ruby-doc.org/stdlib-2.2.1/libdoc/date/rdoc/DateTime.html) class, e.g.:
```ruby
one_day_later = Time.now.to_datetime.next_day.to_time
```

igorpeshansky: Be careful with precision -- calling `Time.now` multiple times will return subtly different results, which could wreak havoc when close to the new year boundary. I'd suggest calling `Time.now` once at the start of `compute_timestamp` and using the same value throughout.

test/plugin/base_test.rb
```diff
@@ -607,32 +607,43 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
 
   def test_timestamps
     setup_gce_metadata_stubs
-    [Time.at(123_456.789), Time.at(0), Time.now].each do |ts|
+    current_time = Time.now
+    current_year = current_time.year
+    future_before_end_of_year = Time.utc(current_year, 12, 31, 23, 59, 59)
```
igorpeshansky: Same here -- `DateTime` is likely your friend...

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: You can explicitly set `timestamp` to `nil` in those cases, can't you, and simply skip the adjustment?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1313,6 +1302,7 @@ def cluster_name_from_kube_env(kube_env)
     end
 
     def compute_timestamp(resource_type, record, time)
+      current_time = Time.now
```
igorpeshansky: [Optional] `now = Time.now`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1345,44 @@ def compute_timestamp(resource_type, record, time)
...
+                end
+      ts_nanos = begin
+                   Integer ts_nanos
+                 rescue ArgumentError, TypeError
+                   ts_nanos
+                 end
+
+      # Adjust timestamps from the future.
+      #
+      # 1. If the parsed timestamp is later than the end of the year.
```
igorpeshansky: Nit: this isn't a complete sentence. Besides, you only hint at the use case for this adjustment.
It would also be easier to read if the cases were presented in the reverse order.
How about something like:
```ruby
      # Adjust timestamps from the future.
      # There are two cases:
      # 1. The parsed timestamp later in the current year:
      # This can happen when system log lines from previous years are missing
      # the year, so the date parser assumes the current year.
      # We treat these lines as coming from last year.  This could label
      # 2-year-old logs incorrectly, but this probably isn't super important.
      #
      # 2. The parsed timestamp is past the end of the current year:
      # Since the year is different from the current year, this isn't the
      # missing year in system logs.  It is unlikely that users explicitly
      # write logs at a future date.  This could result from an unsynchronized
      # clock on a VM, or some random value being parsed as the timestamp.
      # We reset the timestamp on those lines to the default value and let the
      # downstream API handle it.
```
?

test/plugin/base_test.rb
```diff
@@ -607,32 +607,44 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+    one_second_before_next_year = next_year - 1
+    adjusted_to_last_year =
+      one_second_before_next_year.to_datetime.prev_year.to_time
+    one_day_into_next_year = next_year.to_date.next_day.to_time
+    {
+      Time.at(123_456.789) => Time.at(123_456.789),
+      Time.at(0) => Time.at(0),
+      current_time => current_time,
+      one_second_before_next_year => adjusted_to_last_year,
+      one_day_into_next_year => Time.at(0)
```
igorpeshansky: Also `one_second_into_the_next_year` and `next_year`?

lib/fluent/plugin/out_google_cloud.rb
qingling128: Ah, in fact I was trying to avoid having the variable name the same as a method name.

lib/fluent/plugin/out_google_cloud.rb

test/plugin/base_test.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Yeah, that works. Changed.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Why? The method is in a different class, and you're storing its return value in the variable...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1312,18 +1301,27 @@ def cluster_name_from_kube_env(kube_env)
       instance_prefix
     end
 
+    def time_from_seconds_and_nanos(ts_secs, ts_nanos)
```
igorpeshansky: `time_or_nil`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1336,11 +1334,13 @@ def compute_timestamp(resource_type, record, time)
...
+        timestamp = time_from_seconds_and_nanos(ts_secs, ts_nanos)
       elsif resource_type == CLOUDFUNCTIONS_CONSTANTS[:resource_type] &&
             @cloudfunctions_log_match
-        timestamp = DateTime.parse(@cloudfunctions_log_match['timestamp'])
-        ts_secs = timestamp.strftime('%s').to_i
-        ts_nanos = timestamp.strftime('%N').to_i
+        date_time = DateTime.parse(@cloudfunctions_log_match['timestamp'])
+        ts_secs = date_time.strftime('%s').to_i
+        ts_nanos = date_time.strftime('%N').to_i
+        timestamp = time_from_seconds_and_nanos(ts_secs, ts_nanos)
```
igorpeshansky: This one could actually be:
```ruby
       timestamp = DateTime.parse(@cloudfunctions_log_match['timestamp']).to_time
       ts_secs = timestamp.tv_sec
       ts_nanos = timestamp.tv_nsec
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1355,44 @@ def compute_timestamp(resource_type, record, time)
...
+                end
+      ts_nanos = begin
+                   Integer ts_nanos
+                 rescue ArgumentError, TypeError
+                   ts_nanos
+                 end
+
+      # Adjust timestamps from the future.
+      # There are two cases:
+      # 1. The parsed timestamp later in the current year:
```
igorpeshansky: My fault. Should have been `timestamp is later`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1355,44 @@ def compute_timestamp(resource_type, record, time)
...
+      if timestamp
+        next_year = Time.mktime(current_time.year + 1)
+        one_day_later = current_time.to_datetime.next_day.to_time
+        if timestamp >= next_year
+          ts_secs = 0
+          ts_nanos = 0
+        elsif timestamp >= one_day_later
+          adjusted_timestamp = timestamp.to_datetime.prev_year.to_time
+          ts_secs = adjusted_timestamp.tv_sec
+          ts_nanos = adjusted_timestamp.tv_nsec
```
igorpeshansky: [Optional] The value of `ts_nanos` shouldn't change when you subtract a year, but up to you.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1355,44 @@ def compute_timestamp(resource_type, record, time)
...
+      # Since the year is different from the current year, this isn't the
+      # missing year in system logs.  It is unlikely that users explicitly
+      # write logs at a future date.  This could result from an unsynchronized
+      # clock on a VM, or some random value being parsed as the timestamp.
+      # We reset the timestamp on those lines to the default value and let the
+      # downstream API handle it.
+      if timestamp
+        next_year = Time.mktime(current_time.year + 1)
+        one_day_later = current_time.to_datetime.next_day.to_time
+        if timestamp >= next_year
```
igorpeshansky: Let's annotate this as "case 2" from above, and the `elsif` branch as "case 1". Line comments are fine.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1312,18 +1301,27 @@ def cluster_name_from_kube_env(kube_env)
       instance_prefix
     end
 
+    def time_from_seconds_and_nanos(ts_secs, ts_nanos)
+      Time.at((Integer ts_secs) + (Integer ts_nanos) / 1_000_000_000.0)
+    rescue
```
igorpeshansky: `rescue ArgumentError, TypeError`

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1312,18 +1301,27 @@ def cluster_name_from_kube_env(kube_env)
       instance_prefix
     end
 
+    def time_from_seconds_and_nanos(ts_secs, ts_nanos)
+      Time.at((Integer ts_secs) + (Integer ts_nanos) / 1_000_000_000.0)
```
igorpeshansky: This converts nanoseconds to a double, which [loses precision](https://ruby-doc.org/core-2.2.0/Time.html#method-i-nsec). You could instead use the 2-argument version:
```ruby
     Time.at((Integer ts_secs), (Integer ts_nanos) / 1000.0)
```

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Right. Removed this line.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Because... it feels like we are calling the method again?

igorpeshansky: Ok, makes sense.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1355,43 @@ def compute_timestamp(resource_type, record, time)
...
+      # clock on a VM, or some random value being parsed as the timestamp.
+      # We reset the timestamp on those lines to the default value and let the
+      # downstream API handle it.
+      if timestamp
+        next_year = Time.mktime(current_time.year + 1)
+        one_day_later = current_time.to_datetime.next_day.to_time
+        if timestamp >= next_year # Case 2.
+          ts_secs = 0
+          ts_nanos = 0
+        elsif timestamp >= one_day_later # Case 1
```
igorpeshansky: Nit: missing trailing '.'.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: [Optional] Think we should also leave a comment explaining why this line isn't needed?

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Added a comment.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1355,6 +1355,44 @@ def compute_timestamp(resource_type, record, time)
...
+      # downstream API handle it.
+      if timestamp
+        next_year = Time.mktime(current_time.year + 1)
+        one_day_later = current_time.to_datetime.next_day.to_time
+        if timestamp >= next_year # Case 2.
+          ts_secs = 0
+          ts_nanos = 0
+        elsif timestamp >= one_day_later # Case 1.
+          adjusted_timestamp = timestamp.to_datetime.prev_year.to_time
+          # The value of ts_nanos should not change when subtracting a year.
```
igorpeshansky: Let's move this below the `ts_secs` assignment.

PR #185, opened 2017-11-29
Prefer the project id from the credentials file.

igorpeshansky:
Also fix/expand invalid credentials file tests.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -839,11 +839,9 @@ module Fluent
     # 2. If not, try to retrieve it by calling metadata server directly.
     # 3. If still not set, try to obtain it from the credentials.
     def set_project_id
+      @project_id ||= CredentialsInfo.project_id
       @project_id ||= fetch_gce_metadata('project/project-id') if
         @platform == Platform::GCE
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -839,11 +839,9 @@ def set_required_metadata_variables
     # 2. If not, try to retrieve it by calling metadata server directly.
     # 3. If still not set, try to obtain it from the credentials.
     def set_project_id
+      @project_id ||= CredentialsInfo.project_id
       @project_id ||= fetch_gce_metadata('project/project-id') if
         @platform == Platform::GCE
-      @project_id ||= CredentialsInfo.project_id
-    rescue StandardError => e
```
dhrupadb: Why remove the exception handling? Is the log overkill?

igorpeshansky: We want this to fail early and noisily when the credentials file is invalid. The exception handling caused it to fall back to `nil`, which then queried the metadata API, and the agent instead failed when trying to send a message with those credentials.

dhrupadb: Understood. Definitely a good idea to fail earlier.

PR #186, opened 2017-11-29
Add grpc and status code info in request and entry count metrics.

qingling128:
For gRPC, the official 'OK' code (https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core/StatusCode.cs) is 0. Yet for our case, 'successful_requests_count' simply gets mapped as response_code=200 regardless.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -393,7 +393,7 @@ module Fluent
           'The number of log entries that failed to be ingested by the'\
             ' Stackdriver output plugin due to a transient error and were'\
             ' retried')
-        @ok_code = @use_grpc ? 0 : 200
+        @ok_code = if @use_grpc then 0 else 200 end
       end
 
       # Alert on old authentication configuration.
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -393,7 +393,6 @@ module Fluent
           'The number of log entries that failed to be ingested by the'\
             ' Stackdriver output plugin due to a transient error and were'\
             ' retried')
-        @ok_code = if @use_grpc then 0 else 200 end
       end
 
       # Alert on old authentication configuration.
@@ -1877,7 +1876,7 @@ module Fluent
     # Increment the metric for the number of successful requests.
     def increment_successful_requests_count
       return unless @successful_requests_count
-      @successful_requests_count.increment(grpc: @use_grpc, code: @ok_code)
+      @successful_requests_count.increment(grpc: @use_grpc)
     end
 
     # Increment the metric for the number of failed requests, labeled by
@@ -1891,8 +1890,7 @@ module Fluent
     # ingested by the Stackdriver Logging API.
     def increment_ingested_entries_count(count)
       return unless @ingested_entries_count
-      @ingested_entries_count.increment({ grpc: @use_grpc, code: @ok_code },
-                                        count)
+      @ingested_entries_count.increment({ grpc: @use_grpc }, count)
     end
 
     # Increment the metric for the number of log entries that were dropped
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -393,7 +393,7 @@ module Fluent
           'The number of log entries that failed to be ingested by the'\
             ' Stackdriver output plugin due to a transient error and were'\
             ' retried')
-        @ok_code = if @use_grpc then 0 else 200 end
+        @ok_code = @use_grpc ? 0 : 200
       end
 
       # Alert on old authentication configuration.
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -393,6 +393,7 @@ module Fluent
           'The number of log entries that failed to be ingested by the'\
             ' Stackdriver output plugin due to a transient error and were'\
             ' retried')
+        @ok_code = @use_grpc ? 0 : 200
       end
 
       # Alert on old authentication configuration.
@@ -1876,7 +1877,7 @@ module Fluent
     # Increment the metric for the number of successful requests.
     def increment_successful_requests_count
       return unless @successful_requests_count
-      @successful_requests_count.increment(grpc: @use_grpc)
+      @successful_requests_count.increment(grpc: @use_grpc, code: @ok_code)
     end
 
     # Increment the metric for the number of failed requests, labeled by
@@ -1890,7 +1891,8 @@ module Fluent
     # ingested by the Stackdriver Logging API.
     def increment_ingested_entries_count(count)
       return unless @ingested_entries_count
-      @ingested_entries_count.increment({ grpc: @use_grpc }, count)
+      @ingested_entries_count.increment({ grpc: @use_grpc, code: @ok_code },
+                                        count)
     end
 
     # Increment the metric for the number of log entries that were dropped
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -632,17 +632,15 @@ module Fluent
               @log.debug "Retrying #{entries_count} log message(s) later.",
                          error: error.to_s, error_code: error.code.to_s
               raise error
-            when GRPC::Core::StatusCodes::UNIMPLEMENTED,
-                 GRPC::Core::StatusCodes::RESOURCE_EXHAUSTED
-              # Most client errors indicate a problem with the request itself
-              # and should not be retried.
-              increment_dropped_entries_count(entries_count, error.code)
-              @log.warn "Dropping #{entries_count} log message(s)",
-                        error: error.to_s, error_code: error.code.to_s
-            when GRPC::Core::StatusCodes::UNAUTHENTICATED
-              # Authorization error.
-              # These are usually solved via a `gcloud auth` call, or by
-              # modifying the permissions on the Google Cloud project.
+            when \
+                # Most client errors indicate a problem with the request itself
+                # and should not be retried.
+                GRPC::Core::StatusCodes::UNIMPLEMENTED,
+                GRPC::Core::StatusCodes::RESOURCE_EXHAUSTED,
+                # Authorization error.
+                # These are usually solved via a `gcloud auth` call, or by
+                # modifying the permissions on the Google Cloud project.
+                GRPC::Core::StatusCodes::UNAUTHENTICATED
               increment_dropped_entries_count(entries_count, error.code)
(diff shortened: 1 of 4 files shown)


qingling128:
It seems that `stackdriver_successful_requests_count` and `stackdriver_failed_requests_count` can be combined. So are `stackdriver_ingested_entries_count` and `stackdriver_dropped_entries_count`. If we decide to do that, I'll do it in a separate PR.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1892,21 +1893,22 @@ def increment_failed_requests_count(code)
     # ingested by the Stackdriver Logging API.
     def increment_ingested_entries_count(count)
       return unless @ingested_entries_count
-      @ingested_entries_count.increment({}, count)
+      @ingested_entries_count.increment({ grpc: @use_grpc, code: @ok_code },
```
igorpeshansky: There's no point in adding a code to `ingested_entries_count` — it's always the same.

test/plugin/test_out_google_cloud.rb
```diff
@@ -53,24 +53,22 @@ def test_client_401
 
   def test_partial_success
     setup_gce_metadata_stubs
-    {
-      APPLICATION_DEFAULT_CONFIG => 4.0,
```
igorpeshansky: Are we no longer testing this without partial success enabled?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1892,21 +1893,22 @@ def increment_failed_requests_count(code)
...
+                                        count)
     end
 
     # Increment the metric for the number of log entries that were dropped
     # and not ingested by the Stackdriver Logging API.
-    def increment_dropped_entries_count(count)
+    def increment_dropped_entries_count(count, code)
       return unless @dropped_entries_count
-      @dropped_entries_count.increment({}, count)
+      @dropped_entries_count.increment({ grpc: @use_grpc, code: code }, count)
```
igorpeshansky: Rather than sending a boolean to help us interpret what the code means, should we just add two labels, `http_status` and `grpc_error_code`, and only set one of them depending on the value of `@use_grpc`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -635,20 +636,20 @@ def write(chunk)
                  GRPC::Core::StatusCodes::RESOURCE_EXHAUSTED
               # Most client errors indicate a problem with the request itself
               # and should not be retried.
-              increment_dropped_entries_count(entries_count)
+              increment_dropped_entries_count(entries_count, error.code)
               @log.warn "Dropping #{entries_count} log message(s)",
                         error: error.to_s, error_code: error.code.to_s
             when GRPC::Core::StatusCodes::UNAUTHENTICATED
```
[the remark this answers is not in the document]
qingling128: Seems so. Changed.

lib/fluent/plugin/out_google_cloud.rb
qingling128: I was thinking about combining `stackdriver_successful_requests_count` and stackdriver_failed_requests_count`. Also `stackdriver_ingested_entries_count` and `stackdriver_dropped_entries_count`.

For instance, if we look at how these metrics get tranformed (https://github.com/Stackdriver/collectd/pull/125/files). They are combinable in some sense. WDYT?

lib/fluent/plugin/out_google_cloud.rb
qingling128: Does that imply we will have two separate Monarch metrics or two sets of labels in Monarch as well? It feels more difficult to maintain and filter that way IMO.

test/plugin/test_out_google_cloud.rb
qingling128: This has been covered in our regular `test_prometheus_metrics` already, thus redundant.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: I would propose keeping them separate in the agents and combining (if necessary) in the API-side mappings.
I'll comment on that PR separately.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: It means we'd have a single metric with two metric labels, only one of which would be set at a time... For gRPC, we'd submit `dropped_entries_count{grpc_error_code=...}`, and for REST it would have `dropped_entries_count{http_status=...}`.

igorpeshansky: Oh, never mind. I just realized that we've always had the `grpc` metric label. So all we're doing here is adding error code tracking. Please ignore this thread.

qingling128:
@crassirostris Just FYI. This might affect the current Logging Agent request and entry count monitoring setup in Kubernetes.

crassirostris:
@qingling128 OK, thanks for a heads up

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Sure. I've reverted the commit to add code explicitly. The only reason I did it is because the code is different between gRPC and REST. The former is 0, while the latter is 200. This this can be inferred via `use_grpc`.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Actually, looks like I got confused about my earlier confusion. You're actually adding the `grpc` label here. So this thread still applies.

qingling128: [Previously](Reference: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/master/lib/fluent/plugin/out_google_cloud.rb#L1917) we only have `use_grpc` label for `increment_successful_requests_count` and `increment_failed_requests_count`. Here the same label are added for the other three metrics.

At least for `increment_ingested_entries_count`, we have to provide a way to infer the right code (`0` for `gRPC` and `200` for `REST`). The options are:
1. Explicitly pass `0` or `200` as the `OK` code. This way we don't care about the `use_grpc` boolean.
2. Skip the `0` or `200` code but pass in a `use_grpc` boolean, so later on we can infer the code from the boolean.
3. Pass in both the boolean and the code explicitly. This way minimum logic is required when translating the metrics. But it might be a bit redundant.

Because of the fact that we already have `use_grpc` for 2 out of the 5 metrics, the option 2 feels more natural to me.

igorpeshansky: Hmm, why would we care about the _error_ code for _successfully_ ingested metrics?

igorpeshansky: Ok, never mind, I see. We're already combining the metrics, so the only way to find out which ones are successful would be to look at the error code. Sorry for my earlier confusion.
The most consistent thing to do in this case is actually option 3 -- that way all metrics have the same labels. Again, sorry to make you redo the work, but you can reinstate https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/186/commits/3667b07eb17ce5254e3643db914b05b75d1f13d2.

qingling128: Right. We are combining the metrics in the next step (Collectd translation).

No worries. I've just restored that commit (I kept it in its own commit because I was debating between option 2 and option 3).

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1892,21 +1891,22 @@ def increment_failed_requests_count(code)
...
       return unless @ingested_entries_count
-      @ingested_entries_count.increment({}, count)
+      @ingested_entries_count.increment({ grpc: @use_grpc, code: @ok_code },
+                                        count)
     end
 
     # Increment the metric for the number of log entries that were dropped
     # and not ingested by the Stackdriver Logging API.
-    def increment_dropped_entries_count(count)
+    def increment_dropped_entries_count(count, code)
```
jkohen: Can you give some background on this change? It looks like you're adding an RPC error code to this metric, but entries could be dropped for other reasons.

Also the Kubernetes Monitoring team may be using this metric. Have you checked with them that the new metric field won't break anything in their existing collection and queries?

crassirostris: It will, but only when the new fluentd is rolled out, and that's when we're going to fix stuff

qingling128: @jkohen - The reason we are adding an RPC error code is because we drop / retry entries for various reasons and we want to track the reason.

```
            when \
                # HTTP status 400 (Bad Request).
                GRPC::Core::StatusCodes::INVALID_ARGUMENT,
                # HTTP status 401 (Unauthorized).
                # These are usually solved via a `gcloud auth` call, or by
                # modifying the permissions on the Google Cloud project.
                GRPC::Core::StatusCodes::UNAUTHENTICATED,
                # HTTP status 403 (Forbidden).
                GRPC::Core::StatusCodes::PERMISSION_DENIED,
                # HTTP status 404 (Not Found).
                GRPC::Core::StatusCodes::NOT_FOUND,
                # HTTP status 409 (Conflict).
                GRPC::Core::StatusCodes::ABORTED,
                # HTTP status 412 (Precondition Failed).
                GRPC::Core::StatusCodes::FAILED_PRECONDITION,
                # HTTP status 429 (Too Many Requests).
                GRPC::Core::StatusCodes::RESOURCE_EXHAUSTED,
                # HTTP status 499 (Client Closed Request).
                GRPC::Core::StatusCodes::CANCELLED,
                # the remaining http codes in both 4xx and 5xx category.
                # It's debatable whether to retry or drop these log entries.
                # This decision is made to avoid retrying forever due to client
                # errors.
                GRPC::Core::StatusCodes::UNKNOWN
              increment_dropped_entries_count(entries_count, error.code)
```

These only covers dropped entries due to RPC errors though. On the other hand, we also drop log entries for malformed record/tag.
```
        unless record.is_a?(Hash)
          @log.warn 'Dropping log entries with malformed record: ' \
                    "'#{record.inspect}'. " \
                    'A log record should be in JSON format.'
          next
        end
        sanitized_tag = sanitize_tag(tag)
        if sanitized_tag.nil?
          @log.warn "Dropping log entries with invalid tag: '#{tag.inspect}'." \
                    ' A tag should be a string with utf8 characters.'
          next
        end
```
I've just created a ticket to add counter tracking for those as well.

And yes, this will affect Kubernetes metrics. I've given Mik a heads up of the coming changes. Will update when the change is actually released.

qingling128:
@crassirostris - Just FYI, this is released in fluent-plugin-google-cloud [v0.6.11](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/releases/tag/v0.6.11).

PR #188, opened 2017-11-30
Remove redundant GRPC::Cancelled case.

qingling128:
Per https://github.com/grpc/grpc/blob/master/src/ruby/lib/grpc/errors.rb#L91, `GRPC::Cancelled` is a subclass of `GRPC::BadStatus`, thus already covered in the other case. Removing the redundant part.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -600,14 +600,6 @@ module Fluent
               @log.info 'Successfully sent gRPC to Stackdriver Logging API.'
             end
 
-          rescue GRPC::Cancelled => error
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -600,14 +600,6 @@ def write(chunk)
               @log.info 'Successfully sent gRPC to Stackdriver Logging API.'
             end
 
-          rescue GRPC::Cancelled => error
-            increment_failed_requests_count(GRPC::Core::StatusCodes::CANCELLED)
-            increment_retried_entries_count(entries_count, error.code)
```
igorpeshansky: Might want to propagate this line below (i.e., replace `entries.length` with `entries_count`, since we've bothered computing it).

PR #191, opened 2017-11-30
Stop counting retried request towards stackdriver_failed_requests_count metrics.

qingling128:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -601,7 +601,6 @@ def write(chunk)
             end
 
           rescue GRPC::BadStatus => error
-            increment_failed_requests_count(error.code)
```
qingling128: Quoting a comment from the other PR:
```
It looks like when the retry limit is reached, the output plugin just doesn't get invoked, so the buffer data will be discarded silently. We may have to keep track of the retry counter ourselves, and on the last retry increment the "dropped" entry count instead.
```

Some obvious solutions to that issue:
1. Keep track of a `map` (key: `request`, value: `tried_count`), compare the `tried_count` towards `retry_limit`. `GoogleCloudOutput`'s parent class `BufferedOutput` has a variables `@disable_retry_limit` and `@retry_limit`.
Pros: The counter should be quite accurate.
Cons: Additional CPU/memory usage and potential latency due to the computation we need.
Open question: We need to "flush" the mapped request when they are sent successful. That probably means we will compare every successful request against the `map`. Or maybe we simple set a TTL, and flush older `request`.

2. Add a `increment_retried_requests_count`metrics separately.
Pros: Minimum additional CPU/memory usage or latency. Users should be able to rely on this counter to detect `forever retrying issues`.
Cons: The `failed_requests_count` counter will not be very accurate. Requests that exceed `retry_limit` will not be counted.

WDYT?

qingling128: Also mentioned this in the internal ticket. Maybe we can discuss it there. Will open a separate PR to address that.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -721,6 +718,8 @@ def write(chunk)
                           error_code: "google.rpc.Code[#{error_code}]",
                           error: error_message
               end
+              # Consider partially successful request successful.
```
igorpeshansky: Nit: `partially successful requests`.

qingling128:
PTAL.

qingling128:
Rebased. PTAL.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: SG, let's discuss on a separate PR.

qingling128:
This is released in fluent-plugin-google-cloud [v0.6.11](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/releases/tag/v0.6.11) and google-fluentd [v1.5.24](https://github.com/GoogleCloudPlatform/google-fluentd/releases/tag/v1.5.24).

PR #174, opened 2017-10-31
Add partial success for gRPC path.

qingling128:
This PR depends on https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/192.

PR #194, opened 2017-12-01
Add stackdriver_partially_successful_requests_count metric.

qingling128:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1933,6 +1936,13 @@ def increment_successful_requests_count
       @successful_requests_count.increment(grpc: @use_grpc, code: @ok_code)
     end
 
+    # Increment the metric for the number of partially successful requests.
+    def increment_partially_successful_requests_count(code)
+      return unless @partially_successful_requests_count
+      @partially_successful_requests_count.increment(grpc: @use_grpc,
```
bmoyles0117: Will @use_grpc always be false for now while we only invoke this function in the non-grpc block? If so, should we hard code it, or at least parameterize it to the calling function to prevent depending on the class variable?

qingling128: There is a coming PR to add partial success for gRPC. So soon this method will be invoked by both paths.

qingling128: For reference, that PR is https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/174.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -718,8 +722,7 @@ def write(chunk)
                           error_code: "google.rpc.Code[#{error_code}]",
                           error: error_message
               end
-              # Consider partially successful request successful.
-              increment_successful_requests_count
+              increment_partially_successful_requests_count(error.status_code)
```
jkohen: The name is confusing. Partially successful contain an error code, so it's not successful. Also error.status_code is always CLientError, is that what you expected? If not, you may need to move this inside the loop and use error_code.

qingling128: One partially successful request can contain multiple failed entries for different reasons with different error code. If we move it into the loop, it's counting the failed entries not the requests.

On the other hand the error code is not always `ClientError`. It is one of the two:
1. HTTP status 400 (Bad Request) / GRPC InvalidArgument
2. HTTP status 403 (Forbidden) / GRPC PermissionDenied

According to the Logging API's doc https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write,
> If any entry is not written, then the response status is the error associated with one of the failed entries.

From my observation, it seems to be the first failed entry's reason.

jkohen: Thanks for the explanation. I'm still not clear what the intention is for the status code on this metric, given that it's one arbitrary error from the response, and that we have all errors in the dropped_entries_count metric. For my curiousity, what do you plan to do with it?

qingling128: Right. We can't do much with that arbitrary error code. Let's discuss the design in the internal thread before moving on with the implementation here.

qingling128:
This was de-prioritized and we probably won't get to it in the near future. Closing for now.

PR #195, opened 2017-12-05
Add an option to control ingesting the availability zone instead of the region on AWS.

igorpeshansky:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -218,6 +218,9 @@ module InternalConstants
     # Whether to attempt to obtain metadata from the local metadata service.
     # It is safe to specify 'true' even on platforms with no metadata service.
     config_param :use_metadata_service, :bool, :default => true
+    # Whether to set the AWS location to the availability zone instead of the
+    # region.
+    config_param :use_aws_availability_zone, :bool, :default => true
```
qingling128: Should we also give recommendation/clear preference in the comment that `region` is the right thing while `availability zone` is some legacy residue that we are looking forward to deprecate?

test/plugin/base_test.rb
```diff
@@ -257,6 +259,17 @@ def test_one_log_ec2
...
       d.run
     end
+    verify_log_entries(1, EC2_ZONE_PARAMS)
+  end
+
+  def test_one_log_ec2_region
+    ENV['GOOGLE_APPLICATION_CREDENTIALS'] = IAM_CREDENTIALS[:path]
+    setup_ec2_metadata_stubs
+    setup_logging_stubs do
+      d = create_driver(CONFIG_EC2_USE_REGION)
```
qingling128: For consistency, should this be named `CONFIG_EC2_PROJECT_ID_USE_REGION`?

test/plugin/constants.rb
```diff
@@ -509,7 +517,7 @@ module Constants
       type: EC2_CONSTANTS[:resource_type],
       labels: {
         'instance_id' => EC2_VM_ID,
-        'region' => EC2_PREFIXED_ZONE,
+        'region' => EC2_PREFIXED_REGION,
```
qingling128: Should we rename `EC2_PARAMS` to `EC2_REGION_PARAMS`? It's a bit confusing for  `EC2_PARAMS` to use `region` while the default behavior (if the config is not set) still uses `availability zone`.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Also, we meant to leave `default` as `true` for the moment, I assume?

Issue #187, opened 2017-11-30
Memory leak introduced in 0.6.9 or 0.6.10


zbjornson:
Since google-fluentd updated from 1.5.22-1 to 1.5.23, we're seeing a ton of retained memory (820 MB retained for ~30 MB of logs over ~7 hours). The only change in google-fluentd between those versions is upgrading fluent-plugin-google-cloud from 0.6.8 to 0.6.10.

Looking at the [commit log](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/compare/v0.6.8...v0.6.10) here, it looks like this might be from another dependency (google-api-client, google-cloud-logging or gRPC?)...

(Someone else said they saw this also with 1.5.23 and not 1.5.20 in the Google Cloud Platform Slack community.)

igorpeshansky:
@qingling128 Mind taking a look?

qingling128:
Sure. We've noticed this memory leak issue during the recent benchmarking tests as well. Will take a look. I've also created an internal ticket to track it.

bivald:
I'm seeing this as well on 0.6.10 where google-fluentd (1.5.23) easily consumes 1GB of memory for a low log usage server. All logs are flushed to stackdriver (visible in the log viewer)

Perhaps more troubling for us then most other customers, since we run 50+ instances with low ram (g1-small etc)

```
enplore@admin:~$ free -m
             total       used       free     shared    buffers     cached
Mem:          1704       1626         77         16         46        239

enplore@admin:~$ sudo systemctl stop google-fluentd

enplore@admin:~$ free -m
             total       used       free     shared    buffers     cached
Mem:          1704        618       1086         16         46        239
```

Let me know if I can help with any testing or debugging

jrun:
To pile on here, we are seeing the same behavior. It's giving us a headache in our test environment because the instances don't have the extra memory to spare which leads to OOM errors.

We have instances in production where google-fluentd 1.5.23-1 is using 2GB+ RAM.

qingling128:
Sorry for the inconvenience. We are actively looking into this. Will see if we can quickly locate the root cause or at least a safe dependency version to roll back.

qingling128:
[...]

PR #196, opened 2017-12-06
Revert #180 and pin google-cloud-logging to < 1.3 and grpc to < 1.3 for memory leak.

qingling128:
We are seeing a serious / fast memory leak after that comes with gRPC v1.2.5 -> v1.4.5 upgrade which was introduced by a gRPC requirement in google-cloud-logging >= 1.3.0.

For a short-term mitigation for https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/187, we might want to roll it back.

See internal memory leak investigation doc for more details.

### DIFF
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -23,6 +23,7 @@ eos
   gem.add_runtime_dependency 'googleapis-common-protos', '~> 1.3'
   gem.add_runtime_dependency 'google-api-client', '~> 0.14'
   gem.add_runtime_dependency 'google-cloud-logging', '~> 1.2.3'
+  gem.add_runtime_dependency 'googleauth', '~> 0.5'
   gem.add_runtime_dependency 'grpc', '~> 1.2.5'
   gem.add_runtime_dependency 'json', '~> 1.8'
 
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -21,8 +21,8 @@ eos
 
   gem.add_runtime_dependency 'fluentd', '~> 0.10'
   gem.add_runtime_dependency 'googleapis-common-protos', '~> 1.3'
-  gem.add_runtime_dependency 'google-api-client', '~> 0.17'
-  gem.add_runtime_dependency 'google-cloud-logging', '~> 1.3', '>= 1.3.2'
+  gem.add_runtime_dependency 'google-api-client', '~> 0.14'
+  gem.add_runtime_dependency 'google-cloud-logging', '~> 1.2', '>= 1.2.3'
   gem.add_runtime_dependency 'googleauth', '~> 0.6'
   gem.add_runtime_dependency 'grpc', '~> 1.0'
   gem.add_runtime_dependency 'json', '~> 1.8'
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -22,9 +22,9 @@ eos
   gem.add_runtime_dependency 'fluentd', '~> 0.10'
   gem.add_runtime_dependency 'googleapis-common-protos', '~> 1.3'
   gem.add_runtime_dependency 'google-api-client', '~> 0.14'
-  gem.add_runtime_dependency 'google-cloud-logging', '~> 1.2', '>= 1.2.3'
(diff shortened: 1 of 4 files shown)


fluent-plugin-google-cloud.gemspec
```diff
@@ -21,10 +21,10 @@ eos
 
   gem.add_runtime_dependency 'fluentd', '~> 0.10'
   gem.add_runtime_dependency 'googleapis-common-protos', '~> 1.3'
-  gem.add_runtime_dependency 'google-api-client', '~> 0.17'
-  gem.add_runtime_dependency 'google-cloud-logging', '~> 1.3', '>= 1.3.2'
-  gem.add_runtime_dependency 'googleauth', '~> 0.6'
-  gem.add_runtime_dependency 'grpc', '~> 1.0'
+  gem.add_runtime_dependency 'google-api-client', '~> 0.14'
+  gem.add_runtime_dependency 'google-cloud-logging', '~> 1.2', '<= 1.2.3'
+  gem.add_runtime_dependency 'googleauth', '~> 0.5'
```
igorpeshansky: Why do we need to pin `googleauth`? Does that contribute to the leak, or is it not backward-compatible enough for the earlier version of `grpc`?

fluent-plugin-google-cloud.gemspec
```diff
@@ -21,10 +21,10 @@ eos
 
   gem.add_runtime_dependency 'fluentd', '~> 0.10'
   gem.add_runtime_dependency 'googleapis-common-protos', '~> 1.3'
-  gem.add_runtime_dependency 'google-api-client', '~> 0.17'
-  gem.add_runtime_dependency 'google-cloud-logging', '~> 1.3', '>= 1.3.2'
-  gem.add_runtime_dependency 'googleauth', '~> 0.6'
-  gem.add_runtime_dependency 'grpc', '~> 1.0'
+  gem.add_runtime_dependency 'google-api-client', '~> 0.14'
+  gem.add_runtime_dependency 'google-cloud-logging', '~> 1.2', '<= 1.2.3'
```
igorpeshansky: Why the `<=` change?

qingling128: Because `google-cloud-logging` [v1.3.0](https://rubygems.org/gems/google-cloud-logging/versions/1.3.0) requires `google-gax` [v0.10.1](https://rubygems.org/gems/google-gax/versions/0.10.1) which requires `gRPC` `1.7.2`.

To verify the memory leak is not from `google-gax`, I tried the same `google-cloud-logging` and `google-gax` version with different `gRPC` version `v1.2.5` and `v1.4.5`. The memory leak exists in `v1.4.5`, but not `v1.2.5`.

fluent-plugin-google-cloud.gemspec
qingling128: It is not backward-compatible with previous version of `google-cloud-logging`.

qingling128:
Feedback addressed. PTAL.

fluent-plugin-google-cloud.gemspec
igorpeshansky: Why not `'< 1.3'`, then? Should we just make it `'~> 1.2.3'`?

fluent-plugin-google-cloud.gemspec
igorpeshansky: I'm confused — before #180 we already had `googleauth` at `0.6`. Were we just picking up a new enough version of `google-cloud-logging` that this wasn't a problem?

fluent-plugin-google-cloud.gemspec
qingling128: Changed to `'~> 1.2.3'`.

fluent-plugin-google-cloud.gemspec
qingling128: The dependency chains are:

`google-cloud-logging` depends on `google-gax`.
`google-gax` depends on `grpc` and `googleauth`.
`grpc` also depends on `googleauth`.

I just double checked that we need `'googleauth', '~> 0.5'` because:
https://rubygems.org/gems/grpc/versions/1.2.5-universal-darwin.

On a second thought, I think we should just remove the explicit `googleauth` version and let dependencies handle it.

qingling128:
Added in the PR description.

qingling128:
PTAL.

PR #198, opened 2017-12-07
Revert #179, because it breaks authentication.

### LINKED ISSUE #197: Authentication is broken in latest release
The official google-fluentd logger fails to authenticate with stack driver since the 0.6.11 release.

This happens on instances in AWS EC2. Our instances are pretty basic instances running Debian Linux 9.2

We install google-fluentd using the script at https://dl.google.com/cloudagents/install-logging-agent.sh

Our systems have been working fine until the 0.6.11 release yesterday.

If we look in the logs here:
`less /var/log/google-fluentd/google-fluentd.log`

We now see:
```
Dropping 130 log message(s) error="16:#<Signet::AuthorizationError: Authorization failed.  Server message:\n{\n \"error\": \"invalid_
scope\",\n \"error_description\": \"Invalid downscoping, scopes should not be specified as a request parameter.\"\n}>" error_code="16"
```

I can fix it by reverting to 0.6.10 using the following commands:

` sudo /opt/google-fluentd/embedded/bin/fluent-gem uninstall fluent-plugin-google-cloud googleauth google-cloud-logging`

`sudo /opt/google-fluentd/embedded/bin/fluent-gem install fluent-plugin-google-cloud -v 0.6.10`

This reproduces consistently.

igorpeshansky:
Should fix #197.

@pkern FYI.

### DIFF
--- a/test/plugin/test_out_google_cloud_grpc.rb
+++ b/test/plugin/test_out_google_cloud_grpc.rb
@@ -238,10 +238,8 @@ class GoogleCloudOutputGRPCTest < Test::Unit::TestCase
 
     def api_client
       ssl_creds = GRPC::Core::ChannelCredentials.new
-      authentication = Google::Auth.get_application_default(LOGGING_SCOPE)
-      creds = GRPC::Core::CallCredentials.new(lambda do |a_hash|
-        authentication.apply(a_hash, use_configured_scope: true)
-      end)
+      authentication = Google::Auth.get_application_default
+      creds = GRPC::Core::CallCredentials.new(authentication.updater_proc)
       ssl_creds.compose(creds)
 
       @grpc_stub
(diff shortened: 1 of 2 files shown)


test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -238,10 +238,8 @@ def initialize(grpc_stub)
 
     def api_client
       ssl_creds = GRPC::Core::ChannelCredentials.new
-      authentication = Google::Auth.get_application_default(LOGGING_SCOPE)
-      creds = GRPC::Core::CallCredentials.new(lambda do |a_hash|
-        authentication.apply(a_hash, use_configured_scope: true)
-      end)
+      authentication = Google::Auth.get_application_default
+      creds = GRPC::Core::CallCredentials.new(authentication.updater_proc)
       ssl_creds.compose(creds)
```
bmoyles0117: Why isn't this assigned like it is above? `creds = ssl_creds...`

igorpeshansky: This was present in the original test code, which basically means the test wasn't doing the right thing, but unrelated to this rollback.

qingling128:
This has been released in [v0.6.12](https://rubygems.org/gems/fluent-plugin-google-cloud/versions/0.6.12).

PR #201, opened 2017-12-27
Refactor writing requests as prerequisite of tag grouping revamp.

qingling128:
This is a prerequisite of https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/202.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -455,11 +455,9 @@ module Fluent
 
       if @use_grpc
         @construct_log_entry = method(:construct_log_entry_in_grpc_format)
-        @construct_log_request = method(:construct_log_request_in_grpc_format)
         @write_request = method(:write_request_via_grpc)
       else
         @construct_log_entry = method(:construct_log_entry_in_rest_format)
-        @construct_log_request = method(:construct_log_request_in_rest_format)
         @write_request = method(:write_request_via_rest)
       end
 
@@ -549,11 +547,12 @@ module Fluent
         log_name = "projects/#{@project_id}/logs/#{log_name(
           tag, group_level_resource)}"
 
-        requests_to_send << @construct_log_request.call(
-          entries,
-          log_name,
-          group_level_resource,
-          group_level_common_labels)
+        requests_to_send << {
+          'entries' => entries,
+          'log_name' => log_name,
+          'resource' => group_level_resource,
+          'labels' => group_level_common_labels
+        }
       end
 
       requests_to_send.each do |request|
@@ -608,45 +607,20 @@ module Fluent
       )
     end
 
-    def construct_log_request_in_grpc_format(entries,
-                                             log_name,
-                                             resource,
-                                             labels)
-      utf8_encoded_labels = labels.map do |k, v|
-        [k.encode('utf-8'), convert_to_utf8(v)]
-      end.to_h
-      Google::Logging::V2::WriteLogEntriesRequest.new(
-        entries: entries,
-        log_name: log_name,
-        resource: Google::Api::MonitoredResource.new(
-          type: resource.type,
-          labels: resource.labels.to_h
-        ),
-        labels: utf8_encoded_labels,
-        partial_success: @partial_success)
-    end
-
-    def construct_log_request_in_rest_format(entries,
-                                             log_name,
-                                             resource,
-                                             labels)
-      Google::Apis::LoggingV2::WriteLogEntriesRequest.new(
-        entries: entries,
-        log_name: log_name,
-        resource: resource,
-        labels: labels,
-        partial_success: @partial_success)
-    end
-
     def write_request_via_grpc(request)
       client = api_client
-      entries_count = request.entries.length
+      entries_count = request['entries'].length
       client.write_log_entries(
         # Ignore partial_success for gRPC path.
-        request.entries,
-        log_name: request.log_name,
-        resource: request.resource,
-        labels: request.labels.to_h
+        request['entries'],
+        log_name: request['log_name'],
+        resource: Google::Api::MonitoredResource.new(
+          type: request['resource'].type,
+          labels: request['resource'].labels.to_h
+        ),
+        labels: request['labels'].map do |k, v|
+          [k.encode('utf-8'), convert_to_utf8(v)]
+        end.to_h
       )
       increment_successful_requests_count
       increment_ingested_entries_count(entries_count)
@@ -733,9 +707,15 @@ module Fluent
 
     def write_request_via_rest(request)
       client = api_client
-      entries_count = request.entries.length
+      entries_count = request['entries'].length
       client.write_entry_log_entries(
-        request,
+        Google::Apis::LoggingV2::WriteLogEntriesRequest.new(
+          entries: request['entries'],
+          log_name: request['log_name'],
+          resource: request['resource'],
+          labels: request['labels'],
+          partial_success: @partial_success
+        ),
         options: { api_format_version: '2' }
       )
       increment_successful_requests_count
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -295,13 +295,13 @@ module Fluent
     config_param :labels, :hash, :default => nil
 
     # Whether to use gRPC instead of REST/JSON to communicate to the
-    # Cloud Logging API.
+    # Stackdriver Logging API.
     config_param :use_grpc, :bool, :default => false
 
     # Whether valid entries should be written even if some other entries fail
     # due to INVALID_ARGUMENT or PERMISSION_DENIED errors when communicating to
-    # the Cloud Logging API. This is highly recommended. Right now this only
-    # works with the REST path (use_grpc = false).
+    # the Stackdriver Logging API. This is highly recommended. Right now this
+    # only works with the REST path (use_grpc = false).
     config_param :partial_success, :bool, :default => false
 
     # Whether to allow non-UTF-8 characters in user logs. If set to true, any
@@ -578,7 +578,7 @@ module Fluent
           partial_success: @partial_success)
       end
       requests_to_send.each do |request|
-        # Does the actual write to the cloud logging api.
+        # Does the actual write to the Stackdriver Logging API.
         send(write_request_function, request)
       end
     end
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -577,9 +577,14 @@ module Fluent
           labels: group_level_common_labels,
           partial_success: @partial_success)
       end
+      write_request = method(if @use_grpc
+                               :write_request_via_grpc
+                             else
+                               :write_request_via_http
+                             end)
       requests_to_send.each do |request|
         # Does the actual write to the Stackdriver Logging API.
-        send(write_request_function, request)
+        write_request.call(request)
       end
     end
 
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -570,10 +570,6 @@ module Fluent
 
         # Does the actual write to the cloud logging api.
         if @use_grpc
-          labels_utf8_pairs = group_level_common_labels.map do |k, v|
-            [k.encode('utf-8'), convert_to_utf8(v)]
-          end
-
           requests_to_send << Google::Logging::V2::WriteLogEntriesRequest.new(
             entries: entries,
             log_name: log_name,
@@ -581,15 +577,16 @@ module Fluent
               type: group_level_resource.type,
               labels: group_level_resource.labels.to_h
             ),
-            labels: labels_utf8_pairs.to_h)
+            labels: group_level_common_labels,
+            partial_success: @partial_success)
         else
           requests_to_send << \
             Google::Apis::LoggingV2::WriteLogEntriesRequest.new(
+              entries: entries,
               log_name: log_name,
               resource: group_level_resource,
               labels: group_level_common_labels,
-              partial_success: @partial_success,
-              entries: entries)
+              partial_success: @partial_success)
         end
       end
       requests_to_send.each do |request|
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -568,176 +569,188 @@ def write(chunk)
...
+              resource: group_level_resource,
+              labels: group_level_common_labels,
+              partial_success: @partial_success,
+              entries: entries)
+        end
+      end
+      requests_to_send.each do |request|
+        client = api_client
+        if @use_grpc
+          write_request_via_grpc(client, request)
```
igorpeshansky: There's got to be a way to just assign a callable value conditionally somewhere at the top and then invoke it here unconditionally.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -568,176 +569,188 @@ def write(chunk)
           tag, group_level_resource)}"
 
         # Does the actual write to the cloud logging api.
-        client = api_client
         if @use_grpc
-          begin
-            labels_utf8_pairs = group_level_common_labels.map do |k, v|
-              [k.encode('utf-8'), convert_to_utf8(v)]
-            end
+          labels_utf8_pairs = group_level_common_labels.map do |k, v|
```
igorpeshansky: Since we're decomposing the request in `write_request_via_grpc` anyway, would it make sense to fold these tranformations into `write_request_via_grpc` and send in the same request object as on the HTTP path? We could even set `partial_success` on that one and ignore it (with a comment) in `write_request_via_grpc`.

qingling128: The utf8 transformation has been folded into `write_request_via_grpc`. Also made the changes to pipe up requests here in REST format and let gRPC path decompose them later.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Changed it to use a caller function like this:

```
def write_request_function
  @use_grpc ? 'write_request_via_grpc' : 'write_request_via_http'
end
```

Then
```
send(write_request_function, client, request)
```

WDYT?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -567,177 +568,182 @@ def write(chunk)
...
-                partial_success: @partial_success,
-                entries: entries)
-            entries_count = entries.length
-            client.write_entry_log_entries(
-              write_request,
-              options: { api_format_version: '2' }
-            )
+    def write_request_via_grpc(client, request)
+      entries_count = request.entries.length
+      labels_utf8_pairs = request.labels.map do |k, v|
```
igorpeshansky: No longer "pairs", right? This is now a hash. How about `labels_utf8` or `utf8_encoded_labels`?

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Hmm, invoke by name (`send`) wouldn't be my first choice. Some alternatives (don't require a separate method):
1. Bound method:
   ```ruby
   write_request = method(if @use_grpc then
                            :write_request_via_grpc
                          else
                            :write_request_via_http
                          end)
   ...
   write_request.call(client, request)
   ```
2. Proc:
   ```ruby
   write_request = if @use_grpc then
                     proc {|v| write_request_via_grpc(v)}
                   else
                     proc {|v| write_request_via_http(v)}
                   end
   ...
   write_request.call(client, request)
   ```
3. Lambda:
   ```ruby
   write_request = if @use_grpc then
                     ->v {write_request_via_grpc(v)}
                   else
                     ->v {write_request_via_http(v)}
                   end
   ...
   write_request.call(client, request)
   ```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -567,177 +568,182 @@ def write(chunk)
...
+        # them later.
+        requests_to_send << Google::Apis::LoggingV2::WriteLogEntriesRequest.new(
+          entries: entries,
+          log_name: log_name,
+          resource: group_level_resource,
+          labels: group_level_common_labels,
+          partial_success: @partial_success)
+      end
+      requests_to_send.each do |request|
         client = api_client
```
igorpeshansky: You can now inline this directly into the call...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -567,177 +568,182 @@ def write(chunk)
...
-        if @use_grpc
-          begin
-            labels_utf8_pairs = group_level_common_labels.map do |k, v|
-              [k.encode('utf-8'), convert_to_utf8(v)]
-            end
+        # Does the actual write to the cloud logging api.
```
igorpeshansky: Nit: `Cloud Logging API`.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Nice catch. Rename `labels_utf8_pairs` to `utf8_encoded_labels`.

lib/fluent/plugin/out_google_cloud.rb
qingling128: I've folded `api_client` into the call.

BTW, I created b/71806931 to audit the way we use the api client (we are initiating a new one for every request).

lib/fluent/plugin/out_google_cloud.rb
qingling128: Renamed it to `Stackdriver Logging API` since that's how we refer to it in [external docs](https://cloud.google.com/logging/docs/reference/v2/rest/).

lib/fluent/plugin/out_google_cloud.rb
qingling128: Changed it to use Bound method.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: :+1:

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -567,177 +568,188 @@ def write(chunk)
...
+          labels: group_level_common_labels,
+          partial_success: @partial_success)
+      end
+      write_request = method(if @use_grpc then
+                               :write_request_via_grpc
+                             else
+                               :write_request_via_http
+                             end)
+      requests_to_send.each do |request|
+        # Does the actual write to the Stackdriver Logging API.
```
igorpeshansky: This comment may now be more useful on the `write_request` assignment above.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: SG. FWIW, I was thinking of simply folding it into the `write_request(api_client, request)` call, but your way is better.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: 😃

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: In fact, this is a bit weird now... We are send the same request (in REST format) for both REST and gRPC. But within the request, the entries still have different format.

qingling128: I just refactored that part to use callers to construct both entries and requests.

igorpeshansky: Yeah, I agree it's not the most intuitive approach. Now that we only have the single `request` argument, would it make sense to have an array of hashes, send it in as a bunch of decomposed pieces and reconstruct in both methods? E.g.:
```ruby
        requests_to_send << {
          entries: entries,
          log_name: log_name,
          resource: group_level_resource,
          labels: group_level_common_labels,
          partial_success: @partial_success
        }
      end
      # Does the actual write to the Stackdriver Logging API.
      write_request = method(if @use_grpc
                               :write_request_via_grpc
                             else
                               :write_request_via_http
                             end)
      requests_to_send.each do |request|
        write_request.call(**request)
      end
...


    def write_request_via_grpc(entries:, log_name:, resource:, labels:,
                               partial_success:)
      client = api_client
      entries_count = entries.length
      utf8_encoded_labels = labels.map do |k, v|
        [k.encode('utf-8'), convert_to_utf8(v)]
      end.to_h
      client.write_log_entries(
        # Ignore partial_success for gRPC path.
        entries,
        log_name: log_name,
        resource: Google::Api::MonitoredResource.new(
          type: resource.type,
          labels: resource.labels.to_h
        ),
        labels: utf8_encoded_labels
      )
      ...

    def write_request_via_http(entries:, log_name:, resource:, labels:,
                               partial_success:)
      client = api_client
      entries_count = entries.length
      client.write_entry_log_entries(
        Google::Apis::LoggingV2::WriteLogEntriesRequest.new(**request),
        options: { api_format_version: '2' }
      )
```

WDYT?

qingling128: That's an interesting idea. I gave it a try. Unfortunately `Keyword Arguments` is not supported by Ruby 2.1... Right now we support Ruby 2.1 - 2.4 (https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/master/.travis.yml).

BTW, I made some change in commit f7337c56cd44546f39ed1cd7e658af33096d0cec to use callers.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -453,6 +453,16 @@ def configure(conf)
       @resource.labels.freeze
       @common_labels.freeze
 
+      if @use_grpc
+        @construct_log_entry = method(:construct_log_entry_in_grpc_format)
+        @construct_log_request = method(:construct_log_request_in_grpc_format)
```
igorpeshansky: Did you see my proposal of treating requests as hashes? We may still want the `@construct_log_entry` refactoring, but I think constructing the request and then decomposing it is overkill.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: That's fine, you can pass in the hash and access its elements, to make it Ruby 2.1-compatible. Keyword arguments make the code nicer, but aren't essential for this solution.

qingling128: Cool cool. Just changed the code to pass a hash instead of a request object.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Ah, saw it. I've changed it to use a hash instead.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -567,176 +547,226 @@ def write(chunk)
...
-              @successful_call = true
-              @log.info 'Successfully sent to Stackdriver Logging API.'
-            end
+    def construct_log_entry_in_grpc_format(common_labels,
+                                           resource,
+                                           severity,
+                                           ts_secs,
+                                           ts_nanos)
+      entry = Google::Logging::V2::LogEntry.new(
+        labels: common_labels,
```
igorpeshansky: I'm confused -- is this not using named arguments? Is this Ruby 2.1-compatible?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -567,176 +547,226 @@ def write(chunk)
...
-            increment_successful_requests_count
-            increment_ingested_entries_count(entries_count)
-
-            # Let the user explicitly know when the first call succeeded, to aid
-            # with verification and troubleshooting.
-            unless @successful_call
-              @successful_call = true
-              @log.info 'Successfully sent to Stackdriver Logging API.'
-            end
+    def construct_log_entry_in_grpc_format(common_labels,
```
igorpeshansky: s/`common_labels`/`labels`/. Also in `construct_log_entry_in_rest_format`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -567,176 +547,226 @@ def write(chunk)
...
 
-        # Does the actual write to the cloud logging api.
-        client = api_client
-        if @use_grpc
-          begin
-            labels_utf8_pairs = group_level_common_labels.map do |k, v|
-              [k.encode('utf-8'), convert_to_utf8(v)]
-            end
+        requests_to_send << {
+          'entries' => entries,
```
igorpeshansky: Why not use symbols:
```ruby
        requests_to_send << {
          entries: entries,
          log_name: log_name,
          resource: group_level_resource,
          labels: group_level_common_labels
      }
...
      request[:entries]
...
```
?

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Ah, right. Turned out it is not Ruby 2.1 that does not support keyword argument, but Rubocop not configured correctly to recognize the syntax. After setting `TargetRubyVersion` to `2.1` instead of the default `2.0`, I no longer get this error:

```
lib/fluent/plugin/out_google_cloud.rb:708:40: E: unexpected token tCOMMA
    def write_request_via_rest(entries:, log_name:, resource:, labels:)
```

I've changed the implementation to use keyword arguments instead.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -567,176 +547,226 @@ def write(chunk)
...
-            increment_ingested_entries_count(entries_count)
-
-            # Let the user explicitly know when the first call succeeded, to aid
-            # with verification and troubleshooting.
-            unless @successful_call
-              @successful_call = true
-              @log.info 'Successfully sent gRPC to Stackdriver Logging API.'
-            end
+      requests_to_send.each do |request|
+        @write_request.call(request)
```
igorpeshansky: Should this be `**request`, to translate the hash into keyword arguments? Or does it "just work"(tm)?

qingling128: It just works because `request` is the only param here. `**request` is needed only if it's passing "additional" params, e.g.:

```
@write_request.call(entries: something_here, **request)
```

PR #202, opened 2018-01-09
Add a flag to control whether to group log tags when sending requests to Stackdriver Logging API.

qingling128:
This depends on https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/201.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -341,6 +341,10 @@ module Fluent
     config_param :metadata_agent_url, :string,
                  :default => DEFAULT_METADATA_AGENT_URL
 
+    # Whether to split log entries with different log tags into different
+    # requests when talking to Stackdriver Logging API.
+    config_param :split_logs_by_tag, :bool, :default => true
+
     # rubocop:enable Style/HashSyntax
 
     # TODO: Add a log_name config option rather than just using the tag?
@@ -555,8 +559,24 @@ module Fluent
         }
       end
 
-      requests_to_send.each do |request|
-        @write_request.call(request)
+      if @split_logs_by_tag
+        requests_to_send.each do |request|
+          @write_request.call(request)
+        end
+      else
+        # Combine all requests into one. The request level "log_name" will be
+        # ported to the entry level. The request level "resource" and "labels"
+        # are ignored as they should have been folded into the entry level
+        # "resource" and "labels" already anyway.
+        combined_entries = []
+        requests_to_send.each do |request|
+          request[:entries].each do |entry|
+            # Modify entries in-place as they are not needed later on.
+            entry.log_name = request[:log_name]
+          end
+          combined_entries.concat(request[:entries])
+        end
+        @write_request.call(entries: combined_entries)
       end
     end
 
@@ -607,17 +627,23 @@ module Fluent
       )
     end
 
-    def write_request_via_grpc(entries:, log_name:, resource:, labels:)
+    def write_request_via_grpc(entries:,
+                               log_name: '',
+                               resource: nil,
+                               labels: {})
       client = api_client
       entries_count = entries.length
       client.write_log_entries(
         # Ignore partial_success for gRPC path.
         entries,
         log_name: log_name,
-        resource: Google::Api::MonitoredResource.new(
-          type: resource.type,
-          labels: resource.labels.to_h
-        ),
+        # Leave resource nil if it's nil.
+        resource: if resource
+                    Google::Api::MonitoredResource.new(
+                      type: resource.type,
+                      labels: resource.labels.to_h
+                    )
+                  end,
         labels: labels.map do |k, v|
           [k.encode('utf-8'), convert_to_utf8(v)]
         end.to_h
@@ -705,7 +731,10 @@ module Fluent
                  error: error.to_s
     end
 
-    def write_request_via_rest(entries:, log_name:, resource:, labels:)
+    def write_request_via_rest(entries:,
+                               log_name: '',
+                               resource: nil,
+                               labels: {})
       client = api_client
       entries_count = entries.length
       client.write_entry_log_entries(
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -620,7 +620,6 @@ module BaseTest
   def test_split_logs_by_tag
     setup_gce_metadata_stubs
     log_entries_count = 5
-    params = COMPUTE_PARAMS.reject { |k, _| k == :log_name }
     [
       [APPLICATION_DEFAULT_CONFIG, log_entries_count],
       [DISABLE_SPLIT_LOGS_BY_TAG_CONFIG, 1]
@@ -633,7 +632,10 @@ module BaseTest
           d.emit("tag#{i}", 'message' => log_entry(i))
         end
         d.run
-        verify_log_entries(log_entries_count, params, 'textPayload')
+        verify_log_entries(log_entries_count, COMPUTE_PARAMS_WITH_MULTI_TAGS,
+                           'textPayload')
+        # Verify the number of requests is different based on whether the
+        # 'split_logs_by_tag' flag is enabled.
         assert_prometheus_metric_value(:stackdriver_successful_requests_count,
                                        requests_count,
                                        grpc: use_grpc, code: ok_status_code)
@@ -1631,12 +1633,12 @@ module BaseTest
   # The caller can optionally provide a block which is called for each entry.
   def verify_log_entries(n, params, payload_type = 'textPayload')
     jsonify_log_entries
-    entries_count = 0
+    entry_count = 0
     @logs_sent.each do |request|
       request['entries'].each do |entry|
         unless payload_type.empty?
           assert entry.key?(payload_type),
-                 "Entry ##{entries_count} did not contain expected" \
+                 "Entry ##{entry_count} did not contain expected" \
                  " #{payload_type} key: #{entry}."
         end
 
@@ -1655,26 +1657,25 @@ module BaseTest
           else
             # For a request with multiple tags, each entry's log name should be
             # associated to the entry index.
-            "projects/#{params[:project_id]}/logs/tag#{entries_count}"
+            "projects/#{params[:project_id]}/logs/tag#{entry_count}"
           end
         assert_equal expected_log_name, log_name
         assert_equal params[:resource][:type], resource['type']
         check_labels resource['labels'], params[:resource][:labels]
         check_labels labels, params[:labels]
         if block_given?
-          yield(entry, entries_count)
+          yield(entry, entry_count)
         elsif payload_type == 'textPayload'
           # Check the payload for textPayload, otherwise it's up to the caller.
-          verify_default_log_entry_text(entry['textPayload'], entries_count,
+          verify_default_log_entry_text(entry['textPayload'], entry_count,
                                         entry)
         end
-        entries_count += 1
-        assert entries_count <= n, "Number of entries #{entries_count}" \
-                                   " exceeds expected number #{n}."
+        entry_count += 1
+        assert entry_count <= n,
+               "Number of entries #{entry_count} exceeds expected number #{n}."
       end
     end
-    assert entries_count == n, "Number of entries #{entries_count} does not" \
-                               " match expected number #{n}."
+    assert_equal n, entry_count
   end
 
   def verify_container_logs(log_entry_factory, expected_params)
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -620,19 +620,19 @@ module BaseTest
   def test_split_logs_by_tag
     setup_gce_metadata_stubs
     log_entries_count = 5
-    dynamic_log_names = Array(0..log_entries_count - 1).map do |index|
+    dynamic_log_names = (0..log_entries_count - 1).map do |index|
       "projects/test-project-id/logs/tag#{index}"
     end
     [
       [APPLICATION_DEFAULT_CONFIG,
        log_entries_count,
        dynamic_log_names,
-       Array.new(log_entries_count, nil)],
+       Array.new(log_entries_count)],
       [DISABLE_SPLIT_LOGS_BY_TAG_CONFIG,
        1,
-       Array.new(log_entries_count, ''),
+       [''],
        dynamic_log_names]
-    ].each do |(config, requests_count, request_log_names, entry_log_names)|
+    ].each do |(config, request_count, request_log_names, entry_log_names)|
       setup_prometheus
       setup_logging_stubs do
         @logs_sent = []
@@ -641,24 +641,24 @@ module BaseTest
           d.emit("tag#{i}", 'message' => log_entry(i))
         end
         d.run
-        @logs_sent.each_with_index do |request, request_index|
-          assert_equal request_log_names[request_index], request['logName']
-        end
-        verify_log_entries(log_entries_count, COMPUTE_PARAMS_WITH_MULTI_TAGS,
-                           'textPayload') do |entry, entry_index|
-          verify_default_log_entry_text(entry['textPayload'], entry_index,
-                                        entry)
-          assert_equal entry_log_names[entry_index], entry['logName']
-        end
-        # Verify the number of requests is different based on whether the
-        # 'split_logs_by_tag' flag is enabled.
-        assert_prometheus_metric_value(:stackdriver_successful_requests_count,
-                                       requests_count,
-                                       grpc: use_grpc, code: ok_status_code)
-        assert_prometheus_metric_value(:stackdriver_ingested_entries_count,
-                                       log_entries_count,
-                                       grpc: use_grpc, code: ok_status_code)
       end
+      @logs_sent.each_with_index do |request, request_index|
+        assert_equal request_log_names[request_index], request['logName']
+      end
+      verify_log_entries(log_entries_count, COMPUTE_PARAMS_WITH_MULTI_TAGS,
+                         'textPayload') do |entry, entry_index|
+        verify_default_log_entry_text(entry['textPayload'], entry_index,
+                                      entry)
+        assert_equal entry_log_names[entry_index], entry['logName']
+      end
+      # Verify the number of requests is different based on whether the
+      # 'split_logs_by_tag' flag is enabled.
+      assert_prometheus_metric_value(:stackdriver_successful_requests_count,
+                                     request_count,
+                                     grpc: use_grpc, code: ok_status_code)
+      assert_prometheus_metric_value(:stackdriver_ingested_entries_count,
+                                     log_entries_count,
+                                     grpc: use_grpc, code: ok_status_code)
     end
   end
 
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -1647,8 +1647,7 @@ module BaseTest
   end
 
   # The caller can optionally provide a block which is called for each entry.
-  def verify_log_entries(n, params, payload_type = 'textPayload')
-    jsonify_log_entries
+  def verify_json_log_entries(n, params, payload_type = 'textPayload')
     entry_count = 0
     @logs_sent.each do |request|
       request['entries'].each do |entry|
@@ -1834,8 +1833,9 @@ module BaseTest
     _undefined
   end
 
-  # Jsonify the log entries response if it is not in JSON format.
-  def jsonify_log_entries
+  # Verify the number and the content of the log entries match the expectation.
+  # The caller can optionally provide a block which is called for each entry.
+  def verify_log_entries(_n, _params, _payload_type = 'textPayload', &_block)
     _undefined
   end
 
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -649,9 +649,9 @@ module BaseTest
       # Verify the number of requests is different based on whether the
       # 'split_logs_by_tag' flag is enabled.
       assert_prometheus_metric_value(:stackdriver_successful_requests_count,
-                                     request_count, :all)
+                                     request_count, :aggregate)
       assert_prometheus_metric_value(:stackdriver_ingested_entries_count,
-                                     log_entry_count, :all)
+                                     log_entry_count, :aggregate)
     end
   end
 
@@ -1837,7 +1837,7 @@ module BaseTest
   def assert_prometheus_metric_value(metric_name, expected_value, labels = {})
     metric = Prometheus::Client.registry.get(metric_name)
     assert_not_nil(metric)
-    metric_value = if labels == :all
+    metric_value = if labels == :aggregate
                      # Sum up all metric values regardless of the labels.
                      metric.values.values.reduce(0.0, :+)
                    else
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -641,6 +641,30 @@ module BaseTest
     end
   end
 
+  def test_log_name_when_split_logs_by_tag_disabled
+    setup_gce_metadata_stubs
+    log_entries_count = 5
+    setup_prometheus
+    setup_logging_stubs do
+      d = create_driver(DISABLE_SPLIT_LOGS_BY_TAG_CONFIG +
+                        PROMETHEUS_ENABLE_CONFIG, 'test', true)
+      log_entries_count.times do |i|
+        d.emit("tag#{i}", 'message' => log_entry(0))
+      end
+      d.run
+    end
+    emit_index = 0
+    jsonify_log_entries
+    @logs_sent.each do |request_sent|
+      assert_equal '', request_sent['logName']
+      request_sent['entries'].each do |entry|
+        assert_equal "projects/test-project-id/logs/tag#{emit_index}",
+                     entry['logName']
+        emit_index += 1
+      end
+    end
+  end
+
   def test_timestamps
     setup_gce_metadata_stubs
     current_time = Time.now
@@ -1626,7 +1650,8 @@ module BaseTest
   end
 
   # The caller can optionally provide a block which is called for each entry.
-  def verify_json_log_entries(n, params, payload_type = 'textPayload')
+  def verify_log_entries(n, params, payload_type = 'textPayload')
+    jsonify_log_entries
     i = 0
     @logs_sent.each do |request|
       request['entries'].each do |entry|
@@ -1808,9 +1833,8 @@ module BaseTest
     _undefined
   end
 
-  # Verify the number and the content of the log entries match the expectation.
-  # The caller can optionally provide a block which is called for each entry.
-  def verify_log_entries(_n, _params, _payload_type = 'textPayload', &_block)
+  # Jsonify the log entries response if it is not in JSON format.
+  def jsonify_log_entries
     _undefined
   end
 
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -620,14 +620,11 @@ module BaseTest
   def test_split_logs_by_tag
     setup_gce_metadata_stubs
     log_entries_count = 5
+    params = COMPUTE_PARAMS.reject { |k, _| k == :log_name }
     [
-      [true, APPLICATION_DEFAULT_CONFIG, log_entries_count,
-       COMPUTE_PARAMS.reject { |k, _| k == :log_name }],
-      # Request level log name should be '' if should_split is false.
-      [false, DISABLE_SPLIT_LOGS_BY_TAG_CONFIG, 1,
-       COMPUTE_PARAMS.merge(
-         request: { log_name: '' }).reject { |k, _| k == :log_name }]
-    ].each do |(should_split, config, requests_count, params)|
+      [APPLICATION_DEFAULT_CONFIG, log_entries_count],
+      [DISABLE_SPLIT_LOGS_BY_TAG_CONFIG, 1]
+    ].each do |(config, requests_count)|
       setup_prometheus
       setup_logging_stubs do
         @logs_sent = []
@@ -636,14 +633,7 @@ module BaseTest
           d.emit("tag#{i}", 'message' => log_entry(i))
         end
         d.run
-        verify_log_entries(log_entries_count, params,
-                           'textPayload') do |entry, i|
-          verify_default_log_entry_text(entry['textPayload'], i, entry)
-          # expected_log_name is nil if should_split is true.
-          expected_log_name =
-            "projects/test-project-id/logs/tag#{i}" unless should_split
-          assert_equal expected_log_name, entry['logName']
-        end
+        verify_log_entries(log_entries_count, params, 'textPayload')
         assert_prometheus_metric_value(:stackdriver_successful_requests_count,
                                        requests_count,
                                        grpc: use_grpc, code: ok_status_code)
@@ -1643,9 +1633,6 @@ module BaseTest
     jsonify_log_entries
     entries_count = 0
     @logs_sent.each do |request|
-      if params[:request] && params[:request][:log_name]
-        assert_equal params[:request][:log_name], request['logName']
-      end
       request['entries'].each do |entry|
         unless payload_type.empty?
           assert entry.key?(payload_type),
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -605,6 +605,42 @@ module BaseTest
     end
   end
 
+  def test_configure_split_logs_by_tag
+    setup_gce_metadata_stubs
+    {
+      APPLICATION_DEFAULT_CONFIG => true,
+      DISABLE_SPLIT_LOGS_BY_TAG_CONFIG => false
+    }.each do |(config, split_logs_by_tag)|
+      d = create_driver(config)
+      assert_equal split_logs_by_tag,
+                   d.instance.instance_variable_get(:@split_logs_by_tag)
+    end
+  end
+
+  def test_split_logs_by_tag
+    setup_gce_metadata_stubs
+    log_entries_count = 5
+    [
+      [APPLICATION_DEFAULT_CONFIG, log_entries_count, log_entries_count],
+      [DISABLE_SPLIT_LOGS_BY_TAG_CONFIG, 1, log_entries_count]
+    ].each do |(config, successful_requests_count, ingested_entries_count)|
+      setup_prometheus
+      setup_logging_stubs do
+        d = create_driver(config + PROMETHEUS_ENABLE_CONFIG, 'test', true)
+        log_entries_count.times do |i|
+          d.emit("tag#{i}", 'message' => log_entry(0))
+        end
+        d.run
+      end
+      assert_prometheus_metric_value(:stackdriver_successful_requests_count,
+                                     successful_requests_count,
+                                     grpc: use_grpc, code: ok_status_code)
+      assert_prometheus_metric_value(:stackdriver_ingested_entries_count,
+                                     ingested_entries_count,
+                                     grpc: use_grpc, code: ok_status_code)
+    end
+  end
+
   def test_timestamps
     setup_gce_metadata_stubs
     current_time = Time.now
@@ -1751,6 +1787,16 @@ module BaseTest
   end
 
   # This module expects the methods below to be overridden.
+
+  # Whether use_grpc is enabled.
+  def use_grpc
+    _undefined
+  end
+
+  # Response code of OK status.
+  def ok_status_code
+    _undefined
+  end
 
   # Create a Fluentd output test driver with the Google Cloud Output plugin.
   def create_driver(_conf = APPLICATION_DEFAULT_CONFIG, _tag = 'test')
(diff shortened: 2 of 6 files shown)


test/plugin/test_driver.rb
```diff
@@ -0,0 +1,53 @@
+# Copyright 2017 Google Inc. All rights reserved.
```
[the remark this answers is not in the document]
qingling128: Changed.

test/plugin/test_driver.rb
```diff
@@ -0,0 +1,53 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'fluent/engine'
+require 'fluent/event'
+require 'fluent/test/input_test'
+
+module Fluent
+  module Test
+    class MultiTagBufferedOutputTestDriver < InputTestDriver
```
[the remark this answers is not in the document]
qingling128: Added a class level comment:
```
Similar to the standard BufferedOutputTestDriver, but allows multiple tags to exist in one chunk.
```

test/plugin/test_out_google_cloud.rb
```diff
@@ -308,9 +356,16 @@ def setup_logging_stubs
...
 
   # Create a Fluentd output test driver with the Google Cloud Output plugin.
-  def create_driver(conf = APPLICATION_DEFAULT_CONFIG, tag = 'test')
-    Fluent::Test::BufferedOutputTestDriver.new(
-      Fluent::GoogleCloudOutput, tag).configure(conf, true)
+  def create_driver(conf = APPLICATION_DEFAULT_CONFIG,
+                    tag = 'test',
+                    multi_tags = false)
+    if multi_tags
+      Fluent::Test::MultiTagBufferedOutputTestDriver.new(
```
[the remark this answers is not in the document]
qingling128: Done here and also in `test/plugin/test_out_google_cloud_grpc.rb`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -341,6 +341,10 @@ module InternalConstants
     config_param :metadata_agent_url, :string,
                  :default => DEFAULT_METADATA_AGENT_URL
 
+    # Whether to group log entries with different log tags into different
```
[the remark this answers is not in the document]
qingling128: `split_logs_by_tag` sounds good to me. @igorpeshansky WDYT?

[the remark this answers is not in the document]
igorpeshansky: Yes, makes sense.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -568,49 +551,103 @@ def write(chunk)
...
-        client = api_client
-        if @use_grpc
-          write_request_via_grpc(client, request)
-        else
-          write_request_via_http(client, request)
+      else
+        # Combine all requests into one. The request level "log_name" will be
+        # ported to the entry level. The request level "resource" and "labels"
+        # are ignored as they should have been folded into the entry level
+        # "resource" and "labels" already anyway.
```
igorpeshansky: Should mention somewhere that we're modifying the entries in-place...

qingling128:
Rebased off master after https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/201 was merged.

To quickly recap changes for previous feedback:

- Renamed should_group_log_tags to split_logs_by_tag.
- Changed log entries in place instead of modifying a copy.
- Changed copyright year from 2017 to 2018.
- Added class comment for MultiTagBufferedOutputTestDriver.
- Factored out configure method in create_driver methods.
- Combined gRPC and REST tests.

PTAL.

test/plugin/base_test.rb
```diff
@@ -1752,6 +1813,16 @@ def latency_conversion
 
   # This module expects the methods below to be overridden.
 
+  # Whether use_grpc is enabled.
+  def use_grpc
```
igorpeshansky: Hmm, would this enable factoring other calls to `assert_prometheus_metric_value` into this file?

test/plugin/base_test.rb
```diff
@@ -605,6 +605,66 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+    ].each do |(config, successful_requests_count, ingested_entries_count)|
+      setup_prometheus
+      setup_logging_stubs do
+        d = create_driver(config + PROMETHEUS_ENABLE_CONFIG, 'test', true)
+        log_entries_count.times do |i|
+          d.emit("tag#{i}", 'message' => log_entry(0))
+        end
+        d.run
+      end
+      assert_prometheus_metric_value(:stackdriver_successful_requests_count,
```
igorpeshansky: Should we not just check `@logs_sent` (the actual requests) rather than the Prometheus counter values?

test/plugin/base_test.rb
qingling128: A bit more refactoring is needed because the prometheus tests needs some error code to be passed into the `setup_logging_stub` method, and the signature is different among REST and gRPC. I'll address that in a separate PR.

test/plugin/base_test.rb
qingling128: Added some checks. In fact, I just combined the two tests `test_split_logs_by_tag` and `test_log_name_when_split_logs_by_tag_disabled`.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,66 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+                        PROMETHEUS_ENABLE_CONFIG, 'test', true)
+      log_entries_count.times do |i|
+        d.emit("tag#{i}", 'message' => log_entry(0))
+      end
+      d.run
+    end
+    emit_index = 0
+    jsonify_log_entries
+    @logs_sent.each do |request_sent|
+      assert_equal '', request_sent['logName']
```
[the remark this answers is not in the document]
qingling128: Cool cool. Changed `verify_log_entries` to skip the logName check if params[:log_name] is `nil`. I was debating whether to touch that function at all. But this makes sense i think.

As for the refactoring of `verify_json_log_entries`, yeah, it's not very relevant. I'm slightly leaning towards the refactored version simply because I can easily search for `def verify_lo_entries` in the `BaseTest` file. Most of the time I was debugging in that file. Just a very personal preference though.

qingling128:
Friendly ping.

test/plugin/base_test.rb
```diff
@@ -1608,8 +1660,16 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
         labels ||= request['labels']
         labels.merge!(entry['labels'] || {})
 
-        assert_equal \
-          "projects/#{params[:project_id]}/logs/#{params[:log_name]}", log_name
+        # For multi tag requests, each log name should match the index.
+        if params[:log_name] == 'multi'
+          assert_equal \
+            "projects/#{params[:project_id]}/logs/tag#{i}",
```
igorpeshansky: Hmm, why can't you check this in a separate loop for requests instead of inventing a special log name? You already check the entry-level `log_name` in the block.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,54 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+        @logs_sent = []
+        d = create_driver(config + PROMETHEUS_ENABLE_CONFIG, 'test', true)
+        log_entries_count.times do |i|
+          d.emit("tag#{i}", 'message' => log_entry(i))
+        end
+        d.run
+        verify_log_entries(log_entries_count, params,
+                           'textPayload') do |entry, i|
+          verify_default_log_entry_text(entry['textPayload'], i, entry)
+          # expected_log_name should be nil if should_split is true.
```
igorpeshansky: s/`should be`/`is`/.

test/plugin/base_test.rb
```diff
@@ -1590,9 +1638,13 @@ def verify_default_log_entry_text(text, i, entry)
   end
 
   # The caller can optionally provide a block which is called for each entry.
-  def verify_json_log_entries(n, params, payload_type = 'textPayload')
+  def verify_log_entries(n, params, payload_type = 'textPayload')
+    jsonify_log_entries
     i = 0
     @logs_sent.each do |request|
```
igorpeshansky: Might as well use `each_with_index` instead and avoid manually incrementing `i`.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,54 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+          assert_equal expected_log_name, entry['logName']
+        end
+        assert_prometheus_metric_value(:stackdriver_successful_requests_count,
+                                       requests_count,
+                                       grpc: use_grpc, code: ok_status_code)
+        assert_prometheus_metric_value(:stackdriver_ingested_entries_count,
+                                       log_entries_count,
+                                       grpc: use_grpc, code: ok_status_code)
+      end
+    end
```
igorpeshansky: You could just add a loop over `@logs_sent` that checks the `log_name` in each request, e.g.:
```ruby
@logs_sent.each_with_index do |log, i|
  expected_log_name =
    "projects/test-project-id/logs/tag#{i}" if should_split
  assert_equal expected_log_name, log['logName']
end
```

test/plugin/base_test.rb
```diff
@@ -1608,8 +1660,16 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
         labels ||= request['labels']
         labels.merge!(entry['labels'] || {})
 
-        assert_equal \
-          "projects/#{params[:project_id]}/logs/#{params[:log_name]}", log_name
+        # For multi tag requests, each log name should match the index.
+        if params[:log_name] == 'multi'
```
igorpeshansky: I would just check that `params[:log_name]` is non-`nil` and check for the constant log name, and otherwise do this in the block you pass to `verify_log_entries`.

qingling128: Good idea. Changed.

test/plugin/base_test.rb
qingling128: In fact, the complexity comes when I was trying to check both log names at the request level and the entry level. On a second thought, since entry level `log_name` always take precedence of the request level one, maybe it's better to check the `combined` log name instead (That is how `def verify_log_entries` has been doing it).

I just refactored the tests to simplify log name checks. I put it in a separate commit just to emphasize.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,45 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+      d = create_driver(config)
+      assert_equal split_logs_by_tag,
+                   d.instance.instance_variable_get(:@split_logs_by_tag)
+    end
+  end
+
+  def test_split_logs_by_tag
+    setup_gce_metadata_stubs
+    log_entries_count = 5
+    params = COMPUTE_PARAMS.reject { |k, _| k == :log_name }
```
igorpeshansky: This seems like a constant...

test/plugin/base_test.rb
igorpeshansky: Should you not still check that the log name in the request vs. the log name in each entry is set correctly?

test/plugin/base_test.rb
```diff
@@ -1608,22 +1649,32 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
         labels ||= request['labels']
         labels.merge!(entry['labels'] || {})
 
-        assert_equal \
-          "projects/#{params[:project_id]}/logs/#{params[:log_name]}", log_name
+        expected_log_name = \
+          if params[:log_name]
+            "projects/#{params[:project_id]}/logs/#{params[:log_name]}"
+          else
+            # For a request with multiple tags, each entry's log name should be
```
igorpeshansky: I would prefer not to build in this assumption, but if you choose to do so, please document at the top of `verify_log_entries` (because otherwise it's hard to dig for).
I still think it makes sense to do this check separately in the test case for anything that isn't a constant log name.

qingling128:
Does this one look good now?

test/plugin/base_test.rb
```diff
@@ -1608,22 +1649,32 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
...
         elsif payload_type == 'textPayload'
           # Check the payload for textPayload, otherwise it's up to the caller.
-          verify_default_log_entry_text(entry['textPayload'], i, entry)
+          verify_default_log_entry_text(entry['textPayload'], entries_count,
+                                        entry)
         end
-        i += 1
-        assert i <= n, "Number of entries #{i} exceeds expected number #{n}"
+        entries_count += 1
+        assert entries_count <= n, "Number of entries #{entries_count}" \
```
igorpeshansky: If you rename `entries_count` to `entry_count`, the error message will fit on the next line.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,45 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+      setup_prometheus
+      setup_logging_stubs do
+        @logs_sent = []
+        d = create_driver(config + PROMETHEUS_ENABLE_CONFIG, 'test', true)
+        log_entries_count.times do |i|
+          d.emit("tag#{i}", 'message' => log_entry(i))
+        end
+        d.run
+        verify_log_entries(log_entries_count, params, 'textPayload')
+        assert_prometheus_metric_value(:stackdriver_successful_requests_count,
```
igorpeshansky: Do we even need to assert these prometheus counts here? Or are you verifying that the requests were sent successfully?

test/plugin/base_test.rb
igorpeshansky: So it is. I was going to propose writing a wrapper to represent the nested loop, but it's probably not worth it at this point.

test/plugin/base_test.rb
qingling128: Ah, right. Moved to `constants.rb`.

test/plugin/base_test.rb
qingling128: Yeah, I added a comment to clarify:
```
        # Verify the number of requests is different based on whether the
        # 'split_logs_by_tag' flag is enabled.
```

test/plugin/base_test.rb
qingling128: Cool cool.

test/plugin/base_test.rb

test/plugin/base_test.rb
qingling128: I see. Moved this check back to the individual test case.

test/plugin/base_test.rb
qingling128: I kind of feel that it's redundant if we are checking the combined log names (`log_name = entry['logName'] || request['logName']`) as well. But since I'm pulling that out of the `verify_log_entries` method in the latest commit, I've restored the checks for request level log names and entry level log names separately.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,63 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+      "projects/test-project-id/logs/tag#{index}"
+    end
+    [
+      [APPLICATION_DEFAULT_CONFIG,
+       log_entries_count,
+       dynamic_log_names,
+       Array.new(log_entries_count, nil)],
+      [DISABLE_SPLIT_LOGS_BY_TAG_CONFIG,
+       1,
+       Array.new(log_entries_count, ''),
```
igorpeshansky: You'll only expect one request, right? So this should be [''].

test/plugin/base_test.rb
```diff
@@ -1590,13 +1647,15 @@ def verify_default_log_entry_text(text, i, entry)
   end
 
   # The caller can optionally provide a block which is called for each entry.
-  def verify_json_log_entries(n, params, payload_type = 'textPayload')
-    i = 0
+  def verify_log_entries(n, params, payload_type = 'textPayload')
+    jsonify_log_entries
```
igorpeshansky: I still really dislike having to call this (one, because it's easy to miss, and thus error-prone, and two, because it has no return value, and seems to be modifying the data in-place). The old factory method pattern was a lot clearer, IMO.
In any case, I'd call this `compute_json_log_entries`, or even `compute_logs_sent`.

test/plugin/base_test.rb
```diff
@@ -1762,9 +1834,8 @@ def setup_logging_stubs
     _undefined
   end
 
-  # Verify the number and the content of the log entries match the expectation.
-  # The caller can optionally provide a block which is called for each entry.
-  def verify_log_entries(_n, _params, _payload_type = 'textPayload', &_block)
+  # Jsonify the log entries response if it is not in JSON format.
```
igorpeshansky: Careful -- this is not about the response, this is about the ingested entries themselves.

test/plugin/base_test.rb
igorpeshansky: I guess I was suggesting that you check the absence of the log name in places where it wasn't supposed to be, which, together with the combined checks, would have produced the right result. But doing them explicitly works too.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,63 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+      d = create_driver(config)
+      assert_equal split_logs_by_tag,
+                   d.instance.instance_variable_get(:@split_logs_by_tag)
+    end
+  end
+
+  def test_split_logs_by_tag
+    setup_gce_metadata_stubs
+    log_entries_count = 5
+    dynamic_log_names = Array(0..log_entries_count - 1).map do |index|
```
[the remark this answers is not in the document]
qingling128: I was using `log_entries_count.times.map` but rubocop does not like it...

```
C: Use Array.new with a block instead of .times.map
```

`(0..log_entries_count - 1).map` seems to work though. Just changed to that.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,63 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+      "projects/test-project-id/logs/tag#{index}"
+    end
+    [
+      [APPLICATION_DEFAULT_CONFIG,
+       log_entries_count,
+       dynamic_log_names,
+       Array.new(log_entries_count, nil)],
```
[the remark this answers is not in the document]

test/plugin/base_test.rb
qingling128: Right. Changed.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,63 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+      [DISABLE_SPLIT_LOGS_BY_TAG_CONFIG,
+       1,
+       Array.new(log_entries_count, ''),
+       dynamic_log_names]
+    ].each do |(config, requests_count, request_log_names, entry_log_names)|
```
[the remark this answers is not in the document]

test/plugin/base_test.rb
```diff
@@ -605,6 +605,63 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+      setup_prometheus
+      setup_logging_stubs do
+        @logs_sent = []
+        d = create_driver(config + PROMETHEUS_ENABLE_CONFIG, 'test', true)
+        log_entries_count.times do |i|
+          d.emit("tag#{i}", 'message' => log_entry(i))
+        end
+        d.run
+        @logs_sent.each_with_index do |request, request_index|
```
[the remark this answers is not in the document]
qingling128: That seems to be the convention. Moved the checks outside.

test/plugin/constants.rb
```diff
@@ -265,6 +269,8 @@ module Constants
       }
     )
   ).freeze
+  COMPUTE_PARAMS_WITH_MULTI_TAGS =
```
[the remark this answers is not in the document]
qingling128: That's doable, but it will require touching many tests. If we decide to do it, I'll make the change in a separate PR.

```
grep -r COMPUTE_PARAMS .
./test/plugin/base_test.rb:    verify_log_entries(1, COMPUTE_PARAMS)
./test/plugin/base_test.rb:    verify_log_entries(1, COMPUTE_PARAMS.merge(
./test/plugin/base_test.rb:    verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry, i|
./test/plugin/base_test.rb:    verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
./test/plugin/base_test.rb:    verify_log_entries(3, COMPUTE_PARAMS, 'textPayload') do
./test/plugin/base_test.rb:    verify_log_entries(6, COMPUTE_PARAMS, 'jsonPayload') do |entry|
./test/plugin/base_test.rb:    verify_log_entries(1, COMPUTE_PARAMS, 'textPayload') do
./test/plugin/base_test.rb:    verify_log_entries(2, COMPUTE_PARAMS, 'jsonPayload') do |entry|
./test/plugin/base_test.rb:    verify_log_entries(6, COMPUTE_PARAMS, 'jsonPayload') do |entry|
./test/plugin/base_test.rb:      verify_log_entries(0, COMPUTE_PARAMS, 'jsonPayload')
./test/plugin/base_test.rb:      verify_log_entries(1, COMPUTE_PARAMS.merge(log_name: encoded_tag),
./test/plugin/base_test.rb:      verify_log_entries(1, COMPUTE_PARAMS.merge(log_name: sanitized_tag),
./test/plugin/base_test.rb:      verify_log_entries(log_entries_count, COMPUTE_PARAMS_WITH_MULTI_TAGS,
./test/plugin/base_test.rb:        verify_log_entries(emit_index, COMPUTE_PARAMS) do |entry, i|
./test/plugin/base_test.rb:    verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
./test/plugin/base_test.rb:    verify_log_entries(1, COMPUTE_PARAMS)
./test/plugin/base_test.rb:    # make a deep copy of COMPUTE_PARAMS and add the parsed label.
./test/plugin/base_test.rb:    params = Marshal.load(Marshal.dump(COMPUTE_PARAMS))
./test/plugin/base_test.rb:    # make a deep copy of COMPUTE_PARAMS and add the parsed label.
./test/plugin/base_test.rb:    params = Marshal.load(Marshal.dump(COMPUTE_PARAMS))
./test/plugin/base_test.rb:    # make a deep copy of COMPUTE_PARAMS and add the parsed label.
./test/plugin/base_test.rb:    params = Marshal.load(Marshal.dump(COMPUTE_PARAMS))
./test/plugin/base_test.rb:    # make a deep copy of COMPUTE_PARAMS and add the parsed labels.
./test/plugin/base_test.rb:    params = Marshal.load(Marshal.dump(COMPUTE_PARAMS))
./test/plugin/base_test.rb:      verify_log_entries(n, COMPUTE_PARAMS)
./test/plugin/base_test.rb:      verify_log_entries(1, COMPUTE_PARAMS, 'httpRequest') do |entry|
./test/plugin/base_test.rb:      verify_log_entries(1, COMPUTE_PARAMS, 'httpRequest') do |entry|
./test/plugin/base_test.rb:      verify_log_entries(1, COMPUTE_PARAMS, 'httpRequest') do |entry|
./test/plugin/base_test.rb:      verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
./test/plugin/base_test.rb:    verify_log_entries(1, COMPUTE_PARAMS, destination_key) do |entry|
./test/plugin/base_test.rb:    verify_log_entries(1, COMPUTE_PARAMS, destination_key) do |entry|
./test/plugin/base_test.rb:    verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
./test/plugin/constants.rb:  COMPUTE_PARAMS = {
./test/plugin/constants.rb:  COMPUTE_PARAMS_WITH_METADATA_VM_ID_AND_ZONE = COMPUTE_PARAMS.merge(
./test/plugin/constants.rb:    resource: COMPUTE_PARAMS[:resource].merge(
./test/plugin/constants.rb:  COMPUTE_PARAMS_WITH_MULTI_TAGS =
./test/plugin/constants.rb:    COMPUTE_PARAMS.reject { |k, _| k == :log_name }.freeze
./test/plugin/test_out_google_cloud.rb:    verify_log_entries(emit_index, COMPUTE_PARAMS) do |entry|
./test/plugin/test_out_google_cloud.rb:      verify_log_entries(1, COMPUTE_PARAMS) do |entry|
./test/plugin/test_out_google_cloud_grpc.rb:    verify_log_entries(emit_index, COMPUTE_PARAMS) do |entry|
./test/plugin/test_out_google_cloud_grpc.rb:    verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
./test/plugin/test_out_google_cloud_grpc.rb:      verify_log_entries(1, COMPUTE_PARAMS) do |entry|
```

test/plugin/base_test.rb
qingling128: Yeah, that bit could be confusing on a second thought. I've reverted this change.

test/plugin/constants.rb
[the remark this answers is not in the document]
igorpeshansky: I see. Yes, probably no point in switching names now, but you could at least redefine these in an additive way rather than use `reject`, i.e.:
```ruby
   COMPUTE_PARAMS_NO_LOG_NAME = {
     resource: {
       type: COMPUTE_CONSTANTS[:resource_type],
       labels: {
         'instance_id' => VM_ID,
         'zone' => ZONE
       }
     },
     project_id: PROJECT_ID,
     labels: {
     }
   }.freeze
   COMPUTE_PARAMS_WITH_METADATA_VM_ID_AND_ZONE = COMPUTE_PARAMS.merge(
     resource: COMPUTE_PARAMS[:resource].merge(
       labels: {
         'instance_id' => METADATA_VM_ID,
         'zone' => METADATA_ZONE
        }
      )
    ).freeze
    COMPUTE_PARAMS = COMPUTE_PARAMS_NO_LOG_NAME.merge(
      log_name: 'test'
    ).freeze
```

test/plugin/base_test.rb
```diff
@@ -605,6 +605,63 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+      setup_prometheus
+      setup_logging_stubs do
+        @logs_sent = []
+        d = create_driver(config + PROMETHEUS_ENABLE_CONFIG, 'test', true)
+        log_entries_count.times do |i|
+          d.emit("tag#{i}", 'message' => log_entry(i))
+        end
+        d.run
+      end
+      @logs_sent.each_with_index do |request, request_index|
```
igorpeshansky: Alternative:
```ruby
    @logs_sent.zip(request_log_names).each do |request, log_name|
      assert_equal log_name, request['logName']
    end
```

test/plugin/base_test.rb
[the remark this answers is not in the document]
igorpeshansky: Actually, even better, indexing past the end of the array gets you `nil`, so this could be just `[]`. WDYT?

test/plugin/base_test.rb
```diff
@@ -1609,21 +1667,24 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
         labels.merge!(entry['labels'] || {})
 
         assert_equal \
-          "projects/#{params[:project_id]}/logs/#{params[:log_name]}", log_name
+          "projects/#{params[:project_id]}/logs/#{params[:log_name]}",
+          log_name if params[:log_name]
```
igorpeshansky: I suspect this is what rubocop made you do, but personally I find this pretty unreadable -- the `if` applies to the whole `assert_equal` statement, but it *seems* that it only applies to `log_name` (making it effectively `assert_equal "...", nil` when `params[:log_name]` is `nil`).
It would be much more obvious as a conditional statement:
```ruby
        if params[:log_name]
          assert_equal \
            "projects/#{params[:project_id]}/logs/#{params[:log_name]}", log_name
        end
```
IMO, this is a good case for disabling this specific rubocop check for this line (`# rubocop:disable Style/GuardClause`).

test/plugin/base_test.rb
igorpeshansky: Hmm, I also wonder whether we could just make the `use_grpc` and `code` parameters to `assert_prometheus_metric_value` optional... That would allow you to get rid of these virtual methods. The values are really irrelevant to what you're trying to check.
One possible implementation:
```ruby
  def assert_prometheus_metric_value(metric_name, expected_value, labels = nil)
    metric = Prometheus::Client.registry.get(metric_name)
    assert_not_nil(metric)
    metric_value = if labels
                     metric.get(labels)
                   else
                     metric.values.values.reduce(0, :+)
                   end
    assert_equal(expected_value, metric_value)
  end
```
Alternatively, you can leave the `labels` default as `{}` and instead check for a special value, e.g., `:all` to aggregate.

test/plugin/base_test.rb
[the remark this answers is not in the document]

test/plugin/base_test.rb
qingling128: Changed. Rubocop did not complain about it when it multiple lines actually. :)

test/plugin/constants.rb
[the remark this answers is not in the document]
qingling128: Changed.

test/plugin/base_test.rb

test/plugin/base_test.rb
qingling128: Does that imply we don't check the `use_grpc` and `code` labels of the metrics anywhere? I thought we do want to verify them since that affects filtering.

I've created b/72162902 and added this note there. Will refactor and combine the test in a separate PR.

igorpeshansky: We will if they're relevant. In the one check you have in the base test, the label values are not -- the success code is always the same (modulo gRPC vs REST), and we assume that `use_grpc` will be set correctly.
I'm not suggesting that you change all calls to `assert_prometheus_metric_value` -- only the two that are in `base_test.rb`, to avoid introducing extra virtual methods `use_grpc` and `ok_status_code`.
FWIW, I felt that, since this PR actually introduces those virtual methods, it's easier to not do that rather than refactor later. WDYT?

test/plugin/base_test.rb
```diff
@@ -605,6 +605,64 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+  end
+
+  def test_split_logs_by_tag
+    setup_gce_metadata_stubs
+    log_entries_count = 5
+    dynamic_log_names = (0..log_entries_count - 1).map do |index|
+      "projects/test-project-id/logs/tag#{index}"
+    end
+    [
+      [APPLICATION_DEFAULT_CONFIG,
```
igorpeshansky: Modulo the comment, all of the elements fit on one line.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,64 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+       log_entries_count,
+       dynamic_log_names,
+       # An array that returns null for any index.
```
igorpeshansky: s/`null`/`nil`/.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,64 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+       []],
+      [DISABLE_SPLIT_LOGS_BY_TAG_CONFIG,
```
igorpeshansky: All of the elements fit on one line.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,64 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+    }.each do |(config, split_logs_by_tag)|
+      d = create_driver(config)
+      assert_equal split_logs_by_tag,
+                   d.instance.instance_variable_get(:@split_logs_by_tag)
+    end
+  end
+
+  def test_split_logs_by_tag
+    setup_gce_metadata_stubs
+    log_entries_count = 5
```
igorpeshansky: `log_entry_count`, for consistency.

test/plugin/base_test.rb

test/plugin/base_test.rb
qingling128: Oops. Changed.

test/plugin/base_test.rb

test/plugin/base_test.rb
qingling128: Cool cool. Yeah, I was also debating whether to check both explicitly or check the combined. I guess either works.

test/plugin/base_test.rb
qingling128: Ah, I see. Just made labels optional and removed the virtual methods. Will defer combining the prometheus tests till later.

qingling128: Per discussion offline, it makes more sense to explicitly specify `:all` when we want to sum up all metric values regardless of the labels. Just so we can distinguish cases between `get all metrics regardless of the labels` and `get all metrics with empty labels`. Just made the changes accordingly.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,56 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+      verify_log_entries(log_entry_count, COMPUTE_PARAMS_NO_LOG_NAME,
+                         'textPayload') do |entry, entry_index|
+        verify_default_log_entry_text(entry['textPayload'], entry_index,
+                                      entry)
+        assert_equal entry_log_names[entry_index], entry['logName']
+      end
+      # Verify the number of requests is different based on whether the
+      # 'split_logs_by_tag' flag is enabled.
+      assert_prometheus_metric_value(:stackdriver_successful_requests_count,
+                                     request_count, :all)
```
igorpeshansky: If you pass the sentinel positionally, rather than by name (i.e., `labels=:all`), then `:aggregate` might be a better name.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -607,17 +627,23 @@ def construct_log_entry_in_rest_format(labels,
       )
     end
 
-    def write_request_via_grpc(entries:, log_name:, resource:, labels:)
+    def write_request_via_grpc(entries:,
```
igorpeshansky: [Optional] If you rename this to just `write_request_grpc`, the header will fit on one line.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -705,7 +731,10 @@ def write_request_via_grpc(entries:, log_name:, resource:, labels:)
                  error: error.to_s
     end
 
-    def write_request_via_rest(entries:, log_name:, resource:, labels:)
+    def write_request_via_rest(entries:,
```
igorpeshansky: [Optional] If you rename this to just `write_request_rest`, the header will fit on one line.

test/plugin/base_test.rb
qingling128: Changed. PTAL.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Hmm... I found it easier to read with parameters on separate lines especially when some of them have default values.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Hmm... I found it easier to read with parameters on separate lines especially when some of them have default values.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,56 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+      "projects/test-project-id/logs/tag#{index}"
+    end
+    [
+      # [] returns nil for any index.
+      [APPLICATION_DEFAULT_CONFIG, log_entry_count, dynamic_log_names, []],
+      [DISABLE_SPLIT_LOGS_BY_TAG_CONFIG, 1, [''], dynamic_log_names]
+    ].each do |(config, request_count, request_log_names, entry_log_names)|
+      setup_prometheus
+      setup_logging_stubs do
+        @logs_sent = []
```
igorpeshansky: Isn't this what `new_stub_context` is for? But this isn't the only place that needs this change, so let's defer to a later PR.

test/plugin/base_test.rb
```diff
@@ -605,6 +605,56 @@ def test_sanitize_tags_from_container_name_with_require_valid_tags_false
...
+      DISABLE_SPLIT_LOGS_BY_TAG_CONFIG => false
+    }.each do |(config, split_logs_by_tag)|
+      d = create_driver(config)
+      assert_equal split_logs_by_tag,
+                   d.instance.instance_variable_get(:@split_logs_by_tag)
+    end
+  end
+
+  def test_split_logs_by_tag
+    setup_gce_metadata_stubs
```
igorpeshansky: This should probably be within the loop... Feel free to defer until you do the `new_stub_context` cleanup.

test/plugin/base_test.rb
qingling128: Will defer to a follow up PR.

test/plugin/base_test.rb
qingling128: Will defer to a follow up PR.

PR #203, opened 2018-01-16
Reuse gRPC stub for Logging API calls.

qingling128:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1817,6 +1807,19 @@ def api_client
...
+        @client.authorization = Google::Auth.get_application_default(
+          LOGGING_SCOPE)
+      end
+    end
+
+    def api_client
+      # For gRPC side, the Channel will take care of tokens and their renewal.
+      # https://grpc.io/docs/guides/auth.html#authentication-api
+      unless @use_grpc
         unless @client.authorization.expired?
```
igorpeshansky: I strongly believe that this is actually a zero-day bug (introduced in https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/commit/17e75cea415902f5dc8344a4c32c688b818f65cb).
Reading the code, this should be `if`, not `unless`.
This could also potentially be the cause of #197...
So, if we use the boolean operation, as suggested above, let's change it to:
```ruby
if not @use_grpc and @client.authorization.expired?
```
As this is unrelated, we should probably pull that out into a separate (one-line) PR, but then I'd suggest merging that PR first.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1817,6 +1807,19 @@ def api_client
...
+        Google::Apis::ClientOptions.default.application_version = PLUGIN_VERSION
+        @client = Google::Apis::LoggingV2::LoggingService.new
+        @client.authorization = Google::Auth.get_application_default(
+          LOGGING_SCOPE)
+      end
+    end
+
+    def api_client
+      # For gRPC side, the Channel will take care of tokens and their renewal.
+      # https://grpc.io/docs/guides/auth.html#authentication-api
```
igorpeshansky: Suggestion:
```ruby
      # For gRPC side, the Channel will take care of tokens and their renewal
      # (https://grpc.io/docs/guides/auth.html#authentication-api).
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1817,6 +1807,19 @@ def api_client
...
+      unless @use_grpc
```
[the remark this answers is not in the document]
qingling128: Combined. But I changed it to `if !@use_grpc && @client.authorization.expired?` instead to make it a bit more natural to read.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Changed `unless` to `if` in https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/204.

qingling128:
Rebased after https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/204 was merged. PTAL.

PR #204, opened 2018-01-17
Fix API client re-authentication after token expiration.

qingling128:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1816,16 +1816,14 @@ def api_client
...
-            raise Google::APIClient::ClientError, 'Unable to fetch access ' \
-              'token (no scopes configured?)'
-          end
+      elsif @client.authorization.expired?
+        begin
+          @client.authorization.fetch_access_token!
+        rescue MultiJson::ParseError
+          # Workaround an issue in the API client; just re-raise a more
+          # descriptive error for the user (which will still cause a retry).
+          raise Google::APIClient::ClientError, 'Unable to fetch access ' \
```
igorpeshansky: [Optional] While we're at it, should we recombine this error to make it easier to search for the whole error string, and just put the whole thing on the next line?

PR #205, opened 2018-01-19
Restore #180 and remove google-cloud-logging and grpc pins.

qingling128:
This is a revert of #196 , which reverts #180 . 

BTW, there is an issue that I observed after the upgrade of `google-cloud-logging`: https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1938. It does not seem to be a blocking thing though. We are not getting that warning in Fluentd logs.


draffensperger:
Thanks for doing this!

qingling128:
@draffensperger - You're more than welcome. :D

igorpeshansky:
The deprecation warning should be addressed by #206.

PR #206, opened 2018-01-25
Fix use of deprecated parameter name.

igorpeshansky:
This should address https://github.com/GoogleCloudPlatform/google-cloud-ruby/issues/1938.


igorpeshansky:
Yeah, this warning showed up in previous Travis builds, so it was easy to see that it got fixed with this PR.

PR #209, opened 2018-01-29
Use the project_id field in the credentials if available.

igorpeshansky:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -38,6 +38,27 @@ class Map
...
+module Google
+  module Auth
+    # Extract project_id in initialize.
+    class ServiceAccountCredentials
+      singleton_class.send(:alias_method, :super_make_creds, :make_creds)
+      def self.make_creds(options = {})
+        json_key_io, scope = options.values_at(:json_key_io, :scope)
+        if json_key_io
+          json_key = MultiJson.load(json_key_io.read)
+          project_id = json_key['project_id']
```
qingling128: Do we need to check if `project_id` exists first? Just in case some corrupted credential files are passed in. It would be nice to log a warning and clear instructions to fix it instead of crashing the agent.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1393,6 +1414,9 @@ module CredentialsInfo
       # Returns the project ID (as a string) on success, or nil on failure.
       def self.project_id
         creds = Google::Auth.get_application_default(LOGGING_SCOPE)
+        if creds.respond_to?(:project_id)
+          return creds.project_id if creds.project_id
+        end
```
qingling128: Should we add a unit test to test the fallback when `project_id` is not present?

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Ruby hashes return `nil` if the key is not present.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Our existing tests do exactly that, don't they?

lib/fluent/plugin/out_google_cloud.rb
qingling128: Yeah, Ruby does that. Yet I meant something that explicitly indicates the json key is corrupted when `project_id` is not present. Better to error out loudly, right? Kind of similar to what googleauth does [here](https://github.com/google/google-auth-library-ruby/blob/master/lib/googleauth/service_account.rb#L74):

```
        json_key = MultiJson.load(json_key_io.read)
        raise 'missing client_email' unless json_key.key?('client_email')
        raise 'missing private_key' unless json_key.key?('private_key')
```

lib/fluent/plugin/out_google_cloud.rb
qingling128: Our existing test `test_project_id_from_credentials` test three types of credentials independently. Here we seem to have some order/precedence as well: 1. `creds.project_id`, 2. `creds.issuer`, 3. `creds.client_id`. Not directly related to this PR though. We can defer it till later.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: The `project_id` field is optional, so we probably don't want to error out. Setting it to `nil` will default to extracting it out of the email address, which is what we want.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Yes, the test data for the newly added test (`new-style-credentials.json`) has a different project encoded in `client_email`, so we *are* testing that order of precedence. As for `issuer` vs `client_id`, I believe the two refer to the same field (but `issuer` is the newer name for it).

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -38,6 +38,27 @@ class Map
   end
 end
 
+module Google
+  module Auth
+    # Extract project_id in initialize.
+    class ServiceAccountCredentials
+      singleton_class.send(:alias_method, :super_make_creds, :make_creds)
+      def self.make_creds(options = {})
+        json_key_io, scope = options.values_at(:json_key_io, :scope)
```
bmoyles0117: I'm confused when you use `json_key_io` as the variable name. Is this value still a "key" at this point, or would it be better named as `json_io`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -38,6 +38,27 @@ class Map
...
+        if json_key_io
+          json_key = MultiJson.load(json_key_io.read)
```
bmoyles0117: Similar comment applied here, I'm just trying to understand why we're using the world "key" here, would it be fair to rename this to `json_credentials`?

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: This is the IO object that will give us the JSON form of a private key. Actually, the naming comes directly from [google-auth-library-ruby](https://github.com/google/google-auth-library-ruby/blob/master/lib/googleauth/service_account.rb#L56), and I'd like to stay as close as possible to that code.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Same here, this is the JSON form of the private key (see the above code reference).

PR #210, opened 2018-01-31
Turned on partial_success by default and removed obsolete comment of support availability.

qingling128:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -307,8 +307,7 @@ module InternalConstants
 
     # Whether valid entries should be written even if some other entries fail
     # due to INVALID_ARGUMENT or PERMISSION_DENIED errors when communicating to
-    # the Stackdriver Logging API. This is highly recommended. Right now this
-    # only works with the REST path (use_grpc = false).
+    # the Stackdriver Logging API. This is highly recommended.
     config_param :partial_success, :bool, :default => false
```
igorpeshansky: Would it make sense to also flip the default to `true`? This shouldn't break existing customers, right?

qingling128:
Adjusted unit tests. PTAL.

qingling128:
Created b/72756346 to enhance the unit tests.

PR #212, opened 2018-02-23
Support passing metadata agent url via an environment variable STACKDRIVER_METADATA_AGENT_URL.

qingling128:
We need to determine "best practices" on the topic of setting a dynamic metadata_agent_url which is dependent on the environment the logging agent is running in. Having the default set to "http://local-metadata-agent.stackdriver.com:8000" only works well when the metadata agent is discoverable on the same network as the logging agent. 

The decision is to support a new configuration parameter that specifies the environment variable `STACKDRIVER_METADATA_AGENT_URL`, which is evaluated after checking the metadata_agent_url before falling back to default.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -20,7 +20,7 @@ Metrics/AbcSize:
 Metrics/ClassLength:
   Max: 1300
 Metrics/ModuleLength:
-  Max: 1600
+  Max: 1700
 
 # Offense count: 3
 Metrics/CyclomaticComplexity:
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -352,12 +352,13 @@ module Fluent
     # Whether to call metadata agent to retrieve monitored resource.
     config_param :enable_metadata_agent, :bool, :default => false
 
-    # The url of Metadata Agent. If this config is not customized (left as
-    # default instead) while an environment variable
-    # STACKDRIVER_METADATA_AGENT_URL is set, the value of that variable will
-    # take precedence. Otherwise the value of this config will be used.
-    config_param :metadata_agent_url, :string,
-                 :default => DEFAULT_METADATA_AGENT_URL
+    # The URL of Metadata Agent.
+    # If this option is not nil, its value will be used.
+    # If this option is nil, we will first check if the environment variable
+    # STACKDRIVER_METADATA_AGENT_URL is set and use that if possible.
+    # If that environment variable is also absent, the default url
+    # 'http://local-metadata-agent.stackdriver.com:8000' will be used.
+    config_param :metadata_agent_url, :string, :default => nil
 
     # Whether to split log entries with different log tags into different
     # requests when talking to Stackdriver Logging API.
@@ -393,15 +394,18 @@ module Fluent
                   ' ignored.'
       end
 
+      # 1. If @metadata_agent_url is customized (aka not nil), use that.
+      # 2. Otherwise check the presence of the environment variable
+      #    STACKDRIVER_METADATA_AGENT_URL and use that if set.
+      # 3. Fall back to the default if neither is set.
       if @enable_metadata_agent
-        # 1. If @metadata_agent_url is customized, use that.
-        # 2. Otherwise check the environment variable
-        #    STACKDRIVER_METADATA_AGENT_URL.
-        # 3. Fall back to the default
-        if @metadata_agent_url == DEFAULT_METADATA_AGENT_URL && \
-           !ENV[METADATA_AGENT_URL_ENV_VAR_NAME].to_s.empty?
-          @metadata_agent_url = ENV[METADATA_AGENT_URL_ENV_VAR_NAME]
-        end
+        # Convert to string to capture empty string.
+        @metadata_agent_url ||=
+          if ENV[METADATA_AGENT_URL_ENV_VAR_NAME].to_s.empty?
+            DEFAULT_METADATA_AGENT_URL
+          else
+            ENV[METADATA_AGENT_URL_ENV_VAR_NAME]
+          end
       end
 
       # If monitoring is enabled, register metrics in the default registry
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -124,6 +124,7 @@ module Fluent
 
       DEFAULT_METADATA_AGENT_URL =
         'http://local-metadata-agent.stackdriver.com:8000'.freeze
+      METADATA_AGENT_URL_ENV_VAR_NAME = 'STACKDRIVER_METADATA_AGENT_URL'.freeze
     end
 
     # Internal constants.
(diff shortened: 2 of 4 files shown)


.rubocop_todo.yml
```diff
@@ -20,7 +20,7 @@ Metrics/AbcSize:
 Metrics/ClassLength:
   Max: 1300
 Metrics/ModuleLength:
-  Max: 1600
+  Max: 1700
```
qingling128: One day we will split the files. One day... Before that this number is going to keep increasing as we add more tests to the `base_test.rb` file.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -387,6 +388,17 @@ def configure(conf)
                   ' ignored.'
       end
 
+      if @enable_metadata_agent
+        # 1. If @metadata_agent_url is customized, use that.
+        # 2. Otherwise check the environment variable
+        #    STACKDRIVER_METADATA_AGENT_URL.
+        # 3. Fall back to the default
+        if @metadata_agent_url == DEFAULT_METADATA_AGENT_URL && \
+           !ENV[METADATA_AGENT_URL_ENV_VAR_NAME].to_s.empty?
```
qingling128: Convert the env variable to string to capture empty strings (in which case it should be considered not set).

test/plugin/base_test.rb
```diff
@@ -116,6 +116,26 @@ def test_configure_partial_success
     end
   end
 
+  def test_metadata_agent_url_customization
+    [
+      # If @metadata_agent_url is customized, use that even if the environment
+      # variable is set.
+      [CUSTOM_METADATA_AGENT_URL_CONFIG, true, CUSTOM_METADATA_AGENT_URL],
```
igorpeshansky: We don't seem to be testing the "config without environment variable" case. Might as well add it, just to rule out an inverted conditional...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -387,6 +388,17 @@ def configure(conf)
                   ' ignored.'
       end
 
+      if @enable_metadata_agent
+        # 1. If @metadata_agent_url is customized, use that.
+        # 2. Otherwise check the environment variable
```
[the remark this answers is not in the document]
qingling128: Updated.

test/plugin/base_test.rb

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -350,6 +351,11 @@ module InternalConstants
 
     # Whether to call metadata agent to retrieve monitored resource.
     config_param :enable_metadata_agent, :bool, :default => false
+
+    # The url of Metadata Agent. If this config is not customized (left as
+    # default instead) while an environment variable
+    # STACKDRIVER_METADATA_AGENT_URL is set, the value of that variable will
+    # take precedence. Otherwise the value of this config will be used.
```
igorpeshansky: The problem with this approach is that there's no way to distinguish the unset option (where the environment variable should take precedence) and the option that's explicitly set to the default (where the user wants to override the environment variable).
Perhaps we should default this to `nil` and set it to the default value after examining the environment variable if it stays a `nil`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -350,6 +351,11 @@ module InternalConstants
 
     # Whether to call metadata agent to retrieve monitored resource.
     config_param :enable_metadata_agent, :bool, :default => false
+
+    # The url of Metadata Agent. If this config is not customized (left as
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
qingling128: Makes sense. Changed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -350,8 +351,14 @@ module InternalConstants
 
     # Whether to call metadata agent to retrieve monitored resource.
     config_param :enable_metadata_agent, :bool, :default => false
-    config_param :metadata_agent_url, :string,
-                 :default => DEFAULT_METADATA_AGENT_URL
+
+    # The URL of Metadata Agent.
+    # If this option is not nil, its value will be used.
```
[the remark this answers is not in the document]
igorpeshansky: This seems overly verbose. How about something like:
```ruby
# If this option is set, its value is used to contact the Metadata Agent. Otherwise, the value of the STACKDRIVER_METADATA_AGENT_URL environment variable is used. If that is also unset, this defaults to 'http://local-metadata-agent.stackdriver.com:8000'.
```
?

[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -350,8 +351,14 @@ module InternalConstants
 
     # Whether to call metadata agent to retrieve monitored resource.
     config_param :enable_metadata_agent, :bool, :default => false
-    config_param :metadata_agent_url, :string,
-                 :default => DEFAULT_METADATA_AGENT_URL
+
+    # The URL of Metadata Agent.
```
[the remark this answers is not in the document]

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -387,6 +393,20 @@ def configure(conf)
...
       end
 
+      # 1. If @metadata_agent_url is customized (aka not nil), use that.
+      # 2. Otherwise check the presence of the environment variable
+      #    STACKDRIVER_METADATA_AGENT_URL and use that if set.
+      # 3. Fall back to the default if neither is set.
+      if @enable_metadata_agent
+        # Convert to string to capture empty string.
+        @metadata_agent_url ||=
+          if ENV[METADATA_AGENT_URL_ENV_VAR_NAME].to_s.empty?
```
igorpeshansky: Just curious, could this also be:
```ruby
@metadata_agent_url ||= ENV[METADATA_AGENT_URL_ENV_VAR_NAME]
@metadata_agent_url ||= DEFAULT_METADATA_AGENT_URL
```
? Or would it miss some cases?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -124,6 +124,7 @@ module ConfigConstants
 
       DEFAULT_METADATA_AGENT_URL =
         'http://local-metadata-agent.stackdriver.com:8000'.freeze
+      METADATA_AGENT_URL_ENV_VAR_NAME = 'STACKDRIVER_METADATA_AGENT_URL'.freeze
```
igorpeshansky: [Optional] The `_NAME` suffix seems redundant.

lib/fluent/plugin/out_google_cloud.rb
qingling128: This way it will miss the empty string case. If `ENV[METADATA_AGENT_URL_ENV_VAR_NAME]` is an empty string, `@metadata_agent_url` will end up being an empty string.

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Got it. I thought we didn't really need the empty string case, but then realized that it allows one to invoke the agent as `STACKDRIVER_METADATA_AGENT_URL= metadatad` (although it's also possible to unset the variable completely via `/usr/bin/env -u STACKDRIVER_METADATA_AGENT_URL metadatad`). I'm ok with leaving it.

PR #213, opened 2018-02-23
Remove 'Name' from docker and gke local_resource_id prefixes.

qingling128:
This is to align with https://github.com/Stackdriver/metadata-agent/pull/56.

The change here does not affect functionality, just the test stubs.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1182,8 +1182,8 @@ module Fluent
       # Metadata Agent with this key.
       #
       # Examples:
-      #   "container.<container_id>"
-      #   "k8s_pod.<cluster_name>.<namespace_name>.<pod_name>.<location>"
+      # "container.<container_id>" // Docker container.
+      # "k8s_pod.<namespace_name>.<pod_name>" // GKE pod.
       if @enable_metadata_agent && local_resource_id
         @log.debug 'Calling metadata agent with local_resource_id: ' \
                   "#{local_resource_id}."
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1181,11 +1181,9 @@ def determine_group_level_monitored_resource_and_labels(tag,
       # inferred from the tag) if failed to get a monitored resource from
       # Metadata Agent with this key.
       #
-      # Docker container:
```
igorpeshansky: I'd keep the Docker and Kubernetes headers. Otherwise it looks like container.* represents a Kubernetes container...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1181,11 +1181,9 @@ def determine_group_level_monitored_resource_and_labels(tag,
...
+      # Examples:
       #   "container.<container_id>"
-      #   "containerName.<container_name>"
-      # GKE container:
-      #   "gke_containerName.<namespace_id>.<pod_name>.<container_name>"
+      #   "k8s_pod.<cluster_name>.<namespace_name>.<pod_name>.<location>"
```
igorpeshansky: Location?

qingling128: Fixed. I was looking at monitored resource...

lib/fluent/plugin/out_google_cloud.rb
qingling128: Added Docker and Kubernete as inline comments.

PR #214, opened 2018-03-08
Disable split_logs_by_tag by default and fix a corner case when no valid entry is present.

qingling128:
This option controls whether Stackdriver Logging Agent will split logs into separate requests based on the log tag. If true, logs with each distinct tag will be grouped into their own request. By default this is currently enabled. The original intention behind splitting logs with different tags into their own requests is to avoid repeatedly setting the log name for every log entry in the batch send to Stackdriver Logging, compressing the payload size.

While you might expect that the effect of the optimization above would be minimal, we’ve seen significant performance discrepancies when the number of distinct log tags > 30. In other words, what could all fit into one request get split into >30 requests. This substantially increases latency and consumes significantly more resource usage.

On the other side, we are not seeing obvious performance improvements when tags are < 30. And logging backend folks also suggested that this type of optimization we are doing is not longer very relevant.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -382,7 +382,7 @@ module Fluent
 
     # Whether to split log entries with different log tags into different
     # requests when talking to Stackdriver Logging API.
-    config_param :split_logs_by_tag, :bool, :default => true
+    config_param :split_logs_by_tag, :bool, :default => false
 
     # rubocop:enable Style/HashSyntax
 
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -632,7 +632,8 @@ module Fluent
           end
           combined_entries.concat(request[:entries])
         end
-        @write_request.call(entries: combined_entries)
+        @write_request.call(entries: combined_entries) unless
+          combined_entries.empty?
       end
     end
 
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -658,7 +658,7 @@ module BaseTest
     setup_gce_metadata_stubs
     {
       APPLICATION_DEFAULT_CONFIG => false,
-      SPLIT_LOGS_BY_TAG_ENABLE_CONFIG => true
+      ENABLE_SPLIT_LOGS_BY_TAG_CONFIG => true
     }.each do |(config, split_logs_by_tag)|
       d = create_driver(config)
       assert_equal split_logs_by_tag,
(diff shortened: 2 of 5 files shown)


test/plugin/base_test.rb
```diff
@@ -674,8 +674,8 @@ def test_split_logs_by_tag
     end
     [
       # [] returns nil for any index.
-      [APPLICATION_DEFAULT_CONFIG, log_entry_count, dynamic_log_names, []],
-      [DISABLE_SPLIT_LOGS_BY_TAG_CONFIG, 1, [''], dynamic_log_names]
+      [SPLIT_LOGS_BY_TAG_ENABLE_CONFIG, log_entry_count, dynamic_log_names, []],
```
bmoyles0117: Nit picky, but can we please flip these lines so their ordering matches the stub definition? If not, that's fine, I just found myself having a hard time reasoning about this change when the order was flipped.

qingling128: Sure thing. Flipped.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -632,7 +632,8 @@ def write(chunk)
           end
           combined_entries.concat(request[:entries])
         end
-        @write_request.call(entries: combined_entries)
+        @write_request.call(entries: combined_entries) unless
```
qingling128: If all entries in this chunk is invalid and dropped, we should not make a request. This fix is to make the behavior consistent with how we deal with it when `split_logs_by_tag` is true.

Our unit test `test_malformed_log` actually caught this.

test/plugin/constants.rb
```diff
@@ -186,8 +186,8 @@ module Constants
     detect_subservice false
   ).freeze
 
-  DISABLE_SPLIT_LOGS_BY_TAG_CONFIG = %(
-    split_logs_by_tag false
+  SPLIT_LOGS_BY_TAG_ENABLE_CONFIG = %(
```
igorpeshansky: Hmm, the other `ENABLE` config constants are inconsistent in whether `ENABLE_` is a prefix or not. We should probably make them all use it as a prefix...

qingling128: I noticed this as well. Seems the only one that does not follow this pattern is `ENABLE_METADATA_AGENT_CONFIG` though. And that config itself is called `enable_metadata_agent`. So I left it unchanged. Do we want to change the config name to `ENABLE_METADATA_AGENT_ENABLE_CONFIG` or `ENABLE_METADATA_AGENT_TRUE_CONFIG`?

igorpeshansky: I think `ENABLE_METADATA_AGENT_CONFIG` and `DISABLE_METADATA_AGENT_CONFIG` are reasonable here. So I would advocate for `ENABLE_SPLIT_LOGS_BY_TAG_CONFIG`, `ENABLE_PROMETHEUS_CONFIG`, etc.

qingling128: Updated.

PR #222, opened 2018-05-07
Allow overriding k8s_cluster_name and k8s_cluster_location in the config.

### LINKED ISSUE #221: Support for on-prem Kubernetes clusters
Would it be possible to leverage Stackdriver for monitoring on-prem clusters? This would help users have a unified monitoring and logging solution for all their Kubernetes installations (on-prem and cloud) and provide a way to migrate on-prem clusters to the cloud without changing the dev tooling experience.

igorpeshansky:
Fixes #221.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -260,7 +260,7 @@ module Fluent
     config_param :vm_name, :string, :default => nil
     # Kubernetes-specific parameters.
     config_param :k8s_cluster_name, :string, :default => nil
-    config_param :k8s_location, :string, :default => nil
+    config_param :k8s_cluster_location, :string, :default => nil
 
     # Map keys from a JSON payload to corresponding LogEntry fields.
     config_param :http_request_key, :string, :default =>
@@ -2184,7 +2184,7 @@ module Fluent
       begin
         @k8s_cluster_name ||= fetch_gce_metadata(
           'instance/attributes/cluster-name')
-        @k8s_location ||= fetch_gce_metadata(
+        @k8s_cluster_location ||= fetch_gce_metadata(
           'instance/attributes/cluster-location')
       rescue StandardError => e
         @log.error 'Failed to retrieve k8s cluster name and location.', \
@@ -2197,18 +2197,18 @@ module Fluent
           'pod_name' => pod_name,
           'container_name' => container_name,
           'cluster_name' => @k8s_cluster_name,
-          'location' => @k8s_location
+          'location' => @k8s_cluster_location
         }
         fallback_resource = GKE_CONSTANTS[:resource_type]
       when K8S_NODE_CONSTANTS[:resource_type]
         labels = {
           'node_name' => node_name,
           'cluster_name' => @k8s_cluster_name,
-          'location' => @k8s_location
+          'location' => @k8s_cluster_location
         }
         fallback_resource = COMPUTE_CONSTANTS[:resource_type]
       end
-      unless @k8s_cluster_name && @k8s_location
+      unless @k8s_cluster_name && @k8s_cluster_location
         @log.error "Failed to construct #{resource_type} resource locally." \
                    ' Falling back to writing logs against' \
                    " #{fallback_resource} resource.", error: e
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -258,7 +258,9 @@ module Fluent
     config_param :zone, :string, :default => nil
     config_param :vm_id, :string, :default => nil
     config_param :vm_name, :string, :default => nil
-    # Kubernetes-specific parameters.
+    # Kubernetes-specific parameters, only used to override these values in
+    # the fallback path when the metadata agent is temporarily unavailable.
+    # They have to match the configuration of the metadata agent.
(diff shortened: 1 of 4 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -258,6 +258,9 @@ module InternalConstants
     config_param :zone, :string, :default => nil
     config_param :vm_id, :string, :default => nil
     config_param :vm_name, :string, :default => nil
+    # Kubernetes-specific parameters.
+    config_param :k8s_cluster_name, :string, :default => nil
```
qingling128: My only user experience concern:

With the current implementation, it requires users to set `enable_metadata_agent` to true in Logging Agent, but not install metadata agent (or somehow makes sure metadata agent does not return anything given that `local_resource_id`). So that the customized `k8s_cluster_name` and `k8s_location` would take effect. Would that be a bit confusing?

Alternatively, we could detect that `k8s_cluster_name` and `k8s_location` are set and execute specific logic regardless of the value of `enable_metadata_agent`. Something like:

```
      if local_resource_id
        if @enable_metadata_agent
          # Exactly what we were doing before here.
        elsif @k8s_cluster_name && @k8s_location
          constructed_k8s_resource = construct_k8s_resource_locally(
            local_resource_id)
          resource = constructed_k8s_resource if constructed_k8s_resource
        end
      end
```

Thoughts?

qingling128: Per clarification offline, the expected way of doing this is always installing Metadata Agent and enabling the communication from Logging Agent to Metadata Agent. Metadata Agent has a similar mechanism to specify @k8s_cluster_name and @k8s_location. This change is just for fallbacks. So the current implementation makes sense.

qingling128: Can we enrich the comment of this parameter a bit (same with `k8s_location`)? Something like:

`# Customized Kubernetes cluster name. Stackdriver Metadata Agent and Logging Agent detect Kubernetes cluster names in GKE automatically. For other platforms this has to be specified explicitly.`

?

test/plugin/base_test.rb
```diff
@@ -1396,10 +1396,10 @@ def test_k8s_monitored_resource_fallback
       },
       {
         enable_metadata_agent: true,
-        setup_metadata_agent_stub: true,
+        setup_metadata_agent_stub: false,
```
qingling128: Do we need to remove the previous test? We could just add a test case, right?

test/plugin/base_test.rb
```diff
@@ -1442,6 +1442,41 @@ def test_k8s_monitored_resource_fallback
     end
   end
 
+  # Test k8s monitored resource fallback when Metadata Agent restarts
+  # without the metadata server.
+  def test_k8s_monitored_resource_fallback_custom
```
qingling128: We could also refactor `test_k8s_monitored_resource_fallback` to pass in `config: <ACTUAL_CONFIG>` directly instead of `enable_metadata_agent: true/false` and then do:
```
        config: ENABLE_METADATA_AGENT_CONFIG,
        setup_metadata_agent_stub: false,
        setup_k8s_stub: false,
        log_entry: k8s_node_log_entry(log_entry(0)),
        expected_params: K8S_NODE_PARAMS
```
```
        config: CUSTOM_K8S_ENABLE_METADATA_AGENT_CONFIG,
        setup_metadata_agent_stub: false,
        setup_k8s_stub: false,
        log_entry: k8s_container_log_entry(log_entry(0)),
        expected_params: K8S_CONTAINER_PARAMS_CUSTOM
```
```
        config: CUSTOM_K8S_ENABLE_METADATA_AGENT_CONFIG,
        setup_metadata_agent_stub: false,
        setup_k8s_stub: false,
        log_entry: k8s_node_log_entry(log_entry(0)),
        expected_params: K8S_NODE_PARAMS_CUSTOM
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -258,6 +258,9 @@ module InternalConstants
...
+    config_param :k8s_location, :string, :default => nil
```
bmoyles0117: We should add a comment saying that this is the master's location, or whatever it actually is, to ensure they're not setting this to be the node's location.

test/plugin/constants.rb
```diff
@@ -441,6 +450,14 @@ module Constants
       )
     )
   ).freeze
+  K8S_CONTAINER_PARAMS_CUSTOM = K8S_CONTAINER_PARAMS.merge(
```
bmoyles0117: Where is this actually being used?

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: I've added something similar without mentioning GKE and other platforms. PTAL.

test/plugin/constants.rb
bmoyles0117: For whatever reason, searching the page previously only showed one instance of this variable. I see it now.

igorpeshansky: It was a separate file — the diff could have been collapsed by GitHub.

test/plugin/base_test.rb
qingling128: Ah, I see. Did not realize that we had identical tests. Thanks for cleaning that up.

PR #223, opened 2018-05-10
Don't try to parse an error from the metadata agent as JSON.

igorpeshansky:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1451,6 +1451,7 @@ def query_metadata_agent(path)
       end
     rescue StandardError => e
       @log.error "Error calling Metadata Agent at #{url}.", error: e
+      nil
```
qingling128: I thought this should return `nil` already implicitly.

```
$ irb
2.4.3 :001 > def some_method
  raise StandardError('expected')
rescue StandardError => e
  puts "wow"
end
 => :some_method
2.4.3 :006 > some_method()
wow
 => nil
```

What is the log spam about?

igorpeshansky: Unlike `puts`, `@log.error` returns the message. We were getting log spam like:
```
A  2018-05-10 14:22:49 +0000 [error]: Error parsing monitored resource from Metadata Agent. response: "2018-05-10 14:22:49 +0000 [error]: Error calling Metadata Agent at http://<redacted>:8799/monitoredResource/k8s_container.kube-system.fluentd-gcp-v3.0.0-z5wwz.fluentd-gcp. error=#<SocketError: Failed to open TCP connection to <redacted>:8799 (getaddrinfo: Name or service not known)>" error=#<Google::Protobuf::ParseError: Error occurred during parsing: Parse error at '"2018-05-10 14:22:49 +0000 [error]: Error calling Metadata Agent at http://<redacted>... >
```
which indicates that `query_metadata_agent_for_monitored_resource` was calling `decode_json` with the logged error.

qingling128: ugh...... I see. Good to know.

PR #226, opened 2018-06-15
Unset the credential file path environment variable if the value is an empty string.

qingling128:
This way the googleauth lib knows to fetch the credentials following the fallback path.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -156,6 +156,7 @@ module Fluent
 
     # Internal constants.
     module InternalConstants
+      CREDENTIAL_PATH_VAR = 'GOOGLE_APPLICATION_CREDENTIALS'.freeze
       DEFAULT_LOGGING_API_URL = 'https://logging.googleapis.com'.freeze
 
       # The label name of local_resource_id in the json payload. When a record
@@ -481,6 +482,10 @@ module Fluent
       set_regexp_patterns
 
       @platform = detect_platform
+
+      # Unset the credential file path environment variable if the value is set
+      # to an empty string.
+      ENV.delete(CREDENTIAL_PATH_VAR) if ENV[CREDENTIAL_PATH_VAR] == ''
 
       # Set required variables: @project_id, @vm_id, @vm_name and @zone.
       set_required_metadata_variables
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -156,7 +156,7 @@ module Fluent
 
     # Internal constants.
     module InternalConstants
-      CREDENTIAL_PATH_VAR = 'GOOGLE_APPLICATION_CREDENTIALS'.freeze
+      CREDENTIALS_PATH_ENV_VAR = 'GOOGLE_APPLICATION_CREDENTIALS'.freeze
       DEFAULT_LOGGING_API_URL = 'https://logging.googleapis.com'.freeze
 
       # The label name of local_resource_id in the json payload. When a record
@@ -483,9 +483,10 @@ module Fluent
 
       @platform = detect_platform
 
-      # Unset the credential file path environment variable if the value is set
-      # to an empty string.
-      ENV.delete(CREDENTIAL_PATH_VAR) if ENV[CREDENTIAL_PATH_VAR] == ''
+      # Treat an empty setting of the credentials file path environment variable
+      # as unset. This way the googleauth lib could fetch the credentials
+      # following the fallback path.
+      ENV.delete(CREDENTIALS_PATH_ENV_VAR) if ENV[CREDENTIALS_PATH_ENV_VAR] == ''
 
       # Set required variables: @project_id, @vm_id, @vm_name and @zone.
       set_required_metadata_variables
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -486,7 +486,8 @@ module Fluent
       # Treat an empty setting of the credentials file path environment variable
       # as unset. This way the googleauth lib could fetch the credentials
       # following the fallback path.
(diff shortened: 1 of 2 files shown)


test/plugin/base_test.rb
```diff
@@ -274,6 +274,13 @@ def test_invalid_json_credentials
     end
   end
 
+  def test_unset_empty_credential_env_var
+    ENV[CREDENTIAL_PATH_VAR] = ''
+    setup_gce_metadata_stubs
+    create_driver
+    assert_nil ENV[CREDENTIAL_PATH_VAR]
+  end
```
bmoyles0117: Can we add another test for when the environment variable is not present, just to ensure that we test both paths?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -156,6 +156,7 @@ module ConfigConstants
 
     # Internal constants.
     module InternalConstants
+      CREDENTIAL_PATH_VAR = 'GOOGLE_APPLICATION_CREDENTIALS'.freeze
```
igorpeshansky: Why have a named constant? Our test is full of literal uses of this value.
If you do choose to keep it, `CREDENTIALS_PATH_ENV_VAR`, for consistency.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -482,6 +483,10 @@ def configure(conf)
 
       @platform = detect_platform
 
+      # Unset the credential file path environment variable if the value is set
+      # to an empty string.
```
igorpeshansky: Might want to explain why instead (e.g., `Treat an empty setting of the credentials file path environment variable as unset`).

test/plugin/base_test.rb
```diff
@@ -274,6 +274,13 @@ def test_invalid_json_credentials
     end
   end
 
+  def test_unset_empty_credential_env_var
+    ENV[CREDENTIAL_PATH_VAR] = ''
```
igorpeshansky: If you decide to keep the constant, should we change all uses of `'GOOGLE_APPLICATION_CREDENTIALS'` in this file to use this constant?

test/plugin/base_test.rb
igorpeshansky: +1.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Because... I sometimes missed the `S` when spelling it. Just wanna reduce such risks. I've renamed it to `CREDENTIALS_PATH_ENV_VAR` instead. Also refactored the tests.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Changed. PTAL.

test/plugin/base_test.rb
qingling128: Yup. Done.

test/plugin/base_test.rb
qingling128: I thought about it. But that has been covered by a dozens of 'regular' tests that do not explicitly set this var. Because we delete the environment in `setup` for all tests:
```
  def setup
    Fluent::Test.setup
    # delete environment variables that googleauth uses to find credentials.
    ENV.delete(CREDENTIALS_PATH_ENV_VAR)
```

I feel It's kind of redundant to add another test. WDYT?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -482,6 +483,11 @@ def configure(conf)
 
       @platform = detect_platform
 
+      # Treat an empty setting of the credentials file path environment variable
+      # as unset. This way the googleauth lib could fetch the credentials
+      # following the fallback path.
+      ENV.delete(CREDENTIALS_PATH_ENV_VAR) if ENV[CREDENTIALS_PATH_ENV_VAR] == ''
```
igorpeshansky: [Optional] `ENV.delete(CREDENTIALS_PATH_ENV_VAR) unless ENV[CREDENTIALS_PATH_ENV_VAR].empty?`.

test/plugin/base_test.rb
bmoyles0117: I still like the idea of testing things in isolation. By having an explicit test for this, we don't create a passive dependency on other tests that happen to be doing the same thing by convenience.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Sync'ed offline. There was a typo. Igor in fact meant:
```
ENV.delete(CREDENTIALS_PATH_ENV_VAR) unless 
  ENV[CREDENTIALS_PATH_ENV_VAR].empty?
```

The problem with that is if `ENV[CREDENTIALS_PATH_ENV_VAR]` is nil, `.empty?` will error out.
```
$ irb
2.4.3 :001 > ENV['abc'].empty?
NoMethodError: undefined method `empty?' for nil:NilClass
        from (irb):1
        from /usr/local/google/home/lingshi/.rvm/rubies/ruby-2.4.3/bin/irb:11:in `<main>'
```

test/plugin/base_test.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Yeah...

test/plugin/base_test.rb
```diff
@@ -274,11 +274,21 @@ def test_invalid_json_credentials
     end
   end
 
+  def test_unset_empty_credentials_path_env_var
+    # An empty string should be treated as if it's not set.
+    ['', nil].each do |env_var|
```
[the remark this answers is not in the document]
qingling128: Done.

test/plugin/base_test.rb
```diff
@@ -274,11 +274,21 @@ def test_invalid_json_credentials
     end
   end
 
+  def test_unset_empty_credentials_path_env_var
```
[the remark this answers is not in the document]
qingling128: Done.

PR #237, opened 2018-08-18
Set User-Agent on the gRPC path.

igorpeshansky:
Also add unit tests for User-Agent on both paths.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1983,13 +1983,13 @@ module Fluent
           creds = :this_channel_is_insecure
         end
         port = ":#{uri.port}" if uri.port
+        user_agent = \
+          "#{PLUGIN_NAME}/#{PLUGIN_VERSION} grpc-ruby/#{GRPC::VERSION} " \
+          "#{Google::Apis::OS_VERSION}"
         @client = Google::Cloud::Logging::V2::LoggingServiceV2Client.new(
           credentials: GRPC::Core::Channel.new(
-            "#{host}#{port}", {
-              'grpc.primary_user_agent' =>
-                "#{PLUGIN_NAME}/#{PLUGIN_VERSION} grpc-ruby/#{GRPC::VERSION} " \
-                "#{Google::Apis::OS_VERSION}"
-            }, creds))
+            "#{host}#{port}", { 'grpc.primary_user_agent' => user_agent },
+            creds))
       else
         # TODO: Use a non-default ClientOptions object.
         Google::Apis::ClientOptions.default.application_name = PLUGIN_NAME
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1984,7 +1984,12 @@ module Fluent
         end
         port = ":#{uri.port}" if uri.port
         @client = Google::Cloud::Logging::V2::LoggingServiceV2Client.new(
-          credentials: GRPC::Core::Channel.new("#{host}#{port}", nil, creds))
+          credentials: GRPC::Core::Channel.new(
+            "#{host}#{port}", {
+              'grpc.primary_user_agent'=>
+                "#{PLUGIN_NAME}/#{PLUGIN_VERSION} grpc-ruby/#{GRPC::VERSION} " \
+                "#{Google::Apis::OS_VERSION}"
+            }, creds))
       else
         # TODO: Use a non-default ClientOptions object.
         Google::Apis::ClientOptions.default.application_name = PLUGIN_NAME
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1986,7 +1986,7 @@ module Fluent
         @client = Google::Cloud::Logging::V2::LoggingServiceV2Client.new(
           credentials: GRPC::Core::Channel.new(
             "#{host}#{port}", {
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1984,7 +1984,12 @@ def init_api_client
         end
         port = ":#{uri.port}" if uri.port
         @client = Google::Cloud::Logging::V2::LoggingServiceV2Client.new(
-          credentials: GRPC::Core::Channel.new("#{host}#{port}", nil, creds))
+          credentials: GRPC::Core::Channel.new(
+            "#{host}#{port}", {
+              'grpc.primary_user_agent' =>
+                "#{PLUGIN_NAME}/#{PLUGIN_VERSION} grpc-ruby/#{GRPC::VERSION} " \
```
qingling128: "grpc-ruby"? Do we mean to include a Ruby version? What's the "-ruby" for?

test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -27,6 +27,25 @@ def test_configure_use_grpc
...
+      old_initialize = instance_method(:initialize)
+      define_method(:initialize) do |url, args, creds|
+        user_agent = args['grpc.primary_user_agent']
+        old_initialize.bind(self).call(url, args, creds)
+      end
+    end
+
+    d = create_driver
+    d.instance.send :init_api_client
+    assert_match Regexp.new(Fluent::GoogleCloudOutput::PLUGIN_NAME.to_s), \
```
qingling128: Should we test against the full version?
"#{PLUGIN_NAME}/#{PLUGIN_VERSION} grpc-ruby/#{GRPC::VERSION} #{Google::Apis::OS_VERSION}"

lib/fluent/plugin/out_google_cloud.rb
qingling128: Oh, I guess it's the gRPC Ruby library (version)?

igorpeshansky: Yes, this is simulating what the REST libraries set the User Agent to — they concatenate `application_name`/`application_version` with the default User Agent string (`google-cloud-ruby/#{VERSION}`). I'm doing the same in the gRPC case.

test/plugin/test_out_google_cloud_grpc.rb
igorpeshansky: I think the key reason we set the User Agent is to be able to identify the requests coming from the logging agent. So this tests the minimal such identification. I think a more precise test will be more fragile and won't buy us anything — it'll be like a change detector test.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Hmm... I wonder why they chose `google-cloud-ruby` over `google-api-ruby-client`. Because the gRPC library is called [google-cloud-ruby](https://github.com/GoogleCloudPlatform/google-cloud-ruby). While the non-grpc library is called [google-api-ruby-client](https://github.com/google/google-api-ruby-client).

Since this is the gRPC path we are adding, we should call it `google-cloud-ruby`. But then the GRPC::VERSION is not the actual `google-cloud-ruby` version, which is still a bit confusing.

test/plugin/test_out_google_cloud_grpc.rb
qingling128: We also want to guarantee at least the plugin version is set (not empty) and correct. The test can check it against the variable in the main class so it's not a moving target.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: You're right, I got confused. They use `google-api-ruby-client` ([source](https://github.com/google/google-api-ruby-client/blob/master/lib/google/apis/core/base_service.rb#L431)). I do think we should append the default User Agent used by the underlying library, though, which is actually `grpc-ruby/#{VERSION}`.

qingling128: Is this the implementation in the REST path? https://github.com/google/google-api-ruby-client/blob/24be95a758ed3c4e559b8e31fd20ad32d166bd18/lib/google/apis/core/base_service.rb#L431

```
        # Build the user agent header
        # @return [String]
        def user_agent
          sprintf('%s/%s google-api-ruby-client/%s %s (gzip)',
                  client_options.application_name,
                  client_options.application_version,
                  Google::Apis::VERSION,
                  Google::Apis::OS_VERSION)
        end
```

Seems they are using `google-api-ruby-client` with the API version.

igorpeshansky:
@bmoyles0117, [this code](https://github.com/google/google-api-ruby-client/blob/master/lib/google/apis/core/base_service.rb#L431) sets the user agent for HTTP/REST.

test/plugin/test_out_google_cloud_grpc.rb
igorpeshansky: Done.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Our messages crossed. We both pointed to the same code. :-)

igorpeshansky: This is what the current code sends as the value of the User-Agent: `Fluentd Google Cloud Logging plugin/0.6.23 grpc-ruby/1.8.3`. Verified on a round-trip.

PR #241, opened 2018-08-27
Remove global resource in favor of logging an error at the log entry.

bmoyles0117:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1254,67 +1199,77 @@ def determine_group_level_monitored_resource_and_labels(tag,
...
-        instance_id = resource.labels.delete('instance_id')
-        common_labels.merge!(
-          "#{GKE_CONSTANTS[:service]}/instance_id" => instance_id,
-          "#{COMPUTE_CONSTANTS[:service]}/resource_id" => instance_id,
-          "#{GKE_CONSTANTS[:service]}/cluster_name" =>
-            resource.labels.delete('cluster_name'),
-          "#{COMPUTE_CONSTANTS[:service]}/zone" =>
-            resource.labels.delete('zone')
-        )
+      resource_type = determine_agent_level_monitored_resource_type
```
bmoyles0117: It seems like "determine_agent_level" may no longer be a valid name. Any suggestions for alternatives?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1169,12 +1135,12 @@ def determine_agent_level_monitored_resource_labels(type)
 
     # Determine the common labels that should be added to all log entries
     # processed by this logging agent.
-    def determine_agent_level_common_labels
+    def get_common_labels_for_resource_type(resource_type)
```
qingling128: There are additional common labels we extract per specific resource types at entry level.

e.g. https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/master/lib/fluent/plugin/out_google_cloud.rb#L1391.

Maybe the naming should reflect that. How about something like `get_common_labels_for_group_level_resource_type`?

test/plugin/base_test.rb
```diff
@@ -182,7 +171,6 @@ def test_gce_used_when_detect_subservice_is_false
     setup_container_metadata_stubs
     d = create_driver(NO_DETECT_SUBSERVICE_CONFIG)
     d.run
-    assert_equal COMPUTE_CONSTANTS[:resource_type], d.instance.resource.type
```
qingling128: With the assertion removed, this test is not really doing much. We probably should get rid of it. However, we need to make sure every unit test / assertion we get rid of has some equivalent test at the entry level. (Previously we were checking the global variable. It's kind of like a shortcut.)

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -297,9 +296,9 @@ module InternalConstants
 
     # The regular expression to use on Kubernetes logs to extract some basic
     # information about the log source. The regexp must contain capture groups
-    # for pod_name, namespace_name, and container_name.
+    # for pod_id, namespace_id, and container_name.
     config_param :kubernetes_tag_regexp, :string, :default =>
-      '\.(?<pod_name>[^_]+)_(?<namespace_name>[^_]+)_(?<container_name>.+)$'
+      '\.(?<pod_id>[^_]+)_(?<namespace_id>[^_]+)_(?<container_name>.+)$'
```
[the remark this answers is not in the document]
igorpeshansky: +1, the contract for this regexp is that it captures `pod_name` and `namespace_name`. We can't change it.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -571,6 +529,23 @@ def write(chunk)
             determine_entry_level_monitored_resource_and_labels(
               group_level_resource, group_level_common_labels, record)
 
+          unless @logs_viewer_address_displayed
+            instance_resource_types = [
```
igorpeshansky: This should be a constant in `ServiceConstants`. Also, this should probably be a Set.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: `discover_monitored_resource_type`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1254,67 +1199,77 @@ def determine_group_level_monitored_resource_and_labels(tag,
...
-          "#{COMPUTE_CONSTANTS[:service]}/resource_id" => instance_id,
-          "#{GKE_CONSTANTS[:service]}/cluster_name" =>
-            resource.labels.delete('cluster_name'),
-          "#{COMPUTE_CONSTANTS[:service]}/zone" =>
-            resource.labels.delete('zone')
-        )
+      resource_type = determine_agent_level_monitored_resource_type
+      resource = Google::Apis::LoggingV2::MonitoredResource.new(
+        type: resource_type,
+        labels: determine_agent_level_monitored_resource_labels(
```
igorpeshansky: Could we not create a `discover_monitored_resource` function that actually constructs the resource (i.e., resurrect `determine_agent_level_monitored_resource_via_legacy` under a new name)? We can then inline the functions for the type and the labels.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1254,67 +1199,77 @@ def determine_group_level_monitored_resource_and_labels(tag,
...
-              'namespace_name' => 'namespace_id',
-              'pod_name' => 'pod_id'))
+      common_labels = get_common_labels_for_resource_type(resource.type)
 
-          common_labels.merge!(
-            delete_and_extract_labels(
-              common_labels_candidates,
-              GKE_CONSTANTS[:extra_common_labels]
-                .map { |l| [l, "#{GKE_CONSTANTS[:service]}/#{l}"] }.to_h))
+      # Set regexp that we should match tags against later on. Using a list
```
igorpeshansky: I think this comment is out-of-date now.

bmoyles0117:

PR #242, opened 2018-08-27
Determine common labels when processing log entries.

bmoyles0117:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1167,14 +1162,14 @@ def determine_agent_level_monitored_resource_labels(type)
...
-    def determine_agent_level_common_labels
+    # Determine the common labels that should be added to all log entries for
+    # a given resource type.
+    def determine_common_labels_for_resource_type(resource_type)
       labels = {}
       # User can specify labels via config. We want to capture those as well.
       labels.merge!(@labels) if @labels
 
-      case @resource.type
+      case resource_type
```
igorpeshansky: This will always return the same value for a given resource type. Would it make sense to memoize it?

bmoyles0117:

PR #243, opened 2018-08-27
Return early when discovering a resource via local_resource_id.

bmoyles0117:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1224,6 +1224,30 @@ def group_log_entries_by_tag_and_local_resource_id(chunk)
...
+        resource = monitored_resource_from_local_resource_id(local_resource_id)
+        if resource
+          common_labels = determine_common_labels_for_resource_type(
+            resource.type)
+
+          resource.freeze
+          resource.labels.freeze
+          common_labels.freeze
+
+          return [resource, common_labels]
```
qingling128: What happens to `gke_container` when we return early? Additional common labels like `namespace_name` and `pod_name` will be missing (https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/243/files#diff-0d6486c76b449a4c9242958738de9cdbL1296).

As of today, this is not a problem because we don't use Metadata Agent for legacy `gke_container` resource, but Metadata Agent does support it. Should we explicitly exclude it?

bmoyles0117: This is only returned when the local resource is discovered via the metadata agent, the metadata agent should return the "correct" monitored resource.

bmoyles0117: I misunderstood this, I've duplicated the logic for copying common labels as we discussed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1224,6 +1224,41 @@ def group_log_entries_by_tag_and_local_resource_id(chunk)
...
+          resource.freeze
+          resource.labels.freeze
+
+          common_labels = determine_common_labels_for_resource_type(
+            resource.type)
+
+          # TODO(bmoyles0117): Remove this once backend service can map metadata
+          # labels.
+          if resource.type == GKE_CONSTANTS[:resource_type]
+            common_labels.merge!(
```
qingling128: `delete_and_extract_labels` need to extract `namespace_name` and `pod_name` from a hash that include these two fields (in the legacy path it's from the regex matched group). Here only the `resource.labels` have those fields. However, the `delete_and_extract_labels` method also *delete* the specified fields as a side effect.

I would recommend just doing it explicitly.
```
            common_labels.merge!(
               'namespace_name' => resource.labels['namespace_id'],
               'pod_name' => resource.labels['pod_id']
            )
```

bmoyles0117: Thanks, updated

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1224,6 +1224,42 @@ def group_log_entries_by_tag_and_local_resource_id(chunk)
...
+      # inferred from the tag) if failed to get a monitored resource from
+      # Metadata Agent with this key.
+      #
+      # Examples:
+      # "container.<container_id>" // Docker container.
+      # "k8s_pod.<namespace_name>.<pod_name>" // GKE pod.
+      if local_resource_id
+        resource = monitored_resource_from_local_resource_id(local_resource_id)
+        if resource
+          resource.freeze
```
igorpeshansky: Would it make sense for `monitored_resource_from_local_resource_id` to return the resource already frozen?

bmoyles0117:

PR #245, opened 2018-08-28
Extract version from gemspec, rather than repeating it in the plugin.

igorpeshansky:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -228,7 +228,8 @@ module InternalConstants
     Fluent::Plugin.register_output('google_cloud', self)
 
     PLUGIN_NAME = 'Fluentd Google Cloud Logging plugin'.freeze
-    PLUGIN_VERSION = '0.6.23'.freeze
+    PLUGIN_VERSION = \
+      Gem.loaded_specs['fluent-plugin-google-cloud'].version.to_s.freeze
```
bmoyles0117: I can't find `loaded_specs` here:
https://ruby-doc.org/stdlib-2.2.3/libdoc/rubygems/rdoc/Gem/Specification.html

Did you mean `latest_specs`?

bmoyles0117: Following up, I'm really confused, because `latest_specs` doesn't exist in my local irb. Your code seems right, I just can't find the reference.

igorpeshansky: It's not in the `Gem::Specification` class, it's in the `Gem` module: https://ruby-doc.org/stdlib-2.2.3/libdoc/rubygems/rdoc/Gem.html#loaded_specs.

PR #247, opened 2018-08-29
Fix cascading JSON detection with LogEntry fields.

qingling128:
Today nested JSON detection only happens when the record has one field named "log", "message", "msg". The size of the hash matters.

https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/8d26465c0f777a42938e3acae857936627a38511/lib/fluent/plugin/out_google_cloud.rb#L588

Fields like `logging.googleapis.com/trace`, `logging.googleapis.com/spanId` and `logging.googleapis.com/insertId` should not count towards the hash size when determining whether to extract JSON from the nested field.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -600,8 +600,8 @@ module Fluent
               # Extract LogEntry fields if they are present in the nested JSON.
               # This will take precedence of the root level values.
             end
-            # Restore these if necessary. Note that we don't
-            # want to override these keys in the JSON we've just parsed.
+            # Restore these if necessary. Note that we don't want to override
+            # these keys in the JSON we've just parsed.
             record['time'] ||= timestamp if timestamp
             record['severity'] ||= severity if severity
             record[@trace_key] ||= trace if trace
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -602,6 +602,11 @@ module Fluent
             unless record_json.nil?
               record = record_json
               is_json = true
+              # Extract LogEntry fields if they are present in the nested JSON.
+              # This will take precedence of the root level values.
+              fq_trace_id ||= record.delete(@trace_key)
+              span_id ||= record.delete(@span_id_key)
+              insert_id ||= record.delete(@insert_id_key)
             end
             # Restore timestamp and severity if necessary. Note that we don't
             # want to override these keys in the JSON we've just parsed.
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -597,8 +597,6 @@ module Fluent
             unless record_json.nil?
               record = record_json
               is_json = true
-              # Extract LogEntry fields if they are present in the nested JSON.
-              # This will take precedence of the root level values.
             end
             # Restore these if necessary. Note that we don't want to override
             # these keys in the JSON we've just parsed.
@@ -620,11 +618,12 @@ module Fluent
                                             ts_secs,
                                             ts_nanos)
 
-          entry.trace = record.delete(@trace_key) if record.key?(@trace_key)
-          entry.span_id = record.delete(@span_id_key) if
-            record.key?(@span_id_key)
-          entry.insert_id = record.delete(@insert_id_key) if
-            record.key?(@insert_id_key)
+          trace = record.delete(@trace_key)
+          entry.trace = trace if trace
+          span_id = record.delete(@span_id_key)
+          entry.span_id = span_id if span_id
+          insert_id = record.delete(@insert_id_key)
+          entry.insert_id = insert_id if insert_id
 
           set_log_entry_fields(record, entry)
           set_payload(entry_level_resource.type, record, entry, is_json)
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -577,7 +577,7 @@ module Fluent
           # record so that cascading JSON detection happens (we rely on the
           # record having only 1 field to trigger the detection). These will be
           # set later in the LogEntry.
-          fq_trace_id = record.delete(@trace_key)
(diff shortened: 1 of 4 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -612,14 +620,8 @@ def write(chunk)
                                             ts_secs,
                                             ts_nanos)
 
-          # Get fully-qualified trace id for LogEntry "trace" field.
-          fq_trace_id = record.delete(@trace_key)
```
igorpeshansky: This breaks the use case where the user's JSON contains those fields instead.

qingling128: These keys (https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/8d26465c0f777a42938e3acae857936627a38511/lib/fluent/plugin/out_google_cloud.rb#L149) are reserved for LogEntry fields. By default they come with the `logging.googleapis.com/` prefixes. It should not conflict with other fields in users' namespace if that was the concern.

On the other hand, before this PR we already had the logic (https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/8d26465c0f777a42938e3acae857936627a38511/lib/fluent/plugin/out_google_cloud.rb#L617) to delete these fields from the record.

igorpeshansky: You're missing the fact that when the text payload is detected to be serialized JSON, the parsed JSON becomes the record. So users previously could serialize JSON with those fields into text, which would then be extracted into the log entry. This would no longer be possible after this change as-is.
If you want to make the behavior consistent, you should remove it from the record right before detecting JSON, and then reinstate it in the resulting record only if it's not already present (e.g., the user's code dumped its own value of insert_id into the JSON).

That said, what tools would set the trace_id or span_id in a record that also contains serialized JSON?

qingling128: Oh, I think I misunderstood. You were talking about the LogEntry fields being set in the nested JSON. Then maybe we should check both levels.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -594,6 +602,11 @@ def write(chunk)
             unless record_json.nil?
               record = record_json
               is_json = true
+              # Extract LogEntry fields if they are present in the nested JSON.
+              # This will take precedence of the root level values.
+              fq_trace_id ||= record.delete(@trace_key)
```
igorpeshansky: This is wrong too, since the outer fields would override the nested fields, while it should be the other way around.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Oops, race condition. Did not see the second comment before I commented.

How about something like this? https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/247/commits/34effd81aa1433e972ebe8105444f06c33c17412

Is it worth checking these fields on both levels?

I'm not sure if there are tools that would set `trace_id` or `span_id` in a record that also contains serialized JSON. If they use the `record_transformer` plugin to reformat it somehow, it might end up being at the root level.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Switched the order in https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/247/commits/94a2ff67c58ec2d2995f7b032858f6c36b08ae8d.

I would propose this way instead of mocking exactly what we do with `time` and `severity` fields because:
For `time` and `severity` fields, we are not deleting them from the record in the end, so it makes sense to put them back into the record. For `trace`, `spanId` and `insertId`, we eventually delete them. So just storing the value in local variables should be good enough.

WDYT?

qingling128: Updated. PTAL

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -594,11 +597,16 @@ def write(chunk)
             unless record_json.nil?
               record = record_json
               is_json = true
+              # Extract LogEntry fields if they are present in the nested JSON.
+              # This will take precedence of the root level values.
             end
-            # Restore timestamp and severity if necessary. Note that we don't
+            # Restore these if necessary. Note that we don't
```
bmoyles0117: nit: can we expand this comment line up to 80 characters?

qingling128: Updated.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -594,11 +597,16 @@ def write(chunk)
             unless record_json.nil?
               record = record_json
               is_json = true
+              # Extract LogEntry fields if they are present in the nested JSON.
```
igorpeshansky: This comment is no longer accurate, right?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -612,15 +620,11 @@ def write(chunk)
...
-          # Get fully-qualified trace id for LogEntry "trace" field.
-          fq_trace_id = record.delete(@trace_key)
-          entry.trace = fq_trace_id if fq_trace_id
-
-          span_id = record.delete(@span_id_key)
-          entry.span_id = span_id if span_id
-
-          insert_id = record.delete(@insert_id_key)
-          entry.insert_id = insert_id if insert_id
+          entry.trace = record.delete(@trace_key) if record.key?(@trace_key)
```
igorpeshansky: Should we revert to the previous code?

lib/fluent/plugin/out_google_cloud.rb

lib/fluent/plugin/out_google_cloud.rb
qingling128: Seems like the same thing. But i reverted anyway.

igorpeshansky: The previous code you've reverted to does one less lookup.

PR #249, opened 2018-08-30
Simplify the preservation of google record keys when parsing json.

bmoyles0117:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -579,33 +579,27 @@ def write(chunk)
             # Save the following fields if available, then clear them out to
             # allow for determining whether we should parse the log or message
             # field.
-            timestamp = record.delete('time')
-            severity = record.delete('severity')
-            trace = record.delete(@trace_key)
-            span_id = record.delete(@span_id_key)
-            insert_id = record.delete(@insert_id_key)
+            preserved_keys = ['time', 'severity', @trace_key, @span_id_key,
```
igorpeshansky: [Optional] Might be more readable as:
```ruby
            preserved_keys = [
              'time', 'severity', @trace_key, @span_id_key, @insert_id_key
            ]
```
Don't know if Rubocop will like that closing brace on its own line, though, so you might have to join it onto the previous one.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -579,33 +579,27 @@ def write(chunk)
...
-            if record.length == 1
+            if (record.keys - preserved_keys).length == 1
               %w(log message msg).each do |field|
                 if record.key?(field)
                   record_json = parse_json_or_nil(record[field])
                 end
               end
             end
             unless record_json.nil?
+              # Set these if necessary. Note that we don't want to override
```
igorpeshansky: I'd suggest s/`Set`/`Propagate`/.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -579,33 +579,28 @@ def write(chunk)
...
               %w(log message msg).each do |field|
                 if record.key?(field)
                   record_json = parse_json_or_nil(record[field])
                 end
               end
             end
             unless record_json.nil?
+              # Propagate these if necessary. Note that we don't want to
+              # override these keys in the JSON we've just parsed.
+              preserved_keys.each { |key| record_json[key] ||= record[key] }
```
qingling128: Should this be
```
              preserved_keys.each { |key| record_json[key] ||= record[key] if record.key?(key) }
```
instead?

Otherwise the record might have a few `nil` fields when the `key` is not present in `record`, right?

```
2.4.3 :001 > record = {'a' => 'b'}
 => {"a"=>"b"}
2.4.3 :003 > record['c'] = nil
 => nil
2.4.3 :004 > record
 => {"a"=>"b", "c"=>nil}
2.4.3 :008 > record.to_json
 => "{\"a\":\"b\",\"c\":null}"
```

igorpeshansky: Good call!

PR #229, opened 2018-07-03
Use cluster-name directly instead of kube-env to populate cluster_name label

ihmccreery:
When metadata concealment is enabled, kube-env is not available to pods outside of `hostNetwork`.

PR #252, opened 2018-09-06
Add container tag when testing k8s container resource.

qingling128:
https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/229 depends on this.

### DIFF
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -1684,8 +1684,10 @@ module BaseTest
       stub_metadata_request(
         'instance/attributes/',
         "attribute1\ncluster-name\ncluster-location\nlast_attribute")
-      stub_metadata_request('instance/attributes/cluster-location', K8S_LOCATION2)
-      stub_metadata_request('instance/attributes/cluster-name', K8S_CLUSTER_NAME)
+      stub_metadata_request('instance/attributes/cluster-location',
+                            K8S_LOCATION2)
+      stub_metadata_request('instance/attributes/cluster-name',
+                            K8S_CLUSTER_NAME)
     else
       ['cluster-location', 'cluster-name'].each do |metadata_name|
         stub_request(:get, %r{.*instance/attributes/#{metadata_name}.*})
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -1679,8 +1679,8 @@ module BaseTest
                           'KUBE_BEARER_TOKEN: AoQiMuwkNP2BMT0S')
   end
 
-  def setup_k8s_metadata_stubs(should_response = true)
-    if should_response
+  def setup_k8s_metadata_stubs(should_respond = true)
+    if should_respond
       stub_metadata_request(
         'instance/attributes/',
         "attribute1\ncluster-name\ncluster-location\nlast_attribute")
@@ -1740,8 +1740,8 @@ module BaseTest
     WebMock.reset!
   end
 
-  def setup_metadata_agent_stubs(should_response = true)
-    if should_response
+  def setup_metadata_agent_stubs(should_respond = true)
+    if should_respond
       MONITORED_RESOURCE_STUBS.each do |local_resource_id, resource|
         stub_request(:get, metadata_request_url(local_resource_id))
           .to_return(status: 200, body: resource)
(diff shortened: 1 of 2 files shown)


test/plugin/base_test.rb
```diff
@@ -1434,9 +1433,41 @@ def test_k8s_monitored_resource_fallback
...
+        else
+          setup_no_metadata_agent_stubs
+        end
+        if test_params[:setup_k8s_stub]
+          setup_k8s_metadata_stubs
+        else
+          setup_no_k8s_metadata_stubs
+        end
+        setup_logging_stubs do
+          d = create_driver(test_params[:config], CONTAINER_TAG)
```
qingling128: The actual change is just
`          d = create_driver(test_params[:config], CONTAINER_TAG)`
for `k8s_container` and 
`          d = create_driver(test_params[:config])`
for `k8s_node`.

The test is too big so I split it.

test/plugin/base_test.rb
```diff
@@ -1434,9 +1433,41 @@ def test_k8s_monitored_resource_fallback
           local_resource_id: RANDOM_LOCAL_RESOURCE_ID),
         # When 'kube-env' is present, "compute.googleapis.com/resource_name" is
         # not added.
-        expected_params: COMPUTE_PARAMS
-      },
-      # Specific cases for k8s_node.
+        expected_params: COMPUTE_PARAMS_WITH_CONTAINER_TAG
+      }
+    ].each do |test_params|
+      new_stub_context do
```
bmoyles0117: Do we want to create a helper method for this section to avoid duplicated code?

test/plugin/constants.rb
```diff
@@ -510,6 +509,10 @@ module Constants
       )
     )
   ).freeze
+  # Used in fallback tests.
```
igorpeshansky: Want to keep it with the other `COMPUTE_PARAMS_*` constants?

test/plugin/base_test.rb
igorpeshansky: Since `k8s_node` doesn't care about the log tag, why not set its tag to `CONTAINER_TAG` as well?

test/plugin/base_test.rb
```diff
@@ -1373,13 +1372,13 @@ def test_k8s_monitored_resource_fallback
         log_entry: k8s_container_log_entry(log_entry(0)),
         expected_params: K8S_CONTAINER_PARAMS_FROM_LOCAL
       },
-      # When enable_metadata_agent is true.
+      ## When enable_metadata_agent is true.
```
igorpeshansky: Intentional?

test/plugin/base_test.rb
igorpeshansky: +1 on the helper method.

qingling128: A new helper method is a bit overkill and will decrease readability. I refactored the existing helper methods a bit. Hope that helps. PTAL.

test/plugin/base_test.rb
qingling128: Because it's trying to simulate the behavior in Kubernetes. `CONTAINER_TAG` is not part of node logs.

test/plugin/constants.rb
qingling128: Done.

test/plugin/base_test.rb
```diff
@@ -1664,21 +1679,21 @@ def setup_container_metadata_stubs
...
-  end
-
-  def setup_no_k8s_metadata_stubs
-    ['cluster-location', 'cluster-name'].each do |metadata_name|
-      stub_request(:get, %r{.*instance/attributes/#{metadata_name}.*})
-        .to_return(status: 404,
-                   body: 'The requested URL /computeMetadata/v1/instance/' \
-                         "attributes/#{metadata_name} was not found on this" \
-                         ' server.')
+  def setup_k8s_metadata_stubs(should_response = true)
```
bmoyles0117: s/response/respond/

test/plugin/base_test.rb
igorpeshansky: This helps a bit, but Bryan and I were both talking about the code that follows stub setup, which is identical modulo the tag passed into `create_driver`...

igorpeshansky: Good point.

PR #254, opened 2018-09-11
Bump version to 0.6.25.

qingling128:


igorpeshansky:
I'm confused: you merged GoogleCloudPlatform/google-fluentd#120 before this one?..

qingling128:
Yeah. google-fluentd won't build without this one. But we won't release without this new gem. I figured there is no strong dependency.

PR #240, opened 2018-08-24
Upgrade Fluentd to v1.2.5 and bump gem version to v0.7.0.

qingling128:
Upgrade Fluentd to v1.2.5 (latest as of Aug 24, 2018) to apply this fix https://github.com/fluent/fluentd/issues/1941.
Fixes #251.

PR #255, opened 2018-09-13
Use Fluentd v0.12 compatible module path.

qingling128:
Change `Fluent::Plugin::AddInsertIdsFilter` to `Fluent::AddInsertIdsFilter`.

Option 1:
Merge in this PR.
Option 2:
Abandon this PR. Instead revert the new add_insert_ids plugin, do a release without it. Upgrade to Fluentd v1.2 then and release add_insert_ids plugin in the next release.
Option 3:
Abandon this PR. Instead Merge the Fluentd v1.2 upgrade (https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/240) now that includes marking Fluentd v1.2 as a dependency.

### DIFF
--- a/lib/fluent/plugin/filter_add_insert_ids.rb
+++ b/lib/fluent/plugin/filter_add_insert_ids.rb
@@ -12,92 +12,88 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-require 'fluent/plugin/filter'
-
 module Fluent
-  module Plugin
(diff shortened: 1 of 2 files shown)


igorpeshansky:
Sigh, I was going to propose that we simply don't release off the Kubernetes branch, but they need #229, don't they?.. Lets go with option 1 for now.

PR #256, opened 2018-09-13
Use wild card in file permission fixes.

qingling128:


Rakefile
```diff
@@ -21,8 +21,7 @@ end
 desc 'Fix file permissions'
 task :fix_perms do
   files = [
-    'lib/fluent/plugin/out_google_cloud.rb',
-    'lib/fluent/plugin/monitoring.rb',
+    'lib/fluent/plugin/*.rb',
```
igorpeshansky: Just FYI, these don't run anyway when we build the gem. We should switch to using `rake` to build the gem so we can ensure these are invoked.

PR #260, opened 2018-09-25
add config to autoformat stackdriver trace

StevenYCChou:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -393,6 +399,10 @@ module InternalConstants
     config_param :monitoring_type, :string,
                  :default => Monitoring::PrometheusMonitoringRegistry.name
 
+    # Whether to autoformat value of "logging.googleapis.com/trace" to
+    # LogEntry.Trace.
+    config_param :enable_stackdriver_trace_autoformat, :bool, :default => true
```
igorpeshansky: Let's use the imperative form: `autoformat_stackdriver_trace`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -625,9 +635,15 @@ def write(chunk)
                                             severity,
                                             ts_secs,
                                             ts_nanos)
-
           trace = record.delete(@trace_key)
-          entry.trace = trace if trace
+          if trace
+            if @enable_stackdriver_trace_autoformat &&
```
igorpeshansky: Let's extract this into a `compute_trace` method and leave this line as:
```ruby
entry.trace = compute_trace(trace) if trace
```

test/plugin/base_test.rb
```diff
@@ -338,6 +338,81 @@ def test_structured_payload_log
...
+  def test_stackdriver_trace_format_compliant_trace
+    full_stackdriver_trace = \
+      'projects/some-proj/traces/1234567890abcdef1234567890abcdef'
+
+    [
+      {
+        config: APPLICATION_DEFAULT_CONFIG
+      },
+      {
+        config: 'enable_stackdriver_trace_autoformat true'
```
igorpeshansky: Please extract this and the next config string into constants, and define them in terms of `APPLICATION_DEFAULT_CONFIG`.

test/plugin/base_test.rb
```diff
@@ -338,6 +338,81 @@ def test_structured_payload_log
...
+      new_stub_context do
+        setup_gce_metadata_stubs
+        setup_logging_stubs do
+          d = create_driver(test_params[:config])
+          d.emit(DEFAULT_TRACE_KEY => non_stackdriver_trace_compliant_trace)
+          d.run
+
+          verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
+            expected_trace = \
+              "projects/#{d.instance.project_id}" \
```
igorpeshansky: This is a change detector test (it replicates the code under test). Please use the literal value. Better yet, define two constants, one for the compliant trace, and one for the non-compliant trace.

test/plugin/base_test.rb
qingling128: +1. Similar to https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/master/test/plugin/constants.rb#L235.

test/plugin/base_test.rb
```diff
@@ -338,6 +338,81 @@ def test_structured_payload_log
     end
   end
 
+  def test_stackdriver_trace_format_compliant_trace
+    full_stackdriver_trace = \
```
qingling128: Constant `FULL_STACKDRIVER_TRACE`?

test/plugin/base_test.rb
```diff
@@ -338,6 +338,81 @@ def test_structured_payload_log
...
+          verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
+            assert_equal full_stackdriver_trace, entry['trace']
+          end
+        end
+      end
+    end
+  end
+
+  def test_autoformat_enabled_with_stackdriver_trace_format_noncompliant_trace
+    non_stackdriver_trace_compliant_trace = '1234567890abcdef1234567890abcdef'
```
qingling128: Constant `NON_STACKDRIVER_TRACE`?

test/plugin/base_test.rb
```diff
@@ -338,6 +338,81 @@ def test_structured_payload_log
...
+    ].each do |test_params|
+      new_stub_context do
+        setup_gce_metadata_stubs
+        setup_logging_stubs do
+          d = create_driver(test_params[:config])
+          d.emit(DEFAULT_TRACE_KEY => full_stackdriver_trace)
+          d.run
+
+          verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
+            assert_equal full_stackdriver_trace, entry['trace']
```
qingling128: `assert_equal` can take a third parameter as the error message when the assertion fails. In order to help with debugging, we could something like:

```
assert_equal full_stackdriver_trace, entry['trace'],
             "test_params #{test_params} failed with entry #{entry}."
```

Same for below.

test/plugin/base_test.rb
```diff
@@ -338,6 +338,81 @@ def test_structured_payload_log
...
+              "/traces/#{non_stackdriver_trace_compliant_trace}"
+            assert_equal expected_trace, entry['trace']
+          end
+        end
+      end
+    end
+  end
+
+  def test_autoformat_disabled_with_stackdriver_trace_format_noncompliant_trace
+    config = 'enable_stackdriver_trace_autoformat false'
```
qingling128: Constant `AUTOFORMAT_STACKDRIVER_TRACE_DISABLED_CONFIG`?

test/plugin/base_test.rb
```diff
@@ -338,6 +338,81 @@ def test_structured_payload_log
...
+    non_stackdriver_trace_compliant_trace = '1234567890abcdef1234567890abcdef'
```
qingling128: Constant `NON_STACKDRIVER_TRACE`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -625,9 +635,15 @@ def write(chunk)
                                             severity,
                                             ts_secs,
                                             ts_nanos)
-
           trace = record.delete(@trace_key)
-          entry.trace = trace if trace
+          if trace
```
qingling128: We probably wanna check against empty strings as well? I'm leaning towards not formatting the field if it comes as an empty string. WDYT?

BTW, Ruby will take if <empty_string> as `true`.

```
$ irb
2.4.3 :001 > empty_string = ""
 => ""
2.4.3 :002 > if empty_string then true else false end
 => true
```

test/plugin/base_test.rb
```diff
@@ -338,6 +338,81 @@ def test_structured_payload_log
     end
   end
 
+  def test_stackdriver_trace_format_compliant_trace
```
qingling128: The logic in these three tests are almost identical. They could be combined and use something like:

```
      {
        config: APPLICATION_DEFAULT_CONFIG,
        input_trace: XXX,
        expected_trace: XXX
      },
```

This way, it's easier to maintain the shared logic and also execute all tests related to `trace` with a single command. It also makes it easier to add more test cases.

BTW, we should also test cases like below in combination of different config values:
- There's no `logging.googleapis.com/trace` field in payload.
- `logging.googleapis.com/trace` field in payload is an empty string.
- A custom `trace_key` is used instead of the default trace key. (Reference: `trace_key` is configurable: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/master/lib/fluent/plugin/out_google_cloud.rb#L284)

test/plugin/base_test.rb
qingling128: Yeah, instead of `d.instance.project_id`, we could just use `PROJECT`, which is defined in the `constants.rb` file.

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: https://gist.github.com/jfarmer/2647362

test/plugin/base_test.rb
igorpeshansky: I was proposing the `FULL_STACKDRIVER_TRACE` and `NON_STACKDRIVER_TRACE`, same as you... If we define one in terms of the other, we can even use `FULL_STACKDRIVER_TRACE` as the expected value here...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -625,9 +635,15 @@ def write(chunk)
...
+            if @enable_stackdriver_trace_autoformat &&
+               !STACKDRIVER_TRACE_REGEXP.match(trace)
```
igorpeshansky: Just realized that a trace value that isn't in the right format but does have a slash will probably produce a log entry that will fail validation. If we adjust it unconditionally, then the resulting error might not make sense to the customers (e.g., "too many slashes" instead of "too few slashes"). We should confirm this behavior and, if my hunch is right, change this to a positive regexp match instead (i.e., only adjust if it has a certain form). WDYT?

lib/fluent/plugin/out_google_cloud.rb
qingling128: Fun read. Oh Ruby... shoot me:

```
$ irb
2.4.3 :001 > if 0 then true else false end
 => true
```

lib/fluent/plugin/out_google_cloud.rb
qingling128: That's a valid concern. Agreed we should confirm the behavior.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -165,6 +165,12 @@ module InternalConstants
       # monitored resource from Stackdriver Metadata agent.
       LOCAL_RESOURCE_ID_KEY = 'logging.googleapis.com/local_resource_id'.freeze
 
+      # The regexp matches stackdriver trace format. An example string
+      # will match the regexp is:
+      # "projects/my-project/traces/1234567890abcdef1234567890abcdef"
+      STACKDRIVER_TRACE_REGEXP = Regexp.new(
+        '^projects\/[^\/]*\/traces\/[^\/]*$').freeze
```
qingling128: We probably wanna change the `*` to `+` in the Regexp, right?

lib/fluent/plugin/out_google_cloud.rb
StevenYCChou: I found format of TraceID is documented and fully supported under [Trace API v1 (REST)](https://cloud.google.com/trace/docs/reference/v1/rest/v1/projects.traces#Trace), and [Trace API (rpc) - google.devtools.cloudtrace](https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v1#google.devtools.cloudtrace.v1.Trace) - so I will rely on that to write the regexp matching.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -393,6 +398,10 @@ module InternalConstants
     config_param :monitoring_type, :string,
                  :default => Monitoring::PrometheusMonitoringRegistry.name
 
+    # Whether to autoformat value of "logging.googleapis.com/trace" to
```
qingling128: It would be nice to provide more informative description on what this configuration does to help guide users. How about:
```
Whether to autoformat value of "logging.googleapis.com/trace" to comply with Stackdriver Trace format ("projects/[PROJECT-ID]/traces/[TRACE-ID]") when setting the trace field of the LogEntry object.
```

BTW, once this is released, we can update the [public documentation](https://cloud.google.com/logging/docs/agent/configuration#cloud-fluentd-config) to add this new config.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -627,7 +637,8 @@ def write(chunk)
                                             ts_nanos)
 
           trace = record.delete(@trace_key)
-          entry.trace = trace if trace
+          entry.trace = compute_trace(trace) if trace
```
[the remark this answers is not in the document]
qingling128: `if trace` is fine now since we are using positive regex match

test/plugin/base_test.rb
```diff
@@ -338,6 +338,98 @@ def test_structured_payload_log
...
+  end
+
+  def test_non_stackdriver_trace_id_compliant_trace_with_any_autoformat_config
+    configs = [
+      APPLICATION_DEFAULT_CONFIG,
+      AUTOFORMAT_STACKDRIVER_TRACE_CONFIG,
+      NO_AUTOFORMAT_STACKDRIVER_TRACE_CONFIG
+    ]
+    traces = [
+      '',
```
qingling128: Hmm... The empty string case seems to have failed the unit test for gRPC. [Travis build](https://travis-ci.org/GoogleCloudPlatform/fluent-plugin-google-cloud/builds/433786673?utm_source=github_status&utm_medium=notification).

It's probably due to https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/master/test/plugin/test_out_google_cloud_grpc.rb#L439.

test/plugin/base_test.rb
```diff
@@ -338,6 +338,98 @@ def test_structured_payload_log
...
+      new_stub_context do
+        setup_gce_metadata_stubs
+        setup_logging_stubs do
+          d = create_driver(test_config)
+          d.emit(DEFAULT_TRACE_KEY => STACKDRIVER_TRACE_ID)
+          d.run
+          verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
+            expected_trace = \
+              "projects/#{d.instance.project_id}/traces/#{STACKDRIVER_TRACE_ID}"
+            assert_equal expected_trace, entry['trace'], 'stackdriver trace ' \
```
qingling128: Great message!

lib/fluent/plugin/out_google_cloud.rb
StevenYCChou: updated the comment to clarify the config!

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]
StevenYCChou: ack.

test/plugin/base_test.rb
StevenYCChou: thanks!

test/plugin/base_test.rb
StevenYCChou: use the function `assert_equal_with_default` to be able to verify even it's an empty string `''`

test/plugin/base_test.rb
```diff
@@ -338,6 +338,98 @@ def test_structured_payload_log
     end
   end
 
+  def test_autoformat_enabled_with_stackdriver_trace_id_as_trace
+    configs = [
```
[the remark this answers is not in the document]
igorpeshansky: I meant:
```ruby
    [
      APPLICATION_DEFAULT_CONFIG,
      AUTOFORMAT_STACKDRIVER_TRACE_CONFIG
      ENABLE_AUTOFORMAT_STACKDRIVER_TRACE_CONFIG
    ].each do |config|
      ...
    end
```

test/plugin/constants.rb
```diff
@@ -265,6 +267,14 @@ module Constants
     enable_metadata_agent false
   ).freeze
 
+  AUTOFORMAT_STACKDRIVER_TRACE_CONFIG = %(
```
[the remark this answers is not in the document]
igorpeshansky: Sorry to be unclear. The names for the config constants don't necessarily have to include the full config option name — they just need to be identifiable with that option. We have this pattern elsewhere. I was suggesting that we change `<FULL_NAME>_CONFIG` and `NO_<FULL_NAME>_CONFIG` to `ENABLE_<SHORT_NAME>_CONFIG` and `DISABLE_<SHORT_NAME>_CONFIG`. Does this make sense?

test/plugin/base_test.rb
igorpeshansky: The project we feed into the tests is a constant, though... So we could construct the full expected value as a constant.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -676,6 +686,14 @@ def write(chunk)
 
     private
 
+    def compute_trace(trace)
+      if @autoformat_stackdriver_trace &&
```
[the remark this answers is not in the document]
igorpeshansky: The point here is that the important bit is the early return, not the clause. A reader, seeing `return trace unless [...]` will think: "Ah, this function will return one of two values — an untransformed one unless some condition holds true, and a transformed one otherwise". Then they can dig into the condition. In the code you're proposing, the logic goes the other way — the reader first has to get through a large condition to figure out how the value is transformed, and then look at the end of the function to see that it's returned.

Incidentally, even if we agree that your current structure is more readable, I think it's better to say `return "projects/#{@project_id}/traces/#{trace}"` instead of assigning it back to a parameter `trace` only to return it in the end.

test/plugin/base_test.rb
```diff
@@ -338,6 +338,98 @@ def test_structured_payload_log
...
+    configs.each do |test_config|
+      new_stub_context do
+        setup_gce_metadata_stubs
+        setup_logging_stubs do
+          d = create_driver(test_config)
+          d.emit(DEFAULT_TRACE_KEY => STACKDRIVER_TRACE_ID)
+          d.run
+          verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
+            expected_trace = \
+              "projects/#{d.instance.project_id}/traces/#{STACKDRIVER_TRACE_ID}"
```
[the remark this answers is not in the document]
igorpeshansky: The interpolation happens *before* the `freeze`, so the resulting string is a constant. Plus, this would be consistent with the rest of the testing code.

test/plugin/base_test.rb
```diff
@@ -338,6 +338,98 @@ def test_structured_payload_log
...
+            assert_equal expected_trace, entry['trace'], 'stackdriver trace ' \
+                         'id should be autoformatted with config ' \
+                         "#{test_config}."
```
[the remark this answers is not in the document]
igorpeshansky: Ping?

test/plugin/base_test.rb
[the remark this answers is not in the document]
StevenYCChou: updated to be consistent with the rest of the testing code.

test/plugin/constants.rb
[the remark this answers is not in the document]
StevenYCChou: make seonse. still decide to use the `ENABLE/DISABLE_FULL_NAME_CONFIG`.

test/plugin/constants.rb
```diff
@@ -265,6 +267,14 @@ module Constants
...
+    autoformat_stackdriver_trace true
+  ).freeze
+
+  NO_AUTOFORMAT_STACKDRIVER_TRACE_CONFIG = %(
```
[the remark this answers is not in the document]
StevenYCChou: ditto as above

test/plugin/base_test.rb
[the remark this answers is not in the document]
igorpeshansky: Can we do the same in `test_non_stackdriver_trace_id_compliant_trace_with_any_autoformat_config` as well? I don't think we have any other tests that interpolate the full config in the message...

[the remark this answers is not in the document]
StevenYCChou: updated

test/plugin/constants.rb
```diff
@@ -106,6 +106,30 @@ module Constants
   INSERT_ID = 'fah7yr7iw64tg857y'.freeze
   INSERT_ID2 = 'fah7yr7iw64tgaeuf'.freeze
 
+  STACKDRIVER_TRACE_ID = '1234567890abcdef1234567890abcdef'.freeze
+  FULL_STACKDRIVER_TRACE = \
+    "projects/#{PROJECT_ID}/traces/#{STACKDRIVER_TRACE_ID}".freeze
+
+  # Invalid trace id for stackdriver.
+  EMPTY_STRING = ''.freeze
```
igorpeshansky: Personally, I think this one is overkill, but I'll let @qingling128 make this call.

qingling128: Haha, i'm fine with defining an empty string constant. It could be shared by others as well though, not necessarily `Invalid trace id for stackdriver`.

PR #261, opened 2018-10-07
Use file path regex to determine current gem version.

qingling128:
Turned out the original implementation caused memory usage regression.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -233,9 +233,20 @@ module Fluent
     Fluent::Plugin.register_output('google_cloud', self)
 
     PLUGIN_NAME = 'Fluentd Google Cloud Logging plugin'.freeze
-    # Extract plugin version by finding the spec this file was loaded from.
-    PLUGIN_VERSION = __FILE__.match(
-      %r{fluent-plugin-google-cloud-?(?<version>[0-9a-zA-Z\.]*)/})['version']
+    PLUGIN_VERSION = begin
+      # Extract plugin version from file path.
+      __FILE__.match(
+        %r{fluent-plugin-google-cloud-(?<version>[0-9a-zA-Z\.]*)/})['version']
+    rescue StandardError
+      # Extract plugin version by finding the spec this file was loaded from.
+      dependency = Gem::Dependency.new('fluent-plugin-google-cloud')
+      all_specs, = Gem::SpecFetcher.fetcher.spec_for_dependency(dependency)
+      matching_spec, = all_specs.grep(
+        proc { |spec,| __FILE__.include?(spec.full_gem_path) }) do |spec,|
+          spec.version.to_s
+        end
+      matching_spec
+    end.freeze
 
     # Name of the the Google cloud logging write scope.
     LOGGING_SCOPE = 'https://www.googleapis.com/auth/logging.write'.freeze
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -235,7 +235,7 @@ module Fluent
     PLUGIN_NAME = 'Fluentd Google Cloud Logging plugin'.freeze
     # Extract plugin version by finding the spec this file was loaded from.
     PLUGIN_VERSION = __FILE__.match(
-      %r{fluent-plugin-google-cloud(?<version>[-0-9a-zA-Z\.]*)/})['version']
+      %r{fluent-plugin-google-cloud-?(?<version>[0-9a-zA-Z\.]*)/})['version']
 
     # Name of the the Google cloud logging write scope.
     LOGGING_SCOPE = 'https://www.googleapis.com/auth/logging.write'.freeze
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -234,15 +234,8 @@ module InternalConstants
...
-      dependency = Gem::Dependency.new('fluent-plugin-google-cloud')
-      all_specs, = Gem::SpecFetcher.fetcher.spec_for_dependency(dependency)
-      matching_spec, = all_specs.grep(
-        proc { |spec,| __FILE__.include?(spec.full_gem_path) }) do |spec,|
-          spec.version.to_s
-        end
-      matching_spec
-    end.freeze
+    PLUGIN_VERSION = __FILE__.match(
+      %r{fluent-plugin-google-cloud-?(?<version>[0-9a-zA-Z\.]*)/})['version']
```
qingling128: Note that when unit tests run, the `__FILE__` path does not have version embedded. It's simply `.../fluent-plugin-google-cloud/lib/fluent...`. So the regex has to be compatible to that.

igorpeshansky: Does this mean that in local tests we will have an empty version? Can we keep the original code as a fallback?

qingling128: It will not affect testing a gem locally. When installing a gem, it will be installed under the expected file path. It only affects unit tests. Is that something we are concerned with?

igorpeshansky: I think it makes sense to ensure that the plugin version always comes from the gem config, even when testing a local clone without an installed gem. WDYT?

qingling128: Updated. PTAL

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -233,8 +233,12 @@ module InternalConstants
     Fluent::Plugin.register_output('google_cloud', self)
 
     PLUGIN_NAME = 'Fluentd Google Cloud Logging plugin'.freeze
-    # Extract plugin version by finding the spec this file was loaded from.
     PLUGIN_VERSION = begin
+      # Extract plugin version from file path.
+      __FILE__.match(
+        %r{fluent-plugin-google-cloud-(?<version>[0-9a-zA-Z\.]*)/})['version']
+    rescue StandardError
```
igorpeshansky: This seems dangerous — there are many subclasses of `StandardError`... Why not:
```ruby
    PLUGIN_VERSION = begin
      # Extract plugin version from file path.
      match_data = __FILE__.match(
        %r{fluent-plugin-google-cloud-(?<version>[0-9a-zA-Z\.]*)/})
      if match_data
        match_data['version']
      else
        # Extract plugin version by finding the spec this file was loaded from.
        dependency = Gem::Dependency.new('fluent-plugin-google-cloud')
        all_specs, = Gem::SpecFetcher.fetcher.spec_for_dependency(dependency)
        matching_spec, = all_specs.grep(
           proc { |spec,| __FILE__.include?(spec.full_gem_path) }) do |spec,|
             spec.version.to_s
           end
        matching_spec
      end
    end.freeze
```
?

We could also extract this code into a private function, and then use an early `return` instead:
```ruby
    def extract_plugin_version
      # Extract plugin version from file path.
      match_data = __FILE__.match(
        %r{fluent-plugin-google-cloud-(?<version>[0-9a-zA-Z\.]*)/})
      return match_data['version'] if match_data
      # Extract plugin version by finding the spec this file was loaded from.
      dependency = Gem::Dependency.new('fluent-plugin-google-cloud')
      all_specs, = Gem::SpecFetcher.fetcher.spec_for_dependency(dependency)
      matching_spec, = all_specs.grep(
         proc { |spec,| __FILE__.include?(spec.full_gem_path) }) do |spec,|
           spec.version.to_s
         end
      matching_spec
    end
    PLUGIN_VERSION = extract_plugin_version.freeze
```

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -233,8 +233,12 @@ module InternalConstants
...
+      # Extract plugin version by finding the spec this file was loaded from.
       dependency = Gem::Dependency.new('fluent-plugin-google-cloud')
       all_specs, = Gem::SpecFetcher.fetcher.spec_for_dependency(dependency)
       matching_spec, = all_specs.grep(
```
igorpeshansky: [Optional] My fault — I forgot to rename `matching_spec`. Should have been `matching_version` instead.

qingling128:
Updated. PTAL @igorpeshansky @davidbtucker

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -233,15 +233,20 @@ module InternalConstants
     Fluent::Plugin.register_output('google_cloud', self)
 
     PLUGIN_NAME = 'Fluentd Google Cloud Logging plugin'.freeze
-    # Extract plugin version by finding the spec this file was loaded from.
+
     PLUGIN_VERSION = begin
+      # Extract plugin version from file path.
+      match_data = __FILE__.match(
+        %r{fluent-plugin-google-cloud-(?<version>[0-9a-zA-Z\.]*)/})
+      return match_data['version'] if match_data
```
igorpeshansky: I don't think you can `return` from a `begin`/`end` block... This is the pitfall of having different behaviors in unit tests and at runtime.

qingling128: Updated

PR #265, opened 2018-11-08
 Remove LogEntry fields from the record if their values are nil.

jkschulz:


test/plugin/base_test.rb
```diff
@@ -2144,6 +2156,22 @@ def verify_subfields_when_not_hash(payload_key)
...
+    setup_gce_metadata_stubs
+    setup_logging_stubs do
+      d = create_driver
+      d.emit(payload_key => nil)
+      d.run
+    end
+
+    verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
+      fields = get_fields(entry['jsonPayload'])
+      assert_nil fields[payload_key], entry[destination_key]
```
igorpeshansky: I think you want to check that `fields` doesn't have the key anymore, to distinguish from the case where it's left there as a `nil`. You also want to check that the field is not present at the `destination_key`. Plus, there's no need to print one part of the `entry` — just dump the whole thing. So something like:
```ruby
      assert not fields.key?(payload_key), entry
      assert_nil entry[destination_key], entry
```

bmoyles0117: +1

.rubocop_todo.yml
```diff
@@ -1,36 +1,34 @@
...
-  TargetRubyVersion: 2.1
-  Exclude:
-    # Generated files.
-    - lib/google/**/*
-
-# Offense count: 2
+# Offense count: 69
 Metrics/AbcSize:
-  Max: 150
+  Max: 117
```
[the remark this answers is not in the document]
jkschulz: Done.

test/plugin/base_test.rb
jkschulz: Checked that the keys didn't exist in either hash instead, and used assert_false.

igorpeshansky: FWIW, for the `entry` there's no difference between the key not existing and the key being set to `nil` — that is only user-visible in the `jsonPayload`. So I'd change the second `assert_false` to `assert_nil`, to be consistent with the rest of the tests (e.g., line 2155 above).

jkschulz: Changed all asserts for absent LogEntry fields to check for a missing key as opposed to a nil value, since ruby protos will only have keys if they've been explicitly set.

PR #268, opened 2018-11-22
Support log level setting per plugin

okkez:


jkohen:
@qingling128 what does it take to resolve this PR?

qingling128:
Looks like it needs rebasing. I tried the Github UI tool, but there is still a test failure. @okkez - Mind rebasing this off the latest master if it's still needed?

okkez:
Rubocop causes CI failure. I've fixed it.
It is necessary to configure the log level per plugin.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1132,15 +1130,15 @@ def set_vm_id
       @vm_id ||= fetch_gce_metadata('instance/id') if @platform == Platform::GCE
       @vm_id ||= ec2_metadata['instanceId'] if @platform == Platform::EC2
     rescue StandardError => e
-      @log.error 'Failed to obtain vm_id: ', error: e
+      log.error 'Failed to obtain vm_id: ', error: e
```
qingling128: Curious: How did we decide which cases are `log`, and which cases are `$log`?

igorpeshansky: It's hidden in the extra part of the commit message on https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/268/commits/63387b523f3ee703a2ed0cf702d14b87891500cc. Frankly, I'd prefer it as a code comment in `configure`.

okkez:
`$log` is a global variable, so we cannot control its log level per plugin, we can control it in global.
In the `configure` method, we have to use `$log` to output logs, because the logger instance (we use `log` method to fetch it) per plugin has not been configured while invoking `configure` method.

AFAIK, there is no document about `$log` and `log`.
But I can find the document to understand the log level per plugin.
See https://docs.fluentd.org/deployment/logging.

PR #273, opened 2018-12-21
Add httpRequest, sourceLocation and operation keys to preserved_keys.

qingling128:
With `detect_json` true, when the record has only 1 effective field (named 'log', 'message' or 'msg') and the field is in JSON format, the field should be parsed as JSON and sent as jsonPayload. Fields with names in the `preserved_keys` list should not be considered when counting the number of effective fields.


test/plugin/base_test.rb
```diff
@@ -612,6 +612,30 @@ def test_structured_payload_json_log_detect_json_container_parsed
       end
   end
 
+  # Verify that when the log has only one effective field (named 'log',
+  # 'message' or 'msg') and the field is in JSON format, the field is parsed as
+  # JSON and sent as jsonPayload.
```
jkohen: What should happen if more than one of msg, log, or message are defined in the input? Is that possible?

qingling128: It is possible. If that is the case, we will not send the field (named 'msg', 'log' or 'message') as `jsonPayload`. Instead, it will be treated as a `textPayload`. There is some inconsistent behavior regarding to `detect_json`. We have a separate ticket for it: b/119681841.

jkohen: Do we have a test for that behavior? If not, should we add one?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -609,7 +609,14 @@ def write(chunk)
...
             preserved_keys = [
-              'time', 'severity', @trace_key, @span_id_key, @insert_id_key
+              'time',
+              'severity',
+              @http_request_key,
+              @insert_id_key,
+              @source_location_key,
+              @span_id_key,
+              @trace_key,
+              @operation_key
```
igorpeshansky: Did you mean for this to be in alphabetical order?

test/plugin/base_test.rb
igorpeshansky: If more than one field is present in the original record, wouldn't JSON detection not kick in at all?

test/plugin/base_test.rb
```diff
@@ -612,6 +612,30 @@ def test_structured_payload_json_log_detect_json_container_parsed
       end
   end
 
+  # Verify that when the log has only one effective field (named 'log',
+  # 'message' or 'msg') and the field is in JSON format, the field is parsed as
```
igorpeshansky: Oxford comma, please...

test/plugin/constants.rb
```diff
@@ -950,4 +950,15 @@ module Constants
       'grpc-status-details-bin' => status_details.to_proto
     }.freeze
   end
+
+  PERSERVED_KEYS_MAP = {
```
igorpeshansky: Typo: `PERSERVED`.

test/plugin/base_test.rb
igorpeshansky: I meant the comma before the `or`... The "parsed as JSON and sent as jsonPayload" didn't need a comma.

test/plugin/base_test.rb
```diff
@@ -1219,6 +1243,8 @@ def test_cloud_dataflow_log
     verify_log_entries(1, DATAFLOW_PARAMS)
   end
 
+  # Verify the subfields extraction of LogEntry fields.
```
igorpeshansky: IMO, it reads better to say either "subfield extraction" or "extraction of subfields"... Also below.

test/plugin/constants.rb
```diff
@@ -748,7 +756,13 @@ module Constants
   SOURCE_LOCATION_MESSAGE = {
     'file' => 'source/file',
     'function' => 'my_function',
-    'line' => 18
```
igorpeshansky: Good catch!

test/plugin/constants.rb
```diff
@@ -950,4 +970,15 @@ module Constants
...
+
+  PRESERVED_KEYS_MAP = {
+    'time' => K8S_TIMESTAMP,
+    'severity' => CONTAINER_SEVERITY,
+    DEFAULT_HTTP_REQUEST_KEY => HTTP_REQUEST_MESSAGE,
+    DEFAULT_OPERATION_KEY => OPERATION_MESSAGE,
+    DEFAULT_SOURCE_LOCATION_KEY => SOURCE_LOCATION_MESSAGE,
+    DEFAULT_TRACE_KEY => TRACE,
+    DEFAULT_SPAN_ID_KEY => SPAN_ID,
+    DEFAULT_INSERT_ID_KEY => INSERT_ID
```
igorpeshansky: [Optional] You seem to have reordered the tests in alphabetical order. May make sense to reorder these too...

test/plugin/base_test.rb
qingling128: This is the LogEntry fields' subfields' extraction.
e.g.
LogEntry -> sourceLocation -> line.

So I phrased it this way. `LogEntry fields' subfields' extraction` feels weird.

test/plugin/constants.rb
qingling128: Done

test/plugin/base_test.rb
```diff
@@ -2226,9 +2288,36 @@ def verify_cascading_json_detection_with_log_entry_fields(
         d.run
       end
       verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
-        assert_equal expected_value, entry[log_entry_field],
-                     "Index #{index} failed. #{expected_value} is expected" \
-                     " for #{log_entry_field} field."
+        if subfields_with_default
+          # For an optional field with default values, Protobuf omits the field
+          # when it is deserialized to json. So we need to use
```
igorpeshansky: Nit: JSON.

test/plugin/base_test.rb
```diff
@@ -2226,9 +2288,36 @@ def verify_cascading_json_detection_with_log_entry_fields(
         d.run
       end
       verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
-        assert_equal expected_value, entry[log_entry_field],
-                     "Index #{index} failed. #{expected_value} is expected" \
-                     " for #{log_entry_field} field."
+        if subfields_with_default
```
igorpeshansky: This is getting really complex, and very likely doesn't belong in the base test implementation. Normally, we'd leave a TODO and open a bug to refactor later so as not to hold up this PR, but I'll let you decide what the right course of action is here.

test/plugin/base_test.rb
```diff
@@ -2283,20 +2372,32 @@ def verify_field_key(log_entry_field, default_key, custom_key,
         payload_fields = get_fields(entry['jsonPayload'])
         assert_equal input[:expected_payload].size, payload_fields.size, input
         payload_fields.each do |key, value|
-          assert_equal input[:expected_payload][key], get_string(value), input
+          # TODO(qingling128) Refactor this into a more general method that can
+          # be reused everywhere for JSON and gRPC deserialization comparison.
+          if value.is_a?(Hash) && value.key?('structValue')
```
igorpeshansky: Same here — really complex and needs refactoring.

qingling128:
Turned out we missed some test coverage for LogEntry fields that have subfields. Existing test util assumed them to be strings. I tried to add some tests in the same PR. But that triggered a cascading refactoring which made the PR really big... I've reverted the last few commits. Gonna split them into a separate PR.

PR #274, opened 2018-12-27
Add more unit tests for LogEntry fields.

qingling128:
Follow up of https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/273.

### DIFF
--- a/test/plugin/test_out_google_cloud_grpc.rb
+++ b/test/plugin/test_out_google_cloud_grpc.rb
@@ -466,8 +466,78 @@ class GoogleCloudOutputGRPCTest < Test::Unit::TestCase
     field['numberValue']
   end
 
+  def get_bool(field)
+    field['boolValue']
+  end
+
   # The null value.
   def null_value
     { 'nullValue' => 'NULL_VALUE' }
+  end
+
+  def http_request_message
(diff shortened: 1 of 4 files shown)


test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -466,8 +466,78 @@ def get_number(field)
...
+  #         "function" => {
+  #           "stringValue" => "my_function"
+  #         },
+  #         "line" => {
+  #           "numberValue" => 18
+  #         }
+  #       }
+  #     }
+  #   }
+  def assert_json_equal(expected, actual)
```
jkohen: assert_json_equal_proto highlights the fact that these are different structures.

qingling128: I thought about this, then realized this method name has to be shared between gRPC and REST. The REST side does not involve proto. I just added some comments for clarification.

jkohen: How about hash_equal_json or something where the reader understands at the call location, without having to read comments?

qingling128: Just renamed this to `assert_hash_equal_json`. WDYT?

PR #275, opened 2018-12-28
Add field extraction for httpRequest.serverIp.

qingling128:
Currently server IP is not extracted correctly, resulting in log entry like below:

```
{
 httpRequest: {
  requestMethod:  "GET"   
  requestSize:  "777"   
  requestUrl:  "/"   
  status:  200   
  userAgent:  "Opera/12.0"   
 }
 jsonPayload: {
  httpRequest: {
   serverIp:  "192.168.0.1"    
  }
  ...
 }
 ...
}
```

### DIFF
--- a/test/plugin/constants.rb
+++ b/test/plugin/constants.rb
@@ -748,6 +748,7 @@ module Constants
     'responseSize' => 65,
     'userAgent' => 'USER AGENT 1.0',
     'remoteIp' => '55.55.55.55',
+    'serverIp' => '66.66.66.66',
     'referer' => 'http://referer/',
     'cacheHit' => true,
     'cacheValidatedWithOriginServer' => true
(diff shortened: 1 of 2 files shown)


test/plugin/constants.rb
```diff
@@ -748,6 +748,7 @@ module Constants
     'responseSize' => 65,
     'userAgent' => 'USER AGENT 1.0',
     'remoteIp' => '55.55.55.55',
+    'serverIp' => '66.66.66.66',
```
jkohen: It feels like you were planning to add a test. Is the file missing from this branch?

jkschulz: The tests iterate through all fields defined in these maps (so you don't have to hard-code a new test anytime a new field is added), so this effectively does add tests.

jkohen: Oh, I see. Thanks!

qingling128: Yup, exactly what @jkschulz said! Isn't that great? 😃

jkohen: I was going to stay silent, but since you asked... unfortunately, I don't think so. Data-driven tests are tricky. If the test gets broken and doesn't test some or all of the fields, you'll never know, meaning your coverage will drop silently.

PR #281, opened 2019-01-11
Add initial support for /statusz page.

davidbtucker:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -707,6 +728,29 @@ def write(chunk)
 
     private
 
+    def statusz_response
+      # TODO(davidbtucker): Add more status information here.
```
qingling128: Do we expect the amount of code for `statusz` would be significant? I wonder if it would be easier to maintain by splitting `statusz` related methods into a separate module/file like https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/master/lib/fluent/plugin/monitoring.rb

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -707,6 +728,29 @@ def write(chunk)
...
+        "<html>",
+        "<body>",
+        "<h1>Status</h1>",
+        "<b>Command-line flags:</b> #{CGI::escapeHTML(ARGV.join(' '))}",
+        "</body>",
+        "</html>"
+      ].join("\n") + "\n"
+    end
+
+    def write_html_response(data, conn, code, response)
```
qingling128: If `write_html_response` somehow fails, would it crash the agent by any chance? Just checking.

lib/fluent/plugin/out_google_cloud.rb
davidbtucker: Sure, I split out the function for creating the page into a separate module.

lib/fluent/plugin/out_google_cloud.rb
davidbtucker: Ah, what kind of failures are you thinking of? In writing to the network connection? Or some other bug that would cause a crash?

qingling128: I was thinking about some error when writing to the network connection and whether this would affect the agent's ability to ingest logs.

davidbtucker: Ok, I tested this out by writing a bad value to the connection when handling /statusz. I see the error message from fluentd for the /statusz write, and then subsequent logs still go through.

PR #292, opened 2019-02-13
 Add logEntry->labels auto extraction.

qingling128:
Extract `logging.googleapis.com/labels` (the field name can be customized by the configuration `labels_key`) and set those labels in logEntry->labels.

If a `logging.googleapis.com/labels` field (or a customized field whose name is specified by `labels_key`) is present in the log entry, the field is stripped from the payload and assigned to the [LogEntry](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry) labels field.

This is similar to how we handle the other fields for logEntry: https://cloud.google.com/logging/docs/agent/configuration#special-fields

Example:

Original log entry
```
{ 
  "message":  "this is a test message",
  "code": "user",
  "logging.googleapis.com/labels": {
    "app":  "logging-load-test",     
    "node":  "default-pool-3edf6d28-v1np",     
    "pod-template-hash":  "7029"   
  }  
 }
```
logEntry->jsonPayload
```
{ 
  "message":  "this is a test message",
  "code": "user"
}
```
logEntry->labels
```
{
    "app":  "logging-load-test",     
    "node":  "default-pool-3edf6d28-v1np",     
    "pod-template-hash":  "7029"   
}  
```

Note that there are 3 ways of passing in labels now. In case of label name collision, the label value respects the following priority order:
1. Labels from the field `logging.googleapis.com/labels` in payload.
2. Labels from the config `label_map`.
3. Labels from the config `labels`.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -146,12 +146,12 @@ module Fluent
       # Default values for JSON payload keys to set the "httpRequest",
       # "operation", "sourceLocation", "trace" fields in the LogEntry.
       DEFAULT_HTTP_REQUEST_KEY = 'httpRequest'.freeze
+      DEFAULT_INSERT_ID_KEY = 'logging.googleapis.com/insertId'.freeze
       DEFAULT_OPERATION_KEY = 'logging.googleapis.com/operation'.freeze
       DEFAULT_SOURCE_LOCATION_KEY =
         'logging.googleapis.com/sourceLocation'.freeze
-      DEFAULT_TRACE_KEY = 'logging.googleapis.com/trace'.freeze
       DEFAULT_SPAN_ID_KEY = 'logging.googleapis.com/spanId'.freeze
-      DEFAULT_INSERT_ID_KEY = 'logging.googleapis.com/insertId'.freeze
+      DEFAULT_TRACE_KEY = 'logging.googleapis.com/trace'.freeze
 
       DEFAULT_METADATA_AGENT_URL =
         'http://local-metadata-agent.stackdriver.com:8000'.freeze
@@ -204,16 +204,6 @@ module Fluent
           # The non-grpc version class name.
           'Google::Apis::LoggingV2::HttpRequest'
         ],
-        'source_location' => [
-          '@source_location_key',
-          [
-            %w(file file parse_string),
-            %w(function function parse_string),
-            %w(line line parse_int)
-          ],
-          'Google::Logging::V2::LogEntrySourceLocation',
-          'Google::Apis::LoggingV2::LogEntrySourceLocation'
-        ],
         'operation' => [
           '@operation_key',
           [
@@ -224,6 +214,16 @@ module Fluent
           ],
           'Google::Logging::V2::LogEntryOperation',
           'Google::Apis::LoggingV2::LogEntryOperation'
+        ],
+        'source_location' => [
+          '@source_location_key',
+          [
+            %w(file file parse_string),
+            %w(function function parse_string),
+            %w(line line parse_int)
+          ],
+          'Google::Logging::V2::LogEntrySourceLocation',
+          'Google::Apis::LoggingV2::LogEntrySourceLocation'
         ]
       }.freeze
 
@@ -298,12 +298,12 @@ module Fluent
     # Map keys from a JSON payload to corresponding LogEntry fields.
     config_param :http_request_key, :string, :default =>
       DEFAULT_HTTP_REQUEST_KEY
+    config_param :insert_id_key, :string, :default => DEFAULT_INSERT_ID_KEY
     config_param :operation_key, :string, :default => DEFAULT_OPERATION_KEY
     config_param :source_location_key, :string, :default =>
       DEFAULT_SOURCE_LOCATION_KEY
-    config_param :trace_key, :string, :default => DEFAULT_TRACE_KEY
     config_param :span_id_key, :string, :default => DEFAULT_SPAN_ID_KEY
-    config_param :insert_id_key, :string, :default => DEFAULT_INSERT_ID_KEY
+    config_param :trace_key, :string, :default => DEFAULT_TRACE_KEY
 
     # Whether to try to detect if the record is a text log entry with JSON
     # content that needs to be parsed.
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1697,6 +1697,10 @@ module Fluent
     end
 
     def set_log_entry_fields(record, entry)
+      # TODO(qingling128) On the next major after 0.7.4, make all logEntry
+      # subfields behave the same way: if the field is not in the correct
+      # format, log an error in the Fluentd log and remove this field from
+      # payload. This is the preferred behavior per PM decision.
       LOG_ENTRY_FIELDS_MAP.each do |field_name, config|
         payload_key, subfields, grpc_class, non_grpc_class = config
         begin
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1745,13 +1745,13 @@ module Fluent
       payload_labels = record.delete(@labels_key)
       unless payload_labels.is_a?(Hash)
         @log.error "Invalid value of '#{@labels_key}' in the payload: " \
-                   "#{payload_labels}. Labels need to be a JSON object.", err
+                   "#{payload_labels}. Labels need to be a JSON object."
         return nil
       end
       payload_labels.each do |k, v|
         next if k.is_a?(String) && v.is_a?(String)
-        @log.error "Invalid value of '#{@labels_key}' in the payload. " \
-                   "#{payload_labels}. Labels need to have string values.", err
+        @log.error "Invalid value of '#{@labels_key}' in the payload: " \
+                   "#{payload_labels}. Labels need to have string values."
         return nil
       end
       payload_labels
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1743,6 +1743,7 @@ module Fluent
     # Parse labels. Return nil if not set.
     def parse_labels(record)
       payload_labels = record.delete(@labels_key)
+      return nil unless payload_labels
       unless payload_labels.is_a?(Hash)
         @log.error "Invalid value of '#{@labels_key}' in the payload: " \
                    "#{payload_labels}. Labels need to be a JSON object."
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -147,6 +147,7 @@ module Fluent
       # "operation", "sourceLocation", "trace" fields in the LogEntry.
       DEFAULT_HTTP_REQUEST_KEY = 'httpRequest'.freeze
       DEFAULT_INSERT_ID_KEY = 'logging.googleapis.com/insertId'.freeze
+      DEFAULT_LABELS_KEY = 'logging.googleapis.com/labels'.freeze
       DEFAULT_OPERATION_KEY = 'logging.googleapis.com/operation'.freeze
       DEFAULT_SOURCE_LOCATION_KEY =
         'logging.googleapis.com/sourceLocation'.freeze
@@ -299,6 +300,7 @@ module Fluent
     config_param :http_request_key, :string, :default =>
       DEFAULT_HTTP_REQUEST_KEY
     config_param :insert_id_key, :string, :default => DEFAULT_INSERT_ID_KEY
+    config_param :labels_key, :string, :default => DEFAULT_LABELS_KEY
     config_param :operation_key, :string, :default => DEFAULT_OPERATION_KEY
     config_param :source_location_key, :string, :default =>
       DEFAULT_SOURCE_LOCATION_KEY
@@ -620,6 +622,7 @@ module Fluent
               'severity',
               @http_request_key,
               @insert_id_key,
+              @labels_key,
               @operation_key,
               @source_location_key,
               @span_id_key,
@@ -652,6 +655,11 @@ module Fluent
             entry_level_resource.type, record, time)
           severity = compute_severity(
             entry_level_resource.type, record, entry_level_common_labels)
+
+          dynamic_labels_from_payload = parse_labels(record)
+
+          entry_level_common_labels = entry_level_common_labels.merge!(
+            dynamic_labels_from_payload) if dynamic_labels_from_payload
 
           entry = @construct_log_entry.call(entry_level_common_labels,
                                             entry_level_resource,
@@ -1730,6 +1738,29 @@ module Fluent
           @log.error "Failed to set log entry field for #{field_name}.", err
         end
       end
+    end
+
+    # Parse labels. Return nil if not set.
+    def parse_labels(record)
+      payload_value = record.delete(@labels_key)
+      unless payload_value.is_a?(Hash)
+        @log.error 'Failed to set log entry field for labels. Labels need ' \
+                   'to be a hash where both keys and values are strings.' \
+                   "\nValue: #{payload_value}."
+        return nil
+      end
+      payload_value.each do |k, v|
+        next if k.is_a?(String) && v.is_a?(String)
+        @log.error 'Failed to set log entry field for labels. Labels need ' \
+                     'to be a hash where both keys and values are strings.' \
+                     "\nValue: #{payload_value}."
+        return nil
+      end
+      payload_value
+    rescue StandardError => err
+      @log.error 'Failed to set log entry field for labels.' \
+                 "\nValue: #{payload_value}", err
+      return nil
     end
 
     # Values permitted by the API for 'severity' (which is an enum).
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1742,24 +1742,21 @@ module Fluent
 
     # Parse labels. Return nil if not set.
     def parse_labels(record)
-      payload_value = record.delete(@labels_key)
-      unless payload_value.is_a?(Hash)
-        @log.error 'Failed to set log entry field for labels. Labels need ' \
-                   'to be a hash where both keys and values are strings.' \
-                   "\nValue: #{payload_value}."
+      payload_labels = record.delete(@labels_key)
+      unless payload_labels.is_a?(Hash)
+        @log.error "Invalid value of '#{@labels_key}' in the payload: " \
+                   "#{payload_labels}. Labels need to be a JSON object.", err
         return nil
       end
-      payload_value.each do |k, v|
+      payload_labels.each do |k, v|
         next if k.is_a?(String) && v.is_a?(String)
-        @log.error 'Failed to set log entry field for labels. Labels need ' \
-                     'to be a hash where both keys and values are strings.' \
-                     "\nValue: #{payload_value}."
+        @log.error "Invalid value of '#{@labels_key}' in the payload. " \
+                   "#{payload_labels}. Labels need to have string values.", err
         return nil
       end
-      payload_value
+      payload_labels
     rescue StandardError => err
-      @log.error 'Failed to set log entry field for labels.' \
-                 "\nValue: #{payload_value}", err
+      @log.error "Failed to extract '#{@labels_key}' from payload.", err
       return nil
     end
 
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1749,10 +1749,14 @@ module Fluent
                    "#{payload_labels}. Labels need to be a JSON object."
         return nil
       end
-      payload_labels.each do |k, v|
-        next if k.is_a?(String) && v.is_a?(String)
+
+      non_string_keys = payload_labels.each_with_object([]) do |(k, v), a|
+        a << k unless k.is_a?(String) && v.is_a?(String)
+      end
+      unless non_string_keys.empty?
         @log.error "Invalid value of '#{@labels_key}' in the payload: " \
-                   "#{payload_labels}. Labels need to have string values."
+                   "#{payload_labels}. Labels need string values for all " \
+                   "keys; keys #{non_string_keys} don't."
         return nil
       end
       payload_labels
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -1287,19 +1287,34 @@ module BaseTest
   # Verify the subfields extraction of LogEntry fields when they are not hashes.
 
   def test_log_entry_http_request_field_when_not_hash
-    verify_subfields_when_not_hash(DEFAULT_HTTP_REQUEST_KEY)
+    # Leave the malformed field as it is.
+    # TODO(qingling128) On the next major after 0.7.4, make the rest of
+    # logEntry subfields behave the same way: if the field is not a hash as
+    # expected, log an error in the Fluentd log and remove this field from
+    # payload. This is the preferred behavior per PM decision.
+    verify_subfields_untouched_when_not_hash(DEFAULT_HTTP_REQUEST_KEY)
   end
 
   def test_log_entry_labels_field_when_not_hash
-    verify_subfields_when_not_hash(DEFAULT_LABELS_KEY)
+    verify_subfields_removed_when_not_hash(DEFAULT_LABELS_KEY)
   end
 
   def test_log_entry_operation_field_when_not_hash
-    verify_subfields_when_not_hash(DEFAULT_OPERATION_KEY)
+    # Leave the malformed field as it is.
+    # TODO(qingling128) On the next major after 0.7.4, make the rest of
+    # logEntry subfields behave the same way: if the field is not a hash as
+    # expected, log an error in the Fluentd log and remove this field from
+    # payload. This is the preferred behavior per PM decision.
+    verify_subfields_untouched_when_not_hash(DEFAULT_OPERATION_KEY)
   end
 
   def test_log_entry_source_location_field_when_not_hash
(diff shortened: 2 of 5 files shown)


qingling128:
Sure thing. Shared via internal channel.

test/plugin/base_test.rb
```diff
@@ -1424,6 +1454,81 @@ def test_cascading_json_detection_with_log_entry_trace_field
       nested_level_value: TRACE2)
   end
 
+  # Expected order.
+  # 1. Labels from the field "logging.googleapis.com/labels" in payload.
+  # 2. Labels from the config "label_map".
+  # 3. Labels from the config "labels".
+  def test_labels_order
```
jkohen: It's not clear to me what this is testing. Is this a priority order for labels provided with different mechanisms, that are all passed on in the same field? If so, I'd be more direct:
Verify that labels present in multiple inputs respect the expected priority order:
<then your list>

test/plugin/base_test.rb
```diff
@@ -2292,9 +2404,22 @@ def verify_subfields_when_not_hash(payload_key)
...
-      assert_false entry.key?(destination_key), entry
+      if payload_key == DEFAULT_LABELS_KEY
+        # The malformed field has been removed from the payload.
+        assert_true get_fields(entry['jsonPayload']).empty?, entry
+        # No additional labels.
+        assert_equal COMPUTE_PARAMS[:labels].size,
+                     entry[destination_key].size, entry
+      else
+        # Leave the malformed field as it is.
+        # TODO(qingling128) In the next breaking change release, make the older
```
jkohen: breaking change -> "on the next major after x.y.z", so when someone looks at this in the future the can more easily make a decision whether it's OK to make this change.

test/plugin/base_test.rb
qingling128: Yup, that's exactly what the test is trying to verify. I've changed the test description accordingly.

test/plugin/base_test.rb
qingling128: Makes sense. Changed.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1732,6 +1740,29 @@ def set_log_entry_fields(record, entry)
       end
     end
 
+    # Parse labels. Return nil if not set.
+    def parse_labels(record)
+      payload_value = record.delete(@labels_key)
+      unless payload_value.is_a?(Hash)
+        @log.error 'Failed to set log entry field for labels. Labels need ' \
```
igorpeshansky: Nit: this method does not set anything — it simply extracts the labels value from the payload.
Another nit: the payload is technically a JSON object, so "hash" is meaningless to the user (it's a Ruby concept), and JSON object keys are guaranteed to be strings.
I would instead output an error like `"Invalid value of "#{@labels_key}" in the payload. Labels need to be a JSON object with string values."`.
You could also vary the message depending on which validation step fails, e.g.:
- if not a hash: error `"Invalid value of '#{@labels_key}' in the payload. Labels need to be a JSON object."`;
- if not a string value: error `"Invalid value of '#{@labels_key}' in the payload. Labels need to have string values."`;
- rescue err: error `"Failed to extract '#{@labels_key}' from payload."`, err

test/plugin/base_test.rb
```diff
@@ -1249,57 +1251,73 @@ def test_log_entry_http_request_field_from_record
...
   # fields.
 
   def test_log_entry_http_request_field_partial_from_record
     verify_subfields_partial_from_record(DEFAULT_HTTP_REQUEST_KEY)
   end
 
-  def test_log_entry_source_location_field_partial_from_record
-    verify_subfields_partial_from_record(DEFAULT_SOURCE_LOCATION_KEY)
-  end
+  # Skip labels for verify_subfields_partial_from_record because labels are free
```
igorpeshansky: What code is this comment attached to?

test/plugin/base_test.rb
```diff
@@ -2374,7 +2506,12 @@ def verify_cascading_json_detection_with_log_entry_fields(
         d.emit(input_log_entry)
         d.run
       end
-      verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
+      expected_params = COMPUTE_PARAMS.dup
+      if log_entry_field == 'labels'
+        expected_value = expected_value.merge(COMPUTE_PARAMS[:labels])
```
igorpeshansky: Are you sure this `merge` is in the right direction? This would make the label values from `COMPUTE_PARAMS[:labels]` override the ones in `expected_value`...

test/plugin/base_test.rb
```diff
@@ -2374,7 +2506,12 @@ def verify_cascading_json_detection_with_log_entry_fields(
         d.emit(input_log_entry)
         d.run
       end
-      verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
+      expected_params = COMPUTE_PARAMS.dup
+      if log_entry_field == 'labels'
```
igorpeshansky: Why are we doing this here, rather than in the invocation of this method? It seems like it would be better to define `COMPUTE_PARAMS` without labels and then verify the labels in the supplied block...

test/plugin/base_test.rb
```diff
@@ -2427,7 +2564,13 @@ def verify_field_key(log_entry_field, default_key, custom_key,
         d.emit(input[:emitted_log])
         d.run
       end
-      verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
+      expected_params = COMPUTE_PARAMS.dup
+      if log_entry_field == 'labels'
+        input[:expected_field_value] = COMPUTE_PARAMS[:labels].merge(
+          input[:expected_field_value] || {})
+        expected_params[:labels] = input[:expected_field_value]
```
igorpeshansky: Is it `:labels` or `'labels'`?

test/plugin/base_test.rb
qingling128: Unless we refactor the `verify_json_log_entries` method, we could not stop `verify_log_entries` from verifying `labels`.

One thing we could do (in fact I started it like this but decided not to do it) is to add two params: `expected_params_from_root` and `expected_params_from_nested`. We can leave the default value as `COMPUTE_PARAMS`, and pass in customized value when testing `labels`. e.g.:

```
  def test_cascading_json_detection_with_log_entry_labels_field
     verify_cascading_json_detection_with_log_entry_fields(
       'labels', DEFAULT_LABELS_KEY,
       root_level_value: LABELS_MESSAGE,
-      nested_level_value: LABELS_MESSAGE2)
+      nested_level_value: LABELS_MESSAGE2,
+      expected_value_from_root: LABELS_MESSAGE.merge(COMPUTE_PARAMS[:labels]),
+      expected_value_from_nested: LABELS_MESSAGE2.merge(
+        COMPUTE_PARAMS[:labels]),
+      expected_params_from_root: COMPUTE_PARAMS.merge(
+        labels: COMPUTE_PARAMS[:labels].merge(LABELS_MESSAGE)),
+      expected_params_from_nested: COMPUTE_PARAMS.merge(
+        labels: COMPUTE_PARAMS[:labels].merge(LABELS_MESSAGE2)))
   end

   def test_cascading_json_detection_with_log_entry_operation_field
@@ -2460,6 +2467,10 @@ module BaseTest
       :expected_value_from_root, root_level_value)
     expected_value_from_nested = expectation.fetch(
       :expected_value_from_nested, nested_level_value)
+    expected_params_from_root = expectation.fetch(
+      :expected_params_from_root, COMPUTE_PARAMS)
+    expected_params_from_nested = expectation.fetch(
+      :expected_params_from_nested, COMPUTE_PARAMS)

     setup_gce_metadata_stubs

@@ -2498,21 +2509,27 @@ module BaseTest
       default_key => root_level_value)

     {
-      log_entry_with_root_level_field => expected_value_from_root,
-      log_entry_with_nested_level_field => expected_value_from_nested,
-      log_entry_with_both_level_fields => expected_value_from_nested
-    }.each_with_index do |(input_log_entry, expected_value), index|
+      log_entry_with_root_level_field => {
+        expected_value: expected_value_from_root,
+        expected_params: expected_params_from_root
+      },
+      log_entry_with_nested_level_field => {
+        expected_value: expected_value_from_nested,
+        expected_params: expected_params_from_nested
+      },
+      log_entry_with_both_level_fields => {
+        expected_value: expected_value_from_nested,
+        expected_params: expected_params_from_nested
+      }
+    }.each_with_index do |(input_log_entry, expected), index|
+      expected_params = expected[:expected_params]
+      expected_value = expected[:expected_value]
       setup_logging_stubs do
         @logs_sent = []
         d = create_driver(DETECT_JSON_CONFIG)
         d.emit(input_log_entry)
         d.run
       end
-      expected_params = COMPUTE_PARAMS.dup
-      if log_entry_field == 'labels'
-        expected_value = expected_value.merge(COMPUTE_PARAMS[:labels])
-        expected_params[:labels] = expected_value
-      end
```

I chose not to do it because it increased complexity and is kind of distracting. `expected_value` is what we need to focus on. `expected_params` just happens to be something we need to placate.

Does that make sense?

test/plugin/base_test.rb
qingling128: The test data has no overlap, so it should not matter. But I changed it anyway to match the actual precedence.

test/plugin/base_test.rb
qingling128: `:labels` because:

```
  COMPUTE_PARAMS_NO_LOG_NAME = {
    resource: {
      type: COMPUTE_CONSTANTS[:resource_type],
      labels: {
        'instance_id' => VM_ID,
        'zone' => ZONE
      }
    },
    project_id: PROJECT_ID,
    labels: {
    }
  }.freeze
  COMPUTE_PARAMS = COMPUTE_PARAMS_NO_LOG_NAME.merge(
    log_name: 'test'
  ).freeze
```

test/plugin/base_test.rb
```diff
@@ -1249,57 +1251,75 @@ def test_log_entry_http_request_field_from_record
...
 
   def test_log_entry_http_request_field_partial_from_record
     verify_subfields_partial_from_record(DEFAULT_HTTP_REQUEST_KEY)
   end
 
-  def test_log_entry_source_location_field_partial_from_record
-    verify_subfields_partial_from_record(DEFAULT_SOURCE_LOCATION_KEY)
-  end
+  # We don't need a test like 'test_log_entry_labels_field_partial_from_record'
+  # becasue labels are free range strings. Everything in the labels field should
```
igorpeshansky: Typo: becasue.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1732,6 +1740,26 @@ def set_log_entry_fields(record, entry)
...
+    def parse_labels(record)
+      payload_labels = record.delete(@labels_key)
+      unless payload_labels.is_a?(Hash)
+        @log.error "Invalid value of '#{@labels_key}' in the payload: " \
+                   "#{payload_labels}. Labels need to be a JSON object.", err
+        return nil
+      end
+      payload_labels.each do |k, v|
+        next if k.is_a?(String) && v.is_a?(String)
+        @log.error "Invalid value of '#{@labels_key}' in the payload. " \
```
igorpeshansky: Was this supposed to end in a ':' instead of a '.'?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1732,6 +1740,26 @@ def set_log_entry_fields(record, entry)
       end
     end
 
+    # Parse labels. Return nil if not set.
+    def parse_labels(record)
+      payload_labels = record.delete(@labels_key)
+      unless payload_labels.is_a?(Hash)
+        @log.error "Invalid value of '#{@labels_key}' in the payload: " \
+                   "#{payload_labels}. Labels need to be a JSON object.", err
```
igorpeshansky: What's `err`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1732,6 +1740,26 @@ def set_log_entry_fields(record, entry)
...
+        return nil
+      end
+      payload_labels.each do |k, v|
+        next if k.is_a?(String) && v.is_a?(String)
+        @log.error "Invalid value of '#{@labels_key}' in the payload. " \
+                   "#{payload_labels}. Labels need to have string values.", err
```
igorpeshansky: What's `err`?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1732,6 +1740,26 @@ def set_log_entry_fields(record, entry)
...
 
+    # Parse labels. Return nil if not set.
+    def parse_labels(record)
+      payload_labels = record.delete(@labels_key)
+      unless payload_labels.is_a?(Hash)
+        @log.error "Invalid value of '#{@labels_key}' in the payload: " \
+                   "#{payload_labels}. Labels need to be a JSON object.", err
+        return nil
+      end
+      payload_labels.each do |k, v|
```
igorpeshansky: [Optional] The following might be somewhat more helpful to users:
```ruby
non_string_keys = payload_labels.each_with_object([]) do |k, v, a|
  a << k unless k.is_a?(String) && v.is_a?(String)
end
unless non_string_keys.empty?
  @log.error "Invalid value of '#{@labels_key}' in the payload: " \
             "#{payload_labels}. Labels need string values for all keys; " \
             "keys #{non_string_keys} don't."
  return nil
end
```

test/plugin/base_test.rb
igorpeshansky: Note that I suggested removing labels from `COMPUTE_PARAMS`, which would make the expected and actual params identical. You can then use the block within the verification step to verify the labels separately. WDYT?

test/plugin/base_test.rb
igorpeshansky: Forgot this one. Sigh, this is confusing — the keys in `expected_params` are symbols, but the keys in `entry` are strings... Oh, well...

test/plugin/base_test.rb
qingling128: Fixed.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Oops, copied it over by accident. Removed.

lib/fluent/plugin/out_google_cloud.rb
qingling128: Yup. Changed.

lib/fluent/plugin/out_google_cloud.rb

test/plugin/base_test.rb
qingling128: How would we bypass [this check](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/946798d787bef1ef2e38ec8ec327bdd4ac7e4bce/test/plugin/base_test.rb#L2203) though? The actual params has `LABELS_MESSAGE.merge(COMPUTE_PARAMS[:labels])` as the labels. Removing labels from `COMPUTE_PARAMS` won't help, right? Unless we want to refactor `verify_json_log_entries`.

lib/fluent/plugin/out_google_cloud.rb
qingling128: I'm leaning towards the current implementation, which exits early and performs minimum calculation for performance concerns. Since we will print `payload_labels`, it should be sufficient for users to figure out what went wrong.

test/plugin/base_test.rb
igorpeshansky: I'm actually leaning towards refactoring `verify_json_log_entries`, because it conflates two checks — the fact that certain labels are present _somewhere_ in the write logs request, and the fact that the individual log entry contains specific labels. One fix would be to change `check_labels` to only verify the existence of the expected labels, rather than full equality of the label set. We might need to add extra checks for the resource labels (to ensure that there's exact equality there). WDYT?

qingling128: Refactored `verify_json_log_entries`. I added a flag `check_exact_entry_labels`. Also refactored `check_labels` a little bit in the end.

test/plugin/base_test.rb
```diff
@@ -2154,18 +2265,25 @@ def log_entry(i)
...
-        'is not a string: ' + value.class.name
-      assert expected_labels.key?(key), "Unexpected label #{key} => #{value}"
-      assert_equal expected_labels[key], value, 'Value mismatch - expected ' \
-        "#{expected_labels[key]} in #{key} => #{value}"
+  # If check_exact_labels is true, assert 'labels' and 'expected_labels' match
+  # exactly. If check_exact_labels is false, assert 'labels' is a subset of
+  # 'expected_labels'.
+  def check_labels(expected_labels, labels, check_exact_labels = true)
+    return if expected_labels.empty? && labels.empty?
+    expected_labels.each do |expected_key, expected_value|
```
igorpeshansky: [Optional] Since the key is shared between the two hashes, you could rename `expected_key` to just `key`.

test/plugin/base_test.rb
```diff
@@ -2154,18 +2265,25 @@ def log_entry(i)
...
-      assert value.is_a?(String), "Value #{value} for label #{key} " \
-        'is not a string: ' + value.class.name
-      assert expected_labels.key?(key), "Unexpected label #{key} => #{value}"
-      assert_equal expected_labels[key], value, 'Value mismatch - expected ' \
-        "#{expected_labels[key]} in #{key} => #{value}"
+  # If check_exact_labels is true, assert 'labels' and 'expected_labels' match
+  # exactly. If check_exact_labels is false, assert 'labels' is a subset of
+  # 'expected_labels'.
+  def check_labels(expected_labels, labels, check_exact_labels = true)
+    return if expected_labels.empty? && labels.empty?
```
igorpeshansky: This check seems superfluous — if `expected_labels` is empty, the loop will be skipped anyway, and the last assert will still check the length of `labels`.

test/plugin/base_test.rb
```diff
@@ -2199,8 +2318,10 @@ def verify_json_log_entries(n, params, payload_type = 'textPayload')
             log_name
         end
         assert_equal params[:resource][:type], resource['type']
-        check_labels resource['labels'], params[:resource][:labels]
-        check_labels labels, params[:labels]
+
```
igorpeshansky: I'd put this blank line between the two `check_labels` calls instead, to group the resource checks together...

test/plugin/base_test.rb
```diff
@@ -2374,7 +2521,8 @@ def verify_cascading_json_detection_with_log_entry_fields(
         d.emit(input_log_entry)
         d.run
       end
-      verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
+      verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload',
+                         false) do |entry|
```
igorpeshansky: This still fits on the previous line...

test/plugin/base_test.rb
```diff
@@ -2427,7 +2575,11 @@ def verify_field_key(log_entry_field, default_key, custom_key,
         d.emit(input[:emitted_log])
         d.run
       end
-      verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
+      verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload', false) do |entry|
+        if log_entry_field == 'labels'
+          input[:expected_field_value] = COMPUTE_PARAMS[:labels].merge(
```
igorpeshansky: Can't you now do this unconditionally in `test_log_entry_labels_field`? If this is because of the hard-coded `nil` in the inputs, you could instead add an `expected_original_value` parameter that defaults to `nil`.

test/plugin/base_test.rb
```diff
@@ -2258,7 +2381,12 @@ def verify_subfields_from_record(payload_key)
       d.emit(payload_key => payload_value)
       d.run
     end
-    verify_log_entries(1, COMPUTE_PARAMS, destination_key) do |entry|
+    expected_params = COMPUTE_PARAMS.dup
+    if payload_key == DEFAULT_LABELS_KEY
+      payload_value = payload_value.merge(COMPUTE_PARAMS[:labels])
+      expected_params[:labels] = payload_value
```
igorpeshansky: Can you do this directly in `log_entry_subfields_params`?

test/plugin/base_test.rb
```diff
@@ -2292,9 +2420,22 @@ def verify_subfields_when_not_hash(payload_key)
       d.run
     end
     verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
-      field = get_fields(entry['jsonPayload'])[payload_key]
-      assert_equal 'a_string', get_string(field), entry
-      assert_false entry.key?(destination_key), entry
+      if payload_key == DEFAULT_LABELS_KEY
```
igorpeshansky: This seems like two separate methods in one: `verify_subfields_removed_when_not_hash` and `verify_subfields_untouched_when_not_hash`.

test/plugin/base_test.rb
```diff
@@ -2292,9 +2420,22 @@ def verify_subfields_when_not_hash(payload_key)
...
+        # The malformed field has been removed from the payload.
+        assert_true get_fields(entry['jsonPayload']).empty?, entry
+        # No additional labels.
+        assert_equal COMPUTE_PARAMS[:labels].size,
+                     entry[destination_key].size, entry
+      else
+        # Leave the malformed field as it is.
+        # TODO(qingling128) On the next major after 0.7.4, make the rest of
```
igorpeshansky: Would it be better to have this TODO in the implementation, rather than in the test?

test/plugin/base_test.rb
qingling128: This is to guard against `nil` values.

test/plugin/base_test.rb
qingling128: The actual keys in the labels might be different from expected keys. I would prefer being explicit.

test/plugin/base_test.rb
qingling128: Done

test/plugin/base_test.rb
qingling128: Done

test/plugin/base_test.rb
qingling128: Done

test/plugin/base_test.rb
qingling128: Done

test/plugin/base_test.rb
qingling128: Done

test/plugin/base_test.rb
qingling128: Done

test/plugin/base_test.rb
```diff
@@ -1249,41 +1251,70 @@ def test_log_entry_http_request_field_from_record
...
 
-  def test_log_entry_source_location_field_when_not_hash
-    verify_subfields_when_not_hash(DEFAULT_SOURCE_LOCATION_KEY)
+  def test_log_entry_labels_field_when_not_hash
+    verify_subfields_removed_when_not_hash(DEFAULT_LABELS_KEY)
   end
 
   def test_log_entry_operation_field_when_not_hash
-    verify_subfields_when_not_hash(DEFAULT_OPERATION_KEY)
+    # Leave the malformed field as it is.
```
igorpeshansky: This one-line comment probably belongs within `verify_subfields_untouched_when_not_hash`, similarly to `verify_subfields_removed_when_not_hash`.

test/plugin/base_test.rb
igorpeshansky: I meant in the actual implementation, in `out_google_cloud.rb`, probably in `set_log_entry_fields`.

test/plugin/base_test.rb
```diff
@@ -1360,30 +1395,54 @@ def test_http_request_skip_setting_latency
...
+                     custom_key_config: CONFIG_CUSTOM_INSERT_ID_KEY_SPECIFIED,
+                     sample_value: INSERT_ID)
+  end
+
+  def test_log_entry_labels_field
+    verify_field_key('labels',
+                     default_key: DEFAULT_LABELS_KEY,
+                     custom_key: 'custom_labels_key',
+                     custom_key_config: CONFIG_CUSTOM_LABELS_KEY_SPECIFIED,
+                     sample_value: COMPUTE_PARAMS[:labels].merge(
```
igorpeshansky: [Optional] I'd put the whole `COMPUTE_PARAMS[:labels].merge(LABELS_MESSAGE)` on the next line, with a backslash.

test/plugin/base_test.rb
```diff
@@ -2387,8 +2569,13 @@ def verify_cascading_json_detection_with_log_entry_fields(
     end
   end
 
-  def verify_field_key(log_entry_field, default_key, custom_key,
-                       custom_key_config, sample_value)
+  def verify_field_key(log_entry_field, test_params)
+    default_key = test_params[:default_key]
```
igorpeshansky: I'm not too crazy about this, but we can clean it up later.

test/plugin/base_test.rb
```diff
@@ -1249,41 +1251,70 @@ def test_log_entry_http_request_field_from_record
     verify_subfields_from_record(DEFAULT_HTTP_REQUEST_KEY)
   end
 
-  def test_log_entry_source_location_field_from_record
-    verify_subfields_from_record(DEFAULT_SOURCE_LOCATION_KEY)
+  def test_log_entry_labels_field_from_record
+    verify_subfields_from_record(DEFAULT_LABELS_KEY, false)
```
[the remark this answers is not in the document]
qingling128: Sync'ed offline. This is as expected.

test/plugin/base_test.rb
qingling128: Done

test/plugin/base_test.rb
qingling128: Sync'ed offline. Added TODO in both places.

test/plugin/base_test.rb
qingling128: I would prefer not using backslash unless it's absolutely necessary.

test/plugin/base_test.rb
qingling128: Cool

PR #293, opened 2019-02-14
Error out when attempting to construct a resource with Nil labels.

igorpeshansky:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1181,13 +1181,17 @@ def determine_agent_level_monitored_resource_labels(type)
 
       # GCE.
       when COMPUTE_CONSTANTS[:resource_type]
+        raise "Cannot construct a #{type} resource without vm_id and zone" \
```
qingling128: Weird. It should already error out:
https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/946798d787bef1ef2e38ec8ec327bdd4ac7e4bce/lib/fluent/plugin/out_google_cloud.rb#L1072

What is the use case? Something for `Platform::OTHER`?

igorpeshansky: Yes, exactly. When `use_metadata_service` is set to `false`, the platform is set to `Platform::OTHER`, and sending logs errors out with a hard-to-debug error. I'm just making the error more explicit.

PR #299, opened 2019-03-11
Fix each_codepoint undefined error when input is not a string.

qingling128:
[Warnings]

2019-03-11 20:49:54 +0000 [warn]: #0 [google_cloud] got unrecoverable error in primary and no secondary error_class=NoMethodError error="undefined method `each_codepoint' for #<Hash:0x00007f1bac785ce8>\nDid you mean?  each_cons
2019-03-11 20:49:54 +0000 [warn]: #0 [google_cloud] bad chunk is moved to /var/log/fluentd/backup/worker0/google_cloud/583d7977dc71f7fcd9a7f697a08b4483.log

[Code Snippet where the warn is thrown]
```
    def parse_json_or_nil(input)
      # Only here to please rubocop...
      return nil if input.nil?

      input.each_codepoint do |c|
        if c == 123
          # left curly bracket (U+007B)
          begin
            return JSON.parse(input)
          rescue JSON::ParserError
            return nil
          end
        else
          # Break (and return nil) unless the current character is whitespace,
          # in which case we continue to look for a left curly bracket.
          # Whitespace as per the JSON spec are: tabulation (U+0009),
          # line feed (U+000A), carriage return (U+000D), and space (U+0020).
          break unless c == 9 || c == 10 || c == 13 || c == 32
        end # case
      end # do
      nil
    end
```

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -2040,10 +2040,11 @@ module Fluent
                                json_payload
                              end
       elsif text_payload
+        text_payload = text_payload.to_s
         entry.text_payload = if @use_grpc
-                               convert_to_utf8(text_payload.to_s)
+                               convert_to_utf8(text_payload)
                              else
-                               text_payload.to_s
+                               text_payload
                              end
       end
     end
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -2003,6 +2003,8 @@ module Fluent
       ret
     end
 
+    # TODO(qingling128): Fix the inconsistent behavior of 'message', 'log' and
+    # 'msg' in the next major version 1.0.0.
     def set_payload(resource_type, record, entry, is_json)
       # Only one of {text_payload, json_payload} will be set.
       text_payload = nil
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -2039,9 +2039,9 @@ module Fluent
                              end
       elsif text_payload
         entry.text_payload = if @use_grpc
-                               convert_to_utf8(text_payload)
+                               convert_to_utf8(text_payload.to_s)
                              else
-                               text_payload
+                               text_payload.to_s
                              end
       end
     end
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -976,7 +976,6 @@ module Fluent
     end
 
     def parse_json_or_nil(input)
-      # Only here to please rubocop...
       return nil unless input.is_a?(String)
 
       input.each_codepoint do |c|
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -977,7 +977,7 @@ module Fluent
 
     def parse_json_or_nil(input)
       # Only here to please rubocop...
-      return nil if input.nil?
+      return input unless input.is_a?(String)
 
       input.each_codepoint do |c|
         if c == 123
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -977,7 +977,7 @@ module Fluent
 
     def parse_json_or_nil(input)
       # Only here to please rubocop...
-      return input unless input.is_a?(String)
+      return nil unless input.is_a?(String)
 
       input.each_codepoint do |c|
         if c == 123
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -536,54 +536,75 @@ module BaseTest
     end
   end
 
-  def test_structured_payload_json_log_detect_json_not_parsed_log_msg_hash
+  # TODO(qingling128): Fix the inconsistent behavior of 'message', 'log' and
+  # 'msg' in the next major version 1.0.0.
+  def test_structured_payload_json_log_detect_json_with_hash_input
     hash_value = {
       'msg' => 'test log entry 0',
       'tag2' => 'test',
       'data' => 5000,
       'some_null_field' => nil
     }
-    %w(log msg).each do |field|
+    [
+      {
+        config: APPLICATION_DEFAULT_CONFIG,
+        field_name: 'log',
+        expected_payload: 'jsonPayload'
+      },
+      {
+        config: APPLICATION_DEFAULT_CONFIG,
+        field_name: 'msg',
+        expected_payload: 'jsonPayload'
+      },
+      {
+        config: APPLICATION_DEFAULT_CONFIG,
+        field_name: 'message',
+        expected_payload: 'textPayload'
+      },
+      {
+        config: DETECT_JSON_CONFIG,
+        field_name: 'log',
+        expected_payload: 'jsonPayload'
+      },
+      {
+        config: DETECT_JSON_CONFIG,
+        field_name: 'msg',
+        expected_payload: 'jsonPayload'
+      },
+      {
+        config: DETECT_JSON_CONFIG,
+        field_name: 'message',
+        expected_payload: 'textPayload'
+      }
+    ].each do |test_params|
       new_stub_context do
         setup_gce_metadata_stubs
         setup_logging_stubs do
-          d = create_driver(DETECT_JSON_CONFIG)
-          d.emit(field => hash_value)
+          d = create_driver(test_params[:config])
+          d.emit(test_params[:field_name] => hash_value)
           d.run
         end
-        verify_log_entries(1, COMPUTE_PARAMS, 'jsonPayload') do |entry|
-          json_payload = get_fields(entry['jsonPayload'])
-          assert_equal 1, json_payload.size, entry
-          fields = get_fields(get_struct(json_payload[field]))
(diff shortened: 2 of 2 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -977,7 +977,7 @@ def write_request_via_rest(entries:,
 
     def parse_json_or_nil(input)
       # Only here to please rubocop...
-      return nil if input.nil?
+      return input unless input.is_a?(String)
```
igorpeshansky: Why not `return nil unless input.is_a?(String)`?

qingling128: I was debating between whether to collapse the Hash as we do if we managed to parse a hash from the JSON.

Sync'ed offline. We decided to leave the user input as it is.

Changed.

igorpeshansky: Note that `input.is_a?(String)` being false did not necessarily mean that `input` is a Hash, so leaving it alone was the right choice.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -977,7 +977,7 @@ def write_request_via_rest(entries:,
 
     def parse_json_or_nil(input)
       # Only here to please rubocop...
```
igorpeshansky: This comment is no longer true, right?

test/plugin/base_test.rb
```diff
@@ -536,6 +536,35 @@ def test_structured_payload_json_log_detect_json_not_parsed_json
...
   end
 
+  def test_structured_payload_json_log_detect_json_not_parsed_hash
+    hash_value = {
+      'msg' => 'test log entry 0',
+      'tag2' => 'test',
+      'data' => 5000,
+      'some_null_field' => nil
+    }
+    %w(log msg).each do |field|
```
igorpeshansky: Why not also `message`?

lib/fluent/plugin/out_google_cloud.rb

test/plugin/base_test.rb
qingling128: `message` behaves differently in general (which is why multiple other tests do `log msg` only as well: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/367427e4cb0b619e27b9c5b77f7080eeff4cb4eb/lib/fluent/plugin/out_google_cloud.rb#L2034.

Feels weird, but we couldn't easily change it (maybe the nest major release).

igorpeshansky: Right, but the JSON detection code runs on all three: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/367427e4cb0b619e27b9c5b77f7080eeff4cb4eb/lib/fluent/plugin/out_google_cloud.rb#L640.
What happens if we pass a Hash in the `message` field?

igorpeshansky: Right, but the JSON detection code runs on all three: https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/367427e4cb0b619e27b9c5b77f7080eeff4cb4eb/lib/fluent/plugin/out_google_cloud.rb#L640.
What happens if we pass a Hash in the `message` field?

qingling128: Added a test and a TODO to fix the inconsistency.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -2040,9 +2039,9 @@ def set_payload(resource_type, record, entry, is_json)
                              end
       elsif text_payload
         entry.text_payload = if @use_grpc
-                               convert_to_utf8(text_payload)
+                               convert_to_utf8(text_payload.to_s)
```
qingling128: This is to fix the following error:

```
/usr/local/google/home/lingshi/repo/fluent-plugin-google-cloud/test/plugin/base_test.rb:578:in `test_structured_payload_json_log_detect_json_not_parsed_message_hash'
test/plugin/test_out_google_cloud_grpc.rb:411:in `setup_logging_stubs'
     578:     setup_logging_stubs do
     579:       d = create_driver(DETECT_JSON_CONFIG)
     580:       d.emit('message' => hash_value)
  => 581:       d.run
     582:     end
     583:     verify_log_entries(1, COMPUTE_PARAMS, 'textPayload') do |entry|
     584:       text_payload = entry['textPayload']
/usr/local/google/home/lingshi/repo/fluent-plugin-google-cloud/test/plugin/base_test.rb:581:in `block in test_structured_payload_json_log_detect_json_not_parsed_message_hash'
/usr/local/google/home/lingshi/.rvm/gems/ruby-2.4.3/gems/fluentd-1.2.5/lib/fluent/test/output_test.rb:76:in `run'
/usr/local/google/home/lingshi/.rvm/gems/ruby-2.4.3/gems/fluentd-1.2.5/lib/fluent/test/input_test.rb:123:in `run'
/usr/local/google/home/lingshi/.rvm/gems/ruby-2.4.3/gems/fluentd-1.2.5/lib/fluent/test/base.rb:71:in `run'
/usr/local/google/home/lingshi/.rvm/gems/ruby-2.4.3/gems/fluentd-1.2.5/lib/fluent/test/input_test.rb:124:in `block in run'
/usr/local/google/home/lingshi/.rvm/gems/ruby-2.4.3/gems/fluentd-1.2.5/lib/fluent/test/output_test.rb:94:in `block in run'
/usr/local/google/home/lingshi/.rvm/gems/ruby-2.4.3/gems/fluentd-1.2.5/lib/fluent/compat/output.rb:131:in `write'
/usr/local/google/home/lingshi/repo/fluent-plugin-google-cloud/lib/fluent/plugin/out_google_cloud.rb:607:in `write'
/usr/local/google/home/lingshi/repo/fluent-plugin-google-cloud/lib/fluent/plugin/out_google_cloud.rb:607:in `each'
/usr/local/google/home/lingshi/repo/fluent-plugin-google-cloud/lib/fluent/plugin/out_google_cloud.rb:613:in `block in write'
/usr/local/google/home/lingshi/repo/fluent-plugin-google-cloud/lib/fluent/plugin/out_google_cloud.rb:613:in `each'
/usr/local/google/home/lingshi/repo/fluent-plugin-google-cloud/lib/fluent/plugin/out_google_cloud.rb:686:in `block (2 levels) in write'
/usr/local/google/home/lingshi/repo/fluent-plugin-google-cloud/lib/fluent/plugin/out_google_cloud.rb:2042:in `set_payload'
/usr/local/google/home/lingshi/repo/fluent-plugin-google-cloud/lib/fluent/plugin/out_google_cloud.rb:2124:in `convert_to_utf8'
Error: test_structured_payload_json_log_detect_json_not_parsed_message_hash(GoogleCloudOutputGRPCTest): NoMethodError: undefined method `encode' for #<Hash:0x000055c57c374b18>
```

igorpeshansky: Yeah, I was wondering what would happen if someone sent a `message` field that was not a String...

Would it make sense to change it above, e.g.:
```ruby
      elsif text_payload
        text_payload = text_payload.to_s
        entry.text_payload = if @use_grpc
                               convert_to_utf8(text_payload)
                             else
                               text_payload
                             end
```
?

test/plugin/base_test.rb
```diff
@@ -536,6 +536,57 @@ def test_structured_payload_json_log_detect_json_not_parsed_json
...
+          assert_equal 'test log entry 0', get_string(fields['msg']), entry
+          assert_equal 'test', get_string(fields['tag2']), entry
+          assert_equal 5000, get_number(fields['data']), entry
+          assert_equal null_value, fields['some_null_field'], entry
+        end
+      end
+    end
+  end
+
+  # TODO(qingling128): Fix the inconsistent behavior of 'message', 'log' and
```
igorpeshansky: Should this TODO be in `out_google_cloud.rb` as well (or even instead)?

test/plugin/base_test.rb
```diff
@@ -536,6 +536,57 @@ def test_structured_payload_json_log_detect_json_not_parsed_json
     end
   end
 
+  def test_structured_payload_json_log_detect_json_not_parsed_log_msg_hash
```
igorpeshansky: Do we want to also test that this happens when `detect_json` is false (i.e., the field is left alone)?

test/plugin/base_test.rb
```diff
@@ -536,6 +536,57 @@ def test_structured_payload_json_log_detect_json_not_parsed_json
...
+          assert_equal 5000, get_number(fields['data']), entry
+          assert_equal null_value, fields['some_null_field'], entry
+        end
+      end
+    end
+  end
+
+  # TODO(qingling128): Fix the inconsistent behavior of 'message', 'log' and
+  # 'msg' in the next major version 1.0.0.
+  def test_structured_payload_json_log_detect_json_not_parsed_message_hash
```
igorpeshansky: Do we want to also test that this happens when `detect_json` is false (i.e., the field is serialized to a string)?

lib/fluent/plugin/out_google_cloud.rb
qingling128: Done

test/plugin/base_test.rb

test/plugin/base_test.rb
qingling128: Added

test/plugin/base_test.rb
qingling128: Combined the tests

PR #314, opened 2019-05-20
Bump version to 0.8.0.

qingling128:
Since we've upgraded a Ruby version.

### DIFF
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -10,7 +10,7 @@ eos
   gem.homepage      =
     'https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud'
   gem.license       = 'Apache-2.0'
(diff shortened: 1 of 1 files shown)


igorpeshansky:
Does bumping the embedded Ruby version make this backward-incompatible?

qingling128:
The Ruby is embedded, so it should still be considered backward compatible I think. Minor version is more for new features. Hmm... Does Ruby 2.5 enable new features?

igorpeshansky:
Even if Ruby 2.5 enables new features, it should be backward-compatible...

qingling128:
Yeah, `0.7.0` => `0.8.0` is still for backward compatible changes. Only `0.7.0` => `1.0.0` is for non-compatible change. The difference between patch version and minor version bumps is whether new functionality is enabled (https://semver.org/#summary). Ruby verison bump feels like a grey area.

PR #316, opened 2019-05-29
Automated: Bump gem version to 0.7.13 for future releases.

stackdriver-instrumentation-release:


qingling128:
Manually set CLA tag to yes as we wait for whitelisting this robot account.

PR #318, opened 2019-06-20
Add debugging info for non-hash records.

qingling128:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -1279,7 +1279,7 @@ def group_log_entries_by_tag_and_local_resource_id(chunk)
       chunk.msgpack_each do |tag, time, record|
         unless record.is_a?(Hash)
           @log.warn 'Dropping log entries with malformed record: ' \
-                    "'#{record.inspect}'. " \
+                    "'#{record.inspect}' from tag '#{tag}'. " \
```
igorpeshansky: Worth recording the timestamp as well?

qingling128: The more the better. Added.

PR #319, opened 2019-06-23
Mute "warning: already initialized constant" message in the log.

qingling128:
This message is not harmful, just a bit misleading when debugging other issues.

```
/opt/google-fluentd/embedded/lib/ruby/gems/2.4.0/gems/fluent-plugin-google-cloud-0.7.13/lib/fluent/plugin/out_google_cloud.rb:47: warning: already initialized constant Google::Auth::CredentialsLoader::GCLOUD_CONFIG_COMMAND
/opt/google-fluentd/embedded/lib/ruby/gems/2.4.0/gems/googleauth-0.8.1/lib/googleauth/credentials_loader.rb:52: warning: previous definition of GCLOUD_CONFIG_COMMAND was here
```

PR #310, opened 2019-05-14
Bump googleauth to 0.8.1.

igorpeshansky:
Also bump the fluentd dependency version to 1.4.2.

PR #322, opened 2019-06-24
Completely disable gcloud credential lookups rather than silence them.

igorpeshansky:
This is a follow-on to #310 (and #319).

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -48,6 +48,11 @@ module Google
       warn_level = $VERBOSE
       begin
         $VERBOSE = nil
+        # These constants are used to invoke gcloud on Linux and Windows,
+        # respectively. Ideally, we would have overridden
+        # CredentialsLoader.load_gcloud_project_id, but we cannot catch it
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -42,16 +42,15 @@ class Map
...
-      # Set $VERBOSE to nil to mute the following warning:
-      # "warning: already initialized constant
-      # Google::Auth::CredentialsLoader::GCLOUD_CONFIG_COMMAND".
+      # Set $VERBOSE to nil to mute the "already initialized constant" warnings.
       warn_level = $VERBOSE
       begin
         $VERBOSE = nil
-        GCLOUD_CONFIG_COMMAND =
-          'config config-helper --format json --verbosity none'.freeze
+        GCLOUD_POSIX_COMMAND = '/bin/true'.freeze
```
qingling128: Can we add a comment and maybe a link to https://github.com/googleapis/google-auth-library-ruby/blob/f2f83b20bc1f32db0c90cf5cea0922d0558a35b0/lib/googleauth/credentials_loader.rb#L174 to clarify what we are doing?

PR #325, opened 2019-07-09
Fix crash on unknown monitoring_type.

davidbtucker:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -511,6 +511,10 @@ def configure(conf)
             ' Stackdriver output plugin due to a transient error and were'\
             ' retried')
         @ok_code = @use_grpc ? GRPC::Core::StatusCodes::OK : 200
+        unless @successful_requests_count
```
igorpeshansky: You should be able to ask for the registry `name` and complain if it's `nil`... Otherwise you'll get more errors when trying to update the counters (e.g., `NoMethodError`).

lib/fluent/plugin/monitoring.rb
```diff
@@ -16,7 +16,7 @@ module Monitoring
   # Base class for the monitoring registry.
   class BaseMonitoringRegistry
     def counter(_name, _desc)
-      _undefined
```
igorpeshansky: This was intentional, I think, to prevent you from creating a counter for an undefined registry. See my suggestion for an alternative approach.

test/plugin/test_out_google_cloud.rb
```diff
@@ -346,6 +346,14 @@ def test_statusz_endpoint
     end
   end
 
+  def test_monitoring_config
+    setup_gce_metadata_stubs
+    create_driver(%(
```
igorpeshansky: What is this testing? That the code doesn't crash? Can you please add a comment to that effect?

test/plugin/test_out_google_cloud.rb
```diff
@@ -346,6 +346,14 @@ def test_statusz_endpoint
...
+      enable_monitoring  true
+      monitoring_type    not_prometheus
```
igorpeshansky: For consistency with other tests, this config string constant should be pulled out into `Constants` and named.

test/plugin/constants.rb
```diff
@@ -464,6 +464,12 @@ module Constants
     #{CONFIG_LABEL_MAP_CONFLICTING}
   ).freeze
 
+  # For monitoring config.
+  CONFIG_MONITORING_TYPE_NOT_PROMETHEUS = %(
```
igorpeshansky: How about `CONFIG_UNKNOWN_MONITORING_TYPE`?

test/plugin/test_out_google_cloud.rb
```diff
@@ -346,6 +346,13 @@ def test_statusz_endpoint
     end
   end
 
+  def test_monitoring_config
```
igorpeshansky: How about `test_monitoring_config_ignores_non_prometheus`? Or `test_monitoring_config_requires_prometheus_for_metrics`? Or even `test_monitoring_config_graceful_failure`?

igorpeshansky: Why are you only adding this on the non-gRPC path? Shouldn't this be in `base_test.rb` instead?

test/plugin/test_out_google_cloud.rb
```diff
@@ -346,6 +346,13 @@ def test_statusz_endpoint
...
+    # Verify that driver creation succeeds when monitoring type is not
+    # "prometheus" (in which case, we simply don't record metrics).
```
igorpeshansky: You could assert that no prometheus counters have been defined (i.e., that they're `nil`). See `assert_prometheus_metric_value` in `base_test.rb` for implementation ideas.

test/plugin/test_out_google_cloud.rb
igorpeshansky: Forgot to respond to this one. It's not immediately obvious from the content of the config what is being tested. Giving the constant a meaningful name makes it easier to understand what it is about this config that makes it relevant to the test.

test/plugin/constants.rb
davidbtucker: Done, thanks.

test/plugin/test_out_google_cloud.rb
davidbtucker: How's test_configure_ignores_unknown_monitoring_type?

davidbtucker: Ah, because I didn't understand how these tests are organized. I assumed changes to out_google_cloud.rb were tested in test_out_google_cloub.rb.

Moved to base_test.rb.

test/plugin/test_out_google_cloud.rb
davidbtucker: Done.

test/plugin/test_out_google_cloud.rb
davidbtucker: I don't mind pulling it out into a constant. But I think it would be clearer if the input and expected result were next to each other, not in separate files.

test/plugin/base_test.rb
```diff
@@ -142,6 +142,24 @@ def test_metadata_agent_url_customization
     end
   end
 
+  def test_configure_ignores_unknown_monitoring_type
+    # Verify that driver creation succeeds when monitoring type is not
+    # "prometheus" (in which case, we simply don't record metrics),
+    # and that the counters are set to nil.
+    setup_gce_metadata_stubs
+    create_driver(CONFIG_UNKNOWN_MONITORING_TYPE)
+    assert_nil(Prometheus::Client.registry.get(
```
igorpeshansky: I'm inclined to suggest factoring this out into an `assert_no_prometheus_metric` method, to abstract away the registry lookup, but I'd like @qingling128 to sanity-check this request.
If you do factor it out, this test would read as follows:
```ruby
    setup_gce_metadata_stubs
￼    create_driver(CONFIG_UNKNOWN_MONITORING_TYPE)
    assert_no_prometheus_metric(:stackdriver_successful_requests_count)
    assert_no_prometheus_metric(:stackdriver_failed_requests_count)
    assert_no_prometheus_metric(:stackdriver_ingested_entries_count)
    assert_no_prometheus_metric(:stackdriver_dropped_entries_count)
    assert_no_prometheus_metric(:stackdriver_retried_entries_count)
```

test/plugin/test_out_google_cloud.rb
igorpeshansky: This is a good point. Having that constant in a separate file is consistent with other tests, but we could revisit this decision and possibly pull some constants into test local variables. Let's do that in a separate PR, though, and discuss the pros and cons.

test/plugin/test_out_google_cloud.rb
igorpeshansky: SGTM.

test/plugin/base_test.rb
qingling128: Either way works for me. I like how clean it looks this way. Yet it's not a lot of duplicated code.

igorpeshansky: I like to keep all direct accesses to dependent libraries in one place, rather than sprinkled through the tests, but won't insist if you are ok with leaving it as-is.

PR #326, opened 2019-07-09
Use ERB template for statusz page, and add parsed configuration.

davidbtucker:


lib/fluent/plugin/statusz.rb
```diff
@@ -12,20 +12,114 @@
...
+      <div class="lefthand">
+        Started: <%= SERVER_START %><br>
+        Up <%= uptime_str %><br>
+      </div>
+    </div>
+
+    <h1>Parsed configuration</h1>
+
+    <table>
+      <tr><th>adjust_invalid_timestamps</th><td><%= plugin.adjust_invalid_timestamps %></td></tr>
```
igorpeshansky: This seems like something you can generate programmatically... Making it a hand-typed constant is too error-prone...

test/plugin/test_out_google_cloud.rb
```diff
@@ -339,10 +339,125 @@ def test_statusz_endpoint
     WebMock.disable_net_connect!(allow_localhost: true)
     # TODO(davidbtucker): Consider searching for an unused port
     # instead of hardcoding a constant here.
-    d = create_driver('statusz_port 5678')
+    d = create_driver(%(
+      statusz_port                  5678
```
igorpeshansky: Shouldn't this config live in `Constants`?

test/plugin/test_out_google_cloud.rb
```diff
@@ -339,10 +339,125 @@ def test_statusz_endpoint
...
+      zone                          asia-east2
+    ))
     d.run do
-      assert_match Regexp.new('.*<h1>Status</h1>.*'),
-                   Net::HTTP.get('127.0.0.1', '/statusz', 5678)
+      resp = Net::HTTP.get('127.0.0.1', '/statusz', 5678)
+      assert_match Regexp.new('<h1>Status for .*</h1>.*'), resp
+
+      # rubocop:disable LineLength
+      assert_match Regexp.new('<tr><th>adjust_invalid_timestamps</th>' \
```
igorpeshansky: This definitely looks like something you can match in a loop...

davidbtucker: Done.

test/plugin/test_out_google_cloud.rb
davidbtucker: Done.

lib/fluent/plugin/statusz.rb
davidbtucker: I looked into it, but I didn't see any better way.

test/plugin/test_out_google_cloud.rb
igorpeshansky: You could factor the HTML elements out as well, e.g.:
```ruby
      config_values = {
        adjust_invalid_timestamps: 'false',
        autoformat_stackdriver_trace: 'false',
        coerce_to_utf8: 'false',
        detect_json: 'true',
        detect_subservice: 'false',
        enable_metadata_agent: 'true',
        enable_monitoring: 'true',
        http_request_key: 'test_http_request_key',
        insert_id_key: 'test_insert_id_key',
        k8s_cluster_location: 'test-k8s-cluster-location',
        k8s_cluster_name: 'test-k8s-cluster-name',
        kubernetes_tag_regexp: '.*test-regexp.*',
        label_map: '{"label_map_key"=>"label_map_value"}',
        labels_key: 'test_labels_key',
        labels: '{"labels_key"=>"labels_value"}',
        logging_api_url: 'http://localhost:52000',
        metadata_agent_url: 'http://localhost:12345',
        monitoring_type: 'not_prometheus',
        non_utf8_replacement_string: 'zzz',
        operation_key: 'test_operation_key',
        partial_success: 'false',
        project_id: 'test-project-id-123',
        require_valid_tags: 'true',
        source_location_key: 'test_source_location_key',
        span_id_key: 'test_span_id_key',
        split_logs_by_tag: 'true',
        statusz_port: '5678',
        subservice_name: 'test_subservice_name',
        trace_key: 'test_trace_key',
        trace_sampled_key: 'test_trace_sampled_key',
        use_aws_availability_zone: 'false',
        use_grpc: 'true',
        use_metadata_service: 'false',
        vm_id: '12345',
        vm_name: 'test.hostname.org',
        zone: 'asia-east2'
      }

      must_match = [
        '<h1>Status for .*</h1>.*',
      ] + config_values.map {|k, v| "<tr><th>#{k}</th><td>#{v}</td></tr>"} + [
        '^</html>$'
      ]
```
Note that you don't need to escape the double quotes within the single quotes.

You could also just loop over the config values, and hard-code the prefix/suffix:
```ruby
      resp = Net::HTTP.get('127.0.0.1', '/statusz', 5678)
      assert_match Regexp.new('<h1>Status for .*</h1>.*'), resp

      config_values = ...

      config_values.each do |k, v|
        assert_match Regexp.new("<tr><th>#{k}</th><td>#{v}</td></tr>"), resp
      end

      assert_match Regexp.new('^</html>$'), resp
```

lib/fluent/plugin/statusz.rb
```diff
@@ -12,20 +12,114 @@
...
-    # TODO(davidbtucker): Add more status information here.
-    [
-      '<html>',
-      '<body>',
-      '<h1>Status</h1>',
-      "<b>Command-line flags:</b> #{CGI.escapeHTML(ARGV.join(' '))}",
-      '</body>',
-      '</html>'
-    ].join("\n") + "\n"
+  def response(plugin)
```
igorpeshansky: Where is `plugin` used?

lib/fluent/plugin/statusz.rb
davidbtucker: Ah nice, thanks!

lib/fluent/plugin/statusz.rb
davidbtucker: In the template.

lib/fluent/plugin/statusz.rb
davidbtucker: Done, thanks!

test/plugin/test_out_google_cloud.rb
davidbtucker: I think this one I'll leave as-is. It's a bit clearer to have the raw output values rather than constructing them programmatically.

lib/fluent/plugin/statusz.rb
igorpeshansky: Can you please add a comment before `ERB.new(STATUSZ_TMPL).result(binding)` to that effect?

test/plugin/test_out_google_cloud.rb
igorpeshansky: I'd like @qingling128 to chime in here. I'm also not too crazy about the list concatenation in the first snippet, but I do believe that the second proposal (match prefix, loop over config values, then match suffix) shows the intent of the test better. The raw HTML match values seem too error-prone to me, but I'll defer to @qingling128's decision.

qingling128: I like the raw output value comparison better, because it helps catch subtle errors (e.g. order changed, additional stuff got in by accident). It's also more straightforward to get a sense of what the output is like.

igorpeshansky: Isn't the intent of this test to make sure that the correct configuration values are present in the output? The raw HTML is secondary, isn't it? We could even change the regexps to `'\badjust_invalid_timestamps\b.*\bfalse\b'` to underscore that. IMO, it's the raw HTML tags that make this really fragile. If we remove the raw HTML from the matches, I'd be ok with keeping the literal regexps in the array.

davidbtucker: Yeah, that works for me. Thanks.

test/plugin/test_out_google_cloud.rb
```diff
@@ -339,10 +339,54 @@ def test_statusz_endpoint
...
+        '\bdetect_json\b.*\btrue\b',
+        '\bdetect_subservice\b.*\bfalse\b',
+        '\benable_metadata_agent\b.*\btrue\b',
+        '\benable_monitoring\b.*\btrue\b',
+        '\bhttp_request_key\b.*\btest_http_request_key\b',
+        '\binsert_id_key\b.*\btest_insert_id_key\b',
+        '\bk8s_cluster_location\b.*\btest-k8s-cluster-location\b',
+        '\bk8s_cluster_name\b.*\btest-k8s-cluster-name\b',
+        '\bkubernetes_tag_regexp\b.*\b.*test-regexp.*\b',
+        '\blabel_map\b.*{\"label_map_key\"=>\"label_map_value\"}',
```
igorpeshansky: You don't really need to escape the double quotes within the single quotes.

lib/fluent/plugin/statusz.rb
igorpeshansky: Just floating this up for attention.

davidbtucker: Oops, done.

test/plugin/test_out_google_cloud.rb
davidbtucker: Done.

PR #329, opened 2019-07-31
Skip proxy for local metadata requests

### LINKED ISSUE #101: Attempts to use proxy even for metadata service
I'm using google-fluentd on a GCE instance with no external IP, and running a proxy (Squid) on another instance that has an external IP.

I added:

```
export http_proxy="http://gateway1:3128/"
export no_proxy=169.254.169.254
```

to /etc/default/google-fluentd, but when fetching metadata fluentd uses the open-uri library which does not support no_proxy. Therefore the metadata requests get sent to the proxy, which fails.

As a workaround, setting HTTP_PROXY instead of http_proxy works because open-uri only uses the latter, while httpclient (used to connect to the logging API) uses both.

A more reliable solution would be to disable the proxy for the metadata requests. For example, this change worked for me:

```
--- /home/james/out_google_cloud.rb	2016-12-14 19:42:51.026886681 +0000
+++ /home/james/out_google_cloud.rb.patched	2016-12-15 14:59:28.954799737 +0000
@@ -514,7 +514,7 @@
       end
 
       begin
-        open('http://' + METADATA_SERVICE_ADDR) do |f|
+        open('http://' + METADATA_SERVICE_ADDR, :proxy => false) do |f|
           if f.meta['metadata-flavor'] == 'Google'
             @log.info 'Detected GCE platform'
             return Platform::GCE
@@ -537,7 +537,7 @@
         @platform == Platform::GCE
       # See https://cloud.google.com/compute/docs/metadata
       open('http://' + METADATA_SERVICE_ADDR + '/computeMetadata/v1/' +
-           metadata_path, 'Metadata-Flavor' => 'Google', &:read)
+           metadata_path, 'Metadata-Flavor' => 'Google', :proxy => false, &:read)
     end
 
     def fetch_ec2_metadata
```

This doesn't solve the problem alone because googleauth also accesses the metadata service to get the default credentials, and has the same issue.

jkohen:
We know the metadata server is local, so we don't need a proxy. Today, if a user sets up a proxy with the intention of using it for connections to the Logging service, they'll break the metadata requests.

Fixes #101

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1036,7 +1036,7 @@ module Fluent
       end
 
       begin
-        open('http://' + METADATA_SERVICE_ADDR, :proxy => false) do |f|
+        open('http://' + METADATA_SERVICE_ADDR, proxy: false) do |f|
           if f.meta['metadata-flavor'] == 'Google'
             @log.info 'Detected GCE platform'
             return Platform::GCE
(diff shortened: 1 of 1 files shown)


jkohen:
> This is good, but will it also affect application default credentials? If not, the auth library will still talk to the server via the proxy, defeating the purpose.

According to #101 two users reported that this change is sufficient. Are you talking about a situation that they may not have exercised?

PR #330, opened 2019-08-09
Upgrade Ruby to 2.6, update grpc to 1.18.0, and ugrade webmock to 3.6.2.

qingling128:


fluent-plugin-google-cloud.gemspec
```diff
@@ -25,13 +25,13 @@ eos
   gem.add_runtime_dependency 'google-api-client', '0.28.4'
   gem.add_runtime_dependency 'google-cloud-logging', '1.6.0'
   gem.add_runtime_dependency 'google-protobuf', '3.9.0'
-  gem.add_runtime_dependency 'grpc', '1.14.2'
+  gem.add_runtime_dependency 'grpc', '1.18.0'
```
jkohen: Can we go straight to the latest stable version, 1.22.0 from https://rubygems.org/gems/grpc ?

It seems to be compatible with all the Ruby version we support: https://github.com/grpc/grpc/blob/v1.22.x/grpc.gemspec#L16

qingling128: This one seems fine. We can use latest for this gem. Updated.

I originally upgraded everything to latest, then tests started to break apart. So I reverted it to minimum version upgrades in case any dependencies introduced backward incompatible changes or extra resource usage.

PR #332, opened 2019-08-09
Upgrade and pin development dependencies and fix instance variable initialization.

qingling128:


test/plugin/test_out_google_cloud_grpc.rb
```diff
@@ -416,7 +416,7 @@ def write_log_entries(entries,
         raise @error
       rescue
         # Google::Gax::GaxError will wrap the latest thrown exception as @cause.
-        raise Google::Gax::GaxError, @message
```
qingling128: What happens is here is that we are using an instance variable `@message`, which was never initialized or assigned value to. So basically we were always using `nil` value here. As to the actual test case, this message does not matter because we are in fact verifying the `@cause` inside the `GaxError` class which contains the actual error that we care about.

PR #285, opened 2019-01-25
Add more time keys to the preserved keys for json detection

yanweiguo:
Fixes #211. The newly added keys are from considered in function `compute_timestamp`.

PR #333, opened 2019-08-13
Add more time keys to the preserved keys for JSON detection.

qingling128:


qingling128:
This is a follow up of https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/285.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -640,6 +640,10 @@ def write(chunk)
             # field.
             preserved_keys = [
```
jkohen: Can you add a reminder to keep this list up to date with the user-facing documentation at https://cloud.google.com/logging/docs/agent/configuration#special-fields ?

qingling128: Sure thing. Done.

PR #342, opened 2019-10-07
Upgraded to Prometheus client 0.10.0.

jkohen:
The client library had a breaking change to its API. I decided to fix this as pre-work for making the monitoring code also support OpenCensus, so we can use the new API as the common API for the abstraction on top of both.

### DIFF
--- a/test/plugin/test_out_google_cloud.rb
+++ b/test/plugin/test_out_google_cloud.rb
@@ -88,15 +88,11 @@ class GoogleCloudOutputTest < Test::Unit::TestCase
     assert_prometheus_metric_value(
       :stackdriver_successful_requests_count, 1, grpc: false, code: 200)
     assert_prometheus_metric_value(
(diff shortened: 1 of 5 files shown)


test/plugin/test_out_google_cloud.rb
```diff
@@ -87,16 +87,12 @@ def test_partial_success
     d.run
     assert_prometheus_metric_value(
       :stackdriver_successful_requests_count, 1, grpc: false, code: 200)
-    assert_prometheus_metric_value(
-      :stackdriver_failed_requests_count, 0, grpc: false)
```
yinzhi7: I saw you change the syntax of codes in out_google_cloud.rb. Have you added unit tests for the change? Or test it somewhere making sure it doesn't break anything.

jkohen: These are the actual unit tests. The changes are required for the unit tests to run, because the library changed its API.

PR #343, opened 2019-10-07
Remove waits on test start-up.

jkohen:
This brings down text execution from >5 minutes to <20 seconds. The tests all pass, hopefully they don't become flaky. Upstream is unclear about why the waits were needed, I'll follow up with them: https://github.com/fluent/fluentd/blob/master/lib/fluent/test/base.rb#L64-L75

Upstream discussion (just posted): https://groups.google.com/d/topic/fluentd/ze7YY1ue40s/discussion

### DIFF
--- a/test/plugin/test_driver.rb
+++ b/test/plugin/test_driver.rb
@@ -18,6 +18,17 @@ require 'fluent/test/input_test'
 
 module Fluent
   module Test
+    class BufferedOutputTestDriver < InputTestDriver
+      @@run_method = BufferedOutputTestDriver.instance_method(:run)
+      def run(num_waits = 0)
+        return @@run_method.bind(self).call(num_waits)
+      end
+    end
+  end
+end
+
+module Fluent
+  module Test
     # Similar to the standard BufferedOutputTestDriver, but allows multiple tags
     # to exist in one chunk.
     class MultiTagBufferedOutputTestDriver < InputTestDriver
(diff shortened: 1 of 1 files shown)


jkohen:
@qingling128 @igorpeshansky I was unable to get the code to work with class instance vars as rubocop recommended. Any thoughts? I got this particular sample from here: https://github.com/ruby/ruby/blob/master/lib/uri/rfc3986_parser.rb#L92-L95

test/plugin/test_driver.rb
```diff
@@ -16,6 +16,19 @@
 require 'fluent/event'
 require 'fluent/test/input_test'
 
+module Fluent
+  module Test
+    # rubocop:disable Style/ClassVars
+    class BufferedOutputTestDriver < InputTestDriver
+      @@run_method = BufferedOutputTestDriver.instance_method(:run)
```
qingling128: `@run_method` is considered class instance variable, while `@@run_method` is considered class variable.
eg.: http://www.railstips.org/blog/archives/2006/11/18/class-and-instance-variables-in-ruby/

For our case, we do want to use a class variable in order to call the original method (the method belongs to the class) later on, disabling the rubocop complaints sounds like a reasonable approach to me.

jkohen: Thanks for confirming that for me!

PR #344, opened 2019-10-10
Backported to prometheus client 0.9.0

jkohen:
I didn't simply revert #342 because I want to preserve the changes in out_google_cloud for my my follow-up change to support OpenCensus. It will also be much easier to support prometheus client 0.10.0 on top of this change:

- Add the `labels:`, `docstring:`, and `by:` prefixes to the `counter.increment` and `registry.counter` calls in monitoring.rb.
- Add the `labels:` prefix to the assert_prometheus metric method in base_test.rb.

### DIFF
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -29,7 +29,9 @@ eos
   gem.add_runtime_dependency 'json', '2.2.0'
 
   gem.add_development_dependency 'mocha', '1.9.0'
-  gem.add_development_dependency 'prometheus-client', '0.10.0'
+  # Keep this the same as in
+  # https://github.com/fluent/fluent-plugin-prometheus/blob/master/fluent-plugin-prometheus.gemspec
+  gem.add_development_dependency 'prometheus-client', '< 0.10'
   # TODO(qingling128): Upgrade rake to 11.0+ after the following issues are
   # fixed because rake (11.0+) requires ALL variables to be explicitly
   # initialized.
--- a/lib/fluent/plugin/monitoring.rb
+++ b/lib/fluent/plugin/monitoring.rb
@@ -13,9 +13,27 @@
 # limitations under the License.
 
 module Monitoring
+  # Base class for the counter.
+  class BaseCounter
+    def increment(_labels, _count)
+      nil
+    end
+  end
+
+  # Prometheus implementation of counters.
+  class PrometheusCounter < BaseCounter
+    def initialize(prometheus_counter)
+      @counter = prometheus_counter
+    end
+
+    def increment(by: 1, labels: {})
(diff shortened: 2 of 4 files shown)


fluent-plugin-google-cloud.gemspec
```diff
@@ -29,7 +29,9 @@ eos
   gem.add_runtime_dependency 'json', '2.2.0'
 
   gem.add_development_dependency 'mocha', '1.9.0'
-  gem.add_development_dependency 'prometheus-client', '0.10.0'
+  # Keep this the same as in
+  # https://github.com/fluent/fluent-plugin-prometheus/blob/master/fluent-plugin-prometheus.gemspec
```
qingling128: Can we add a warning here to mention the issue with `prometheus-client` 0.10 so that we don't upgrade by accident? It would be even better if it mentions a ticket / bug to monitor the root cause fix so we know when it's safe to upgrade.

jkohen: Re. comment, I think the goal here clear: keep the import the same, to avoid runtime conflicts. If anyone is feeling curious about how we learned that, they could use git blame to find this PR.

If you feel strongly about that, please suggest the wording you'd like to see and I'll apply the change, or make the change directly to the branch. This is currently breaking our integration test and I was hoping to submit yesterday so we could make a release on time next week.

We are working around the fact that either Ruby doesn't enforce compatible runtime dependencies or the FluentD plugin mechanism accidentally breaks that by loading the plugins at runtime. Upgrading to 0.10.0 would be totally fine, if the other plug-in upgrades to it as well. If at some point we must upgrade for any reason, we could make the change to the prometheus plugin as well, it's just a couple lines.

As for the bug, can you talk more about the benefits of filing a bug to track this? I'm concerned that we would have to keep two things up to date: the bug and this comment.

qingling128: Oh, i completely misunderstood why we reverted to `0.9.X`. I thought it was the `0.10.0` gem itself that included some bug and broke us, so I was asking for _that_ bug ticket. If it's just mismatching versions, that's totally fine. I did not realize it was a gem version conflict thing. It all makes sense now. Thanks for the clarification.

lib/fluent/plugin/monitoring.rb
```diff
@@ -13,9 +13,27 @@
 # limitations under the License.
 
 module Monitoring
+  # Base class for the counter.
+  class BaseCounter
+    def increment(_labels, _count)
```
qingling128: Copying the comment from the other PR here in case we want to merge this one on its own:

`_labels, _count` => `_by: 1, _labels: {}`?

jkohen: Changed from mobile while OOO in order to unblock the release. I hope that works.

PR #354, opened 2019-11-03
Update layout for README

dlackty:
While I have been using this plugin for a while, I've been always confused by the first paragraph of README:

Screenshot:

Screenshot 2019-11-04 01 24 13

I suggest combining them into single lines and make it looks like screenshot below:

image


jkohen:
Nice fix, thank you!

PR #356, opened 2019-11-08
Emit the uptime metric

jkohen:
The format of version is `google-fluentd/1.2.3-foo`. It's quite inconvenient in Ruby to extract the Linux distro in a portable manner like we do in the collectd build system. @igorpeshansky in case you know that we already have this, but otherwise feel free to ignore.

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -254,7 +254,7 @@ module Fluent
 
     PLUGIN_NAME = 'Fluentd Google Cloud Logging plugin'.freeze
 
-    # Follows semver format.
+    # Follows semver.org format.
     PLUGIN_VERSION = begin
       # Extract plugin version from file path.
       match_data = __FILE__.match(
@@ -487,7 +487,7 @@ module Fluent
       @retried_entries_count = nil
 
       @ok_code = nil
-      @uptime_time = 0
+      @uptime_time = Time.now.to_i
     end
 
     def configure(conf)
@@ -560,8 +560,10 @@ module Fluent
                     .create(@monitoring_type, @project_id, @resource)
         # Export metrics every 60 seconds.
         timer_execute(:export_metrics, 60) { @registry.export }
+        # Uptime should be a gauge, but the metric definition is a counter and
+        # we can't change it.
         @uptime_metric = @registry.counter(
-          :uptime, [:version], 'Uptime of Monitoring agent')
+          :uptime, [:version], 'Uptime of Logging agent')
         update_uptime
         timer_execute(:update_uptime, 1) { update_uptime }
         @successful_requests_count = @registry.counter(
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -269,12 +269,9 @@ module Fluent
           proc { |spec,| __FILE__.include?(spec.full_gem_path) }) do |spec,|
             spec.version.to_s
           end
-        if matching_version
-          matching_version
-        else
-          # Return a valid but obviously wrong value.
-          '0.0.0-unknown'
-        end
+        # If no matching version was found, return a valid but obviously wrong
+        # value.
+        matching_version || '0.0.0-unknown'
       end
     end.freeze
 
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -254,6 +254,7 @@ module InternalConstants
 
     PLUGIN_NAME = 'Fluentd Google Cloud Logging plugin'.freeze
 
+    # Follows semver format.
```
igorpeshansky: Want to reference semver.org to help explain the abbreviation (or just expand to "semantic versioning")?

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -268,7 +269,12 @@ module InternalConstants
           proc { |spec,| __FILE__.include?(spec.full_gem_path) }) do |spec,|
             spec.version.to_s
           end
-        matching_version
+        if matching_version
+          matching_version
+        else
+          # Return a valid but obviously wrong value.
+          '0.0.0-unknown'
+        end
```
igorpeshansky: This whole block is just `matching_version || '0.0.0-unknown'`.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -481,6 +487,7 @@ def initialize
       @retried_entries_count = nil
 
       @ok_code = nil
+      @uptime_time = 0
```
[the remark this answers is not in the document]
igorpeshansky: @davidbtucker Uptime is the total time the agent's been up. It starts at 0.

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -553,6 +560,10 @@ def configure(conf)
                     .create(@monitoring_type, @project_id, @resource)
         # Export metrics every 60 seconds.
         timer_execute(:export_metrics, 60) { @registry.export }
+        @uptime_metric = @registry.counter(
+          :uptime, [:version], 'Uptime of Monitoring agent')
```
igorpeshansky: s/`Monitoring`/`Logging`/, right?

test/plugin/base_test.rb
```diff
@@ -2046,6 +2046,29 @@ def test_gke_container_logs
...
+      clear_metrics
+      d = create_driver(config)
+      d.run
+      begin
+        # Retry to avoid time races.
+        retries ||= 0
+        now = Time.now.to_i
+        d.instance.update_uptime
+        assert_metric_value.call(
+          :uptime, now, version: Fluent::GoogleCloudOutput.version_string)
```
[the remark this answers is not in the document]
igorpeshansky: Also in [line 796 of `out_google_cloud.rb`](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/356/files#diff-0d6486c76b449a4c9242958738de9cdbR796)...

lib/fluent/plugin/out_google_cloud.rb
jkohen: Thanks. I prefer the current version, if that's OK with you.

lib/fluent/plugin/out_google_cloud.rb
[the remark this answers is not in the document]
jkohen: Yeah, that's a bug in the counter. Both implementations you suggested are valid. Dave's suggestion fits the current implementation better, so I'll use that.

test/plugin/base_test.rb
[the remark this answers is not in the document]
jkohen: That is correct!

lib/fluent/plugin/out_google_cloud.rb
igorpeshansky: Unfortunately, the Ruby community prefers the other version: https://github.com/rubocop-hq/ruby-style-guide#if-as-a-modifier. It's also consistent with, e.g., [line 790](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/356/files#diff-0d6486c76b449a4c9242958738de9cdbR790) below...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -481,6 +487,7 @@ def initialize
       @retried_entries_count = nil
 
       @ok_code = nil
+      @uptime_time = Time.now.to_i
```
igorpeshansky: Unfortunately, this variable name seems misleading (or at least confusing) now, as the name suggests that it stores the time the agent has been up. Something like `@last_uptime_update_time` would, to me, more accurately describe the implementation.

PR #362, opened 2020-01-02
Handle an edge case for timestamp adjustment (December 31).

igorpeshansky:
Fix a bug in timestamp adjustment (we should accept timestamps that are up to one day in the future, regardless of the date).
Fix broken tests that were computing the wrong expected values on December 31.

### DIFF
--- a/test/plugin/base_test.rb
+++ b/test/plugin/base_test.rb
@@ -943,93 +943,206 @@ module BaseTest
     end
   end
 
-  def test_timestamps
+  def test_compute_timestamp
     setup_gce_metadata_stubs
-    current_time = Time.now
+    d = create_driver(APPLICATION_DEFAULT_CONFIG)
+
+    compute_timestamp = lambda do |driver, record, time|
+      driver.instance.send(:compute_timestamp, record, time)
+    end
+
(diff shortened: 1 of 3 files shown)


test/plugin/base_test.rb
```diff
@@ -946,26 +946,42 @@ def test_split_logs_by_tag
...
+    just_under_one_day_later = one_day_later - 1
     next_year = Time.mktime(current_time.year + 1)
     one_second_before_next_year = next_year - 1
-    adjusted_to_last_year =
-      one_second_before_next_year.to_datetime.prev_year.to_time
     one_second_into_next_year = next_year + 1
-    one_day_into_next_year = next_year.to_date.next_day.to_time
+    one_day_into_next_year = next_year.to_datetime.next_day.to_time
+    # Note: these tests behave differently on December 31.
+    # TODO: use patched Time.now instead of messing with the current time.
```
davidbtucker: Suggestion: Instead of patching Time.now, how about making "now" a parameter to compute_timestamp so that it's a pure function, then testing that function directly?

I feel like compute_timestamp is complex enough that it deserves its own unit test.

igorpeshansky: Great idea. Done.

PR #368, opened 2020-01-28
Add a config param grpc_compression_algorithm to support payload compression for gRPC path.

jkohen:
It only works with gRPC channels, which is the default. This needs to be enabled via configuration, because there is a CPU cost.


jkohen:
Thanks, Ling! Can you apply the change to the message when you merge the
PR? I think it's better if someone who is on the team does the merge.

qingling128:
Sure thing. Done.

PR #371, opened 2020-02-18
Make partial success the default behavior and deprecate the flag.

bmoyles0117:


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -964,7 +959,7 @@ def write_request_via_grpc(entries:,
         @log.warn "Dropping #{entries_count} log message(s)",
                   error: error.to_s, error_code: error.code.to_s
 
-      # If partial_success is enabled, valid entries should have be
+      # As partial_success is enabled, valid entries should have be
```
igorpeshansky: s/be/been/? Might as well...

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -406,11 +406,6 @@ module InternalConstants
                  list: [:none, :gzip],
                  :default => nil
 
-    # Whether valid entries should be written even if some other entries fail
-    # due to INVALID_ARGUMENT or PERMISSION_DENIED errors when communicating to
-    # the Stackdriver Logging API. This is highly recommended.
-    config_param :partial_success, :bool, :default => true
```
igorpeshansky: Should we not still accept it but ignore it, for backward compatibility (in case someone turned it on in their config)?

qingling128: +1

bmoyles0117: Done.

lib/fluent/plugin/out_google_cloud.rb
bmoyles0117: Done.

lib/fluent/plugin/out_google_cloud.rb
bmoyles0117: Done.

qingling128: Can we add a TODO: Breaking change. Remove this flag in Logging Agent 2.0.0 release.

igorpeshansky: Also, might want to add `:deprecated` and `:skip_accessor` (see https://docs.fluentd.org/plugin-development/api-config-types).

bmoyles0117: PTAL

lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -408,8 +408,13 @@ module InternalConstants
...
     # due to INVALID_ARGUMENT or PERMISSION_DENIED errors when communicating to
-    # the Stackdriver Logging API. This is highly recommended.
-    config_param :partial_success, :bool, :default => true
+    # the Stackdriver Logging API. This flag is no longer used, and is kept for
+    # backwards compatibility, partial_success is enabled for all requests.
+    # TODO: Breaking change. Remove this flag in Logging Agent 2.0.0 release.
+    config_param :partial_success, :bool,
+                 :default => true,
+                 :skip_accessor => true,
+                 :deprecated => 'This feature is permanently enabled by default'
```
igorpeshansky: Nit: "by default" implies that it could be disabled. I'd just say `'This feature is permanently enabled'`.

bmoyles0117: Done.

PR #376, opened 2020-02-20
Deprecate metadata agent support.

bmoyles0117:


qingling128:
Shall we bump the minor version?

igorpeshansky:
Bumping the minor version is a good idea, because of all of the config semantics changes. Can we do this in a follow-up?

PR #386, opened 2020-03-23
Replace deprecated Engine.msgpack_factory method with new MessagePackFactory.engine_factory method.

ganmacs:
From https://github.com/fluent/fluentd/issues/2871

Fluentd v1.7.4 has Fluent:: MessagePackFactory.engine_factory which is same as the . Fluent::Engine.msgpack_factory
https://github.com/fluent/fluentd/blob/v1.7.4/lib/fluent/msgpack_factory.rb#L38

### DIFF
--- a/lib/fluent/plugin/out_google_cloud.rb
+++ b/lib/fluent/plugin/out_google_cloud.rb
@@ -1909,7 +1909,11 @@ module Fluent
     end
 
     def format(tag, time, record)
-      Fluent::Engine.msgpack_factory.packer.write([tag, time, record]).to_s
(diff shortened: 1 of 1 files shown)


ganmacs:
@googlebot I signed it!

qingling128:
Looks good.

CPU
image

**Memory**
image

**Throughputs**
image

qingling128:
Move the original title to a note here:

`Fluent::Engine.msgpack_factory` is deprecated in fluentd v1.8.0.

ganmacs:
Thanks!

PR #392, opened 2020-04-10
Upgrade grpc version

sophieyfang:
We are upgrading to Ruby 2.7.0 in google-fluentd which requires grpc to be upgraded first.


sophieyfang:
Please see the tested information in our internal bug ticket.

qingling128:
There seems to be a memory regression (see details in the internal ticket). But it might not be related to this upgrade. Needs more investigation.

igorpeshansky:
This is stale.

PR #396, opened 2020-05-15
Factor out a utils module for managing resources.

davidbtucker:
This will allow other plugins to reuse these functions, which are
required for the monitoring module.


lib/fluent/plugin/out_google_cloud.rb
```diff
@@ -29,6 +29,7 @@
 require 'googleauth'
 
 require_relative 'monitoring'
+require_relative 'resource'
```
qingling128, line 32: `resource` is a common word in this plugin and it typically refers to the Monitored Resource. It can be a bit misleading for this file to be called `resource`. If it is just common utils, maybe we can just call it `common`? Yet it looks like some of the contents that are being moved are indeed exclusinve to the out_google_cloud plugin. So `common` may not be a good name.

BTW, I guess this PR is somehow related to the test failures in https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/394. Is that the reason why we are moving this?

davidbtucker: Done (renamed resource.rb -> common.rb).

The motivation is to factor out a library for creating a monitored resource. We need a monitored resource in order to create a monitory registry (see https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/56eb4b84a33f4d976d2999c78101c75c186c2dd5/lib/fluent/plugin/monitoring.rb#L153).

Thanks for pointing out the test failures on #394, I hadn't noticed (the tests all pass in my client). I'll take a look.

davidbtucker:
FYI, I rebased and added a missing file.

PR #399, opened 2020-05-27
Add OpenCensus support for config analysis metrics.

davidbtucker:


lib/fluent/plugin/filter_analyze_config.rb
```diff
@@ -122,16 +123,55 @@ module Constants
...
+    #   - 'prometheus', in this case default registry in the Prometheus
+    #     client library is used, without actually exposing the endpoint
+    #     to serve metrics in the Prometheus format.
+    #    - any other value will result in the absence of metrics.
+    config_param :monitoring_type, :string,
+                 :default => Monitoring::PrometheusMonitoringRegistry.name
+
+    # Override for the Google Cloud Monitoring service hostname, or
+    # `nil` to leave as the default.
+    config_param :gcm_service_address, :string, :default => nil
```
qingling128, line 172: Are these params supposed to mock the monitored resource related params in the `google_cloud` output plugin?

I wonder if we can directly take the values of these params from what users set in the `google_cloud` output plugin. That way the usage of this filter plugin remains simple even if users overwrite monitored resource. And users do not need to worry about keeping the two in sync. Since we are parsing the output plugin configuration already, we sort of have access to those params already.

What do you think?

qingling128: Discussed offline. We can just blow away these params from this plugin and take the default value.

davidbtucker: Thanks, pushed a new snapshot, PTAL.

PR #400, opened 2020-05-27
Switch to using the gke_container for legacy GKE resources.

igorpeshansky:
This enables writing metrics against that resource as well.

The Logging API would accept either `container` or `gke_container` as aliases.


qingling128:
LGTM

PR #403, opened 2020-06-08
Rename auto-approve-yml to auto-approve.yml

igorpeshansky:
A clean-up to give the file the correct extension.


igorpeshansky:
The workflow seems to have been triggered after this PR was sent, so the change itself may not be needed at all. Let's hold off on submitting until the next automated PR — if that one doesn't work, we can try submitting this one.

jkschulz:
It should still be renamed so that it has the correct extension.

igorpeshansky:
Thanks for confirming.

PR #406, opened 2020-06-16
Revert "Remove waits on test start-up. (#343)".

qingling128:
Reverts #343 to restore test waits to avoid broken tests being shown as passed.

### DIFF
--- a/test/plugin/test_driver.rb
+++ b/test/plugin/test_driver.rb
@@ -18,19 +18,6 @@ require 'fluent/test/input_test'
 
 module Fluent
   module Test
-    # rubocop:disable Style/ClassVars
-    class BufferedOutputTestDriver < InputTestDriver
-      @@run_method = BufferedOutputTestDriver.instance_method(:run)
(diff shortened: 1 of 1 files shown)


qingling128:
Ah, I'm still trying different things to see if there is a way to avoid skipping test while keeping the test run time lower. Will ping when it's ready for a review.

qingling128:
The https://groups.google.com/g/fluentd/c/ze7YY1ue40s/discussion thread mentioned that the wait is still needed if flush_interval is used. Our output plugin is a buffered output plugin that uses flush interval. So I guess the wait is still necessary. The implementation forces every unit test to wait for at least 0.05 seconds. I tried decreasing the interval, but it did not help with the test run time.

Since we have seen reports that broken tests get "skipped" and are shown as passed, I think it's safer to revert this for now.

qingling128:
It's ready for review now.

qingling128:
Since the latest code is identical with the previously approved version. I'm gonna merge it.

PR #430, opened 2020-09-16
Make metrics prefix parameterized in the opencensus path.

qingling128:
Per discussion offline, we want to send the metrics under `internal/logging/config` instead.

Tested via local Prometheus endpoint in b/153356846. Also tested metrics prefix customization by changing the default prefix temporarily and verify the ingestion of `request_count` metrics.

### DIFF
--- a/lib/fluent/plugin/monitoring.rb
+++ b/lib/fluent/plugin/monitoring.rb
@@ -52,7 +52,7 @@ module Monitoring
     def initialize(_project_id, _monitored_resource, _gcm_service_address)
     end
 
-    def counter(_name, _labels, _docstring)
+    def counter(_name, _labels, _docstring, _prefix = None)
       BaseCounter.new
     end
 
@@ -75,7 +75,7 @@ module Monitoring
     end
 
     # Exception-driven behavior to avoid synchronization errors.
-    def counter(name, _labels, docstring)
+    def counter(name, _labels, docstring, _prefix = None)
       # When we upgrade to Prometheus client 0.10.0 or higher, pass the
       # labels in the metric constructor. The 'labels' field in
       # Prometheus client 0.9.0 has a different function and will not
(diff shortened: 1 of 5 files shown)


lib/fluent/plugin/monitoring.rb
```diff
@@ -97,21 +99,27 @@ def initialize(project_id, monitored_resource, gcm_service_address)
...
-        project_id: project_id,
-        metric_prefix: 'agent.googleapis.com/agent',
-        resource_type: monitored_resource.type,
-        resource_labels: monitored_resource.labels,
-        gcm_service_address: gcm_service_address
-      )
+      @project_id = project_id
+      @monitored_resource = monitored_resource
+      @gcm_service_address = gcm_service_address
+      @recorders = { DEFAULT_PREFIX => OpenCensus::Stats.ensure_recorder }
```
davidbtucker, line 105: Can we just make @recorders and @exporters empty here?

davidbtucker (changes requested):

lib/fluent/plugin/monitoring.rb
qingling128: Good point. Removed.

lib/fluent/plugin/monitoring.rb
```diff
@@ -97,21 +99,14 @@ def initialize(project_id, monitored_resource, gcm_service_address)
...
       @log = $log # rubocop:disable Style/GlobalVars
-      @recorder = OpenCensus::Stats.ensure_recorder
-      @exporter = OpenCensus::Stats::Exporters::Stackdriver.new(
-        project_id: project_id,
-        metric_prefix: 'agent.googleapis.com/agent',
-        resource_type: monitored_resource.type,
-        resource_labels: monitored_resource.labels,
-        gcm_service_address: gcm_service_address
-      )
-      OpenCensus.configure do |c|
```
qingling128, line 108: Tested that opencensus ingestion still works:

image

qingling128 (commented):
This is ready for review.

davidbtucker (approved):

PR #437, opened 2020-11-06
Add has_ruby_snippet label to the enabled_plugins counter registry.

qingling128:


lib/fluent/plugin/filter_analyze_config.rb
```diff
@@ -245,7 +245,8 @@ def configure(conf)
 
         plugin_usage = registry.counter(
           :enabled_plugins,
-          [:plugin_name, :is_default_plugin, :has_default_config],
+          [:plugin_name, :is_default_plugin,
```
davidbtucker, line 248: Oops, not sure why I didn't include this label. Thanks for the fix.

davidbtucker (approved):

PR #440, opened 2020-11-11
Fix monitoring metrics ingestion via open census.

qingling128:
* Add the accidentally removed resource_type back.
* Send the config analysis metrics as GAUGE instead of CUMULATIVE.
* Schedule the config analysis metrics to be exported.

### DIFF
--- a/lib/fluent/plugin/filter_analyze_config.rb
+++ b/lib/fluent/plugin/filter_analyze_config.rb
@@ -15,6 +15,7 @@
 require 'fileutils'
 require 'fluent/config'
 require 'fluent/config/v1_parser'
+require 'fluent/plugin_helper'
 require 'googleauth'
 require 'google/apis/logging_v2'
 require 'open-uri'
@@ -31,6 +32,10 @@ module Fluent
   class AnalyzeConfigFilter < Filter
     include Fluent::Config
     Fluent::Plugin.register_filter('analyze_config', self)
+
+    # Required for the timer_execute method below
+    include PluginHelper::Mixin
+    helpers :timer
 
     module Constants
       # Built-in plugins that are ok to reference in metrics.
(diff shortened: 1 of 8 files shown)


lib/fluent/plugin/filter_analyze_config.rb
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 require 'fileutils'
+require 'fluent/plugin_helper'
```
davidbtucker, line 16: Nit: Alphabetize?

davidbtucker (approved):

lib/fluent/plugin/filter_analyze_config.rb
qingling128: Fixed

test/plugin/base_test.rb
```diff
@@ -1984,54 +1994,88 @@ def test_metrics
...
             end
             # rubocop:enable Lint/HandleExceptions
+            failed_requests_count, dropped_entries_count,
+            retried_entries_count = metric_values
+
+            successful_requests_count = \
+              if code != ok_status_code
+                0
+              elsif config == ENABLE_OPENCENSUS_CONFIG
+                # The test driver instance variables can not survive between
```
qingling128, line 2028: This is not ideal. There is some limitation of the test driver: e.g. it does not support a `flush` method to mock the actual behavior to bundle entries into requests as I browsed through their doc and code base.

Any suggestion to improve this is welcome. I guess potentially we can write a custom test driver to handle something like this. But I did not go all the way to do that as the additional benefit seems minimal (aka the prometheus path already covers the full story). Let me know if that is concerning.

qingling128:
Rebased from master, added a TODO and fixed some unit test failures.

davidbtucker (approved):

qingling128:
Tested:

image
image

PR #447, opened 2021-02-03
chore: update to fluentd v1.12.3

Cryptophobia:
- increment gem version for plugin
  - update to allow fluentd v1.12.3 gem
  - update json to 2.4.1 for [CVE-2020-10663](https://www.ruby-lang.org/en/news/2020/03/19/json-dos-cve-2020-10663/)

Signed-off-by: Anton Ouzounov <aouzounov@vmware.com>

Hello maintainers! We are using fluentd-plugin-google-cloud as a plugin in the [kube-fluentd-operator](https://github.com/vmware/kube-fluentd-operator) . We want to be able to migrate to using fluentd v1.12.0 for the benefit of our users, this is the only plugin currently forcing us to use fluentd v1.11.2 . Can we help upgrade? I don't see any breaking changes in the [fluentd changelogs](https://github.com/fluent/fluentd/blob/master/CHANGELOG.md) that will present problems.

### DIFF
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -10,7 +10,7 @@ eos
   gem.homepage      =
     'https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud'
   gem.license       = 'Apache-2.0'
-  gem.version       = '0.10.9'
+  gem.version       = '0.11.0'
   gem.authors       = ['Stackdriver Agents Team']
   gem.email         = ['stackdriver-agents@google.com']
   gem.required_ruby_version = Gem::Requirement.new('>= 2.2')
@@ -19,14 +19,14 @@ eos
   gem.test_files    = gem.files.grep(/^(test)/)
   gem.require_paths = ['lib']
 
-  gem.add_runtime_dependency 'fluentd', '1.11.2'
+  gem.add_runtime_dependency 'fluentd', '1.12.3'
   gem.add_runtime_dependency 'googleapis-common-protos', '1.3.10'
   gem.add_runtime_dependency 'googleauth', '0.9.0'
   gem.add_runtime_dependency 'google-api-client', '0.30.8'
   gem.add_runtime_dependency 'google-cloud-logging', '1.6.6'
   gem.add_runtime_dependency 'google-protobuf', '3.15.8'
   gem.add_runtime_dependency 'grpc', '1.31.1'
-  gem.add_runtime_dependency 'json', '2.2.0'
+  gem.add_runtime_dependency 'json', '2.4.1'
   gem.add_runtime_dependency 'opencensus', '0.5.0'
   gem.add_runtime_dependency 'opencensus-stackdriver', '0.3.2'
 


Cryptophobia:
@googlebot I signed it!

Cryptophobia:
@googlebot I signed it!

Cryptophobia:
@davidbtucker, do you think this is a safe update or does it need more testing?

Cryptophobia:
@qingling128 @ekund , can we relax the runtime dependency fluentd fixed at a particular fluentd version?

qingling128:
Hi @Cryptophobia - The reason we pin dependency versions is to achieve reproducible builds. Any specific features / bugs you are looking for in Fluentd 1.12? What's the impact scope?

Cryptophobia:
@qingling128 , that is fine and you can version fluentd separately in your build system or when you build your containers. There is no reason why your gem, which is a dependency of fluentd needs to specify a required version for it.

For example, your gem is the only one that does not allow us to upgrade past fluentd v1.11.2 here: https://github.com/vmware/kube-fluentd-operator/blob/master/base-image/Gemfile . We have so many plugins, and this is the only one that constrains us from upgrading to v0.12.0 for absolutely no reason.

All of the other plugin gems have very relaxed constraints such as this one for example ((fluent-plugin-systemd)[https://github.com/fluent-plugin-systemd/fluent-plugin-systemd]): `spec.add_runtime_dependency 'fluentd', ['>= 0.14.11', '< 2']`

qingling128:
@Cryptophobia - Hi, thanks for the suggestion. That is another possibility (aka refactor the build infra to move the version pinning away from gem spec). We'll need to evaluate its impact and get it prioritized. Given the current workload on the team's plate, it might not happen in the short term.

We could consider upgrading Fluentd version though. Note that we have internal integration and load tests that need to pass for each release. Fluentd version turned out to be one of main causes when CPU / memory regression happened in the past. So we need to go through a set of validation process to upgrade the Fluentd version.

Cryptophobia:
@qingling128 , okay, we would appreciate it very much! Hopefully, there is not many breaking changes from `fluentd v1.11.x` to `fluentd v1.12.0`. When I looked at the changelogs, I did not notice anything immediately obvious that may be a problem.

https://github.com/fluent/fluentd/blob/master/CHANGELOG.md

cosmo0920:
We're investigating some breakages on introducing `follow_inodes` feature on `in_tail`:
https://github.com/fluent/fluentd/issues/3239

Any feedback is appreciated.

Cryptophobia:
@cosmo0920 , are you hitting that particular issue on `1.12.0` with this particular plugin, have you tried downgrading to `1.11.x`?

Alternative to see if `1.12.2` will fix the issue.

cosmo0920:
> @cosmo0920 , are you hitting that particular issue on `1.12.0` with this particular plugin, have you tried downgrading to `1.11.x`?

Yeah, downgrading to v1.11.x vanishes this issue.

> Alternative to see if 1.12.2 will fix the issue.

We'd like to go Fluentd v1.12 world. Fluentd v1.12.2 will fix almost `in_tail` issues.
But we're still investigating whether unknown issues still exists or not on users environments.

qingling128:
Sounds like `v1.12.2` is the version we want? Can someone confirm it so that we can start the internal validation process on it?

Cryptophobia:
https://www.fluentd.org/blog/fluentd-v1.12.2-has-been-released

`1.12.2` has all of the `in_tail` fixes but there is still an issue open that @cosmo0920 linked to above:

I'm going to include this gem built from my fork using fluentd v1.12.2 for our KFO project.

Cryptophobia:
I had to change the dependency version pinning to this:

```
  gem.add_runtime_dependency 'fluentd', '~> 1.12', '>= 1.12.2', '< 1.13'
```

Cryptophobia:
Also added upgrade of json gem for [CVE-2020-10663](https://www.ruby-lang.org/en/news/2020/03/19/json-dos-cve-2020-10663/) to this PR:

Cryptophobia:
Looks like fluentd maintainers will be releasing `v1.12.3` with bug fixes for some remaining `in_tail` issue. https://github.com/fluent/fluentd/issues/3323#issuecomment-819134977

I will update this PR once that version is released.

Cryptophobia:
fluentd v1.12.3 has been released! can we get this merged in now?

fluent-plugin-google-cloud.gemspec
```diff
@@ -19,14 +19,14 @@ eos
   gem.test_files    = gem.files.grep(/^(test)/)
   gem.require_paths = ['lib']
 
-  gem.add_runtime_dependency 'fluentd', '1.11.2'
+  gem.add_runtime_dependency 'fluentd', '~> 1.12', '>= 1.12.2', '< 1.13'
```
qingling128, line 22: Let me kick off an internal validation process for Fluentd `1.12.3`. As mentioned in https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/447#issuecomment-810655681, the version has to pinned to an exact version for now, e.g.

`gem.add_runtime_dependency 'fluentd', '1.12.3'`

Please update.

fluent-plugin-google-cloud.gemspec
```diff
@@ -19,14 +19,14 @@ eos
...
   gem.add_runtime_dependency 'googleapis-common-protos', '1.3.10'
   gem.add_runtime_dependency 'googleauth', '0.9.0'
   gem.add_runtime_dependency 'google-api-client', '0.30.8'
   gem.add_runtime_dependency 'google-cloud-logging', '1.6.6'
   gem.add_runtime_dependency 'google-protobuf', '3.15.8'
   gem.add_runtime_dependency 'grpc', '1.31.1'
-  gem.add_runtime_dependency 'json', '2.2.0'
+  gem.add_runtime_dependency 'json', '~> 2.3.0'
```
qingling128, line 29: Which exact version of `json` are we looking for here?

qingling128 (changes requested):

fluent-plugin-google-cloud.gemspec
Cryptophobia: We are looking for any version above 2.3.0 which is not vulnerable. We can try `2.4.1` as that is recommended in the fluentd provided docker images?

Cryptophobia: Information here: https://github.com/fluent/fluentd-docker-image/blob/master/v1.12/debian/Dockerfile

fluent-plugin-google-cloud.gemspec
Cryptophobia: Okay, updated the PR to force exact version `1.12.3`. Ideally, can we move away from forcing fluentd to particular version. Otherwise, anytime we have to upgrade [KFO](https://github.com/vmware/kube-fluentd-operator) we will always have to fork this gem.

Cryptophobia:
Updated @qingling128 .

qingling128:
> Updated @qingling128 .

Thanks! Our internal validation process takes a few days to soak to catch any potential memory leaks etc. We should have some result on Monday.

qingling128:
Initial numbers look good.

CPU
image

RSS
image

Log entry ingestion rate
image

Merging this PR so that the internal 4-day validation process can pick it up tonight.

qingling128 (commented):
LGTM

qingling128 (approved):

Cryptophobia:
Thank you @qingling128 ! Numbers indeed look stable. Will make other PRs in the future when we need to update fluentd version!

qingling128:
The release was delayed for a bit because we saw some resource usage regression (turned out to be false alarm). The May 12 release candidate is going through the 4-day validation. If everything passes, the release will happen early next week.

qingling128:
Actually, it turned out there is still a memory regression, and related to this PR. Currently investigating the issue and try to figure out whether it's due to `fluentd` or `json` gem version update.

qingling128:
The issue was caused by the Fluentd version bump:

The memory and CPU usage of Fluentd `1.12.3` is higher than `1.11.2`, resulting in the logging agent not being able to catch up with 3000 log entries / second throughput.

image
image
image

Cryptophobia:
@qingling128 , are we going to report this to https://github.com/fluent/fluentd/ repo?

Are the other basegem dependencies versions equivalent to these: https://github.com/vmware/kube-fluentd-operator/blob/master/base-image/basegems/Gemfile ? It is possible that it could be a ruby version issue or another base gem that needs to be upgraded for better memory optimization?

qingling128:
Filed an upstream bug at https://github.com/fluent/fluentd/issues/3389. The ruby version, dep gems are the same. See details in the upstream bug.

Cryptophobia:
Looks like fluentd v1.13.0 with a few in_tail fixes will release soon. https://github.com/fluent/fluentd/issues/3333#issuecomment-847450586 . Maybe they will get to looking at the memory/cpu issues. We should try to update to v1.13.0 when it goes out and test again.

cosmo0920:
> Looks like fluentd v1.13.0 with a few in_tail fixes will release soon. [fluent/fluentd#3333 (comment)](https://github.com/fluent/fluentd/issues/3333#issuecomment-847450586) . Maybe they will get to looking at the memory/cpu issues.

We're also planning to fix in_tail fixes on Fluentd v1.12.4.

PR #452, opened 2021-02-22
Upgrade google-protobuf to 3.15.0

hsmatulisgoogle:


igorpeshansky (approved):
LGTM :shipit:

Be careful when merging — a PR with a single commit usually gets the description of that commit, not of the PR, so you'd need to explicitly copy the PR one.

igorpeshansky (changes requested):
Actually, looks like the tests need to be adjusted first, because the serialization of duration values has changed…

igorpeshansky (approved):
LGTM :shipit:

PR #458, opened 2021-04-28
Automated: Bump gem version to 0.10.10 for future releases.

stackdriver-instrumentation-release:


qingling128:
This is not needed because the version is already bumped by https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/commit/595f5555cef7a8117722ae77e297390d9d10579b. A fix for the release automation is tracked by b/186685931.

PR #474, opened 2021-09-03
Switch to GitHub Actions.

igorpeshansky:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+          - 2.6
+          - 2.5
+          - 2.4
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        env:
+          BUNDLE_GEMFILE: ${{ github.workspace }}/Gemfile
+        with:
```
qingling128, line 18: Does this automatically handle `gem update bundler` already? If so, feel free to submit. The rest LGTM.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
...
+          ruby-version: ${{ matrix.rvm }}
+          bundler-cache: true
+          # Run bundle install outside the workspace to keep gems out of the
+          # repo checkout.
+          working-directory: ${{ runner.temp }}
+      - run: |
+          bundle exec rake
+      # TODO: notifications
```
qingling128, line 26:
```suggestion
      # TODO: send email notifications
```

qingling128 (approved):
LGTM with a question and a suggested edit.

.github/workflows/ci.yml
igorpeshansky: It does not update the `bundler` gem explicitly, but I've found that this wasn't needed — the `ruby/setup-ruby` task already installs the latest `bundler`. Using `bundle-cache: true` causes it to run `bundle install` automatically.

.github/workflows/ci.yml
igorpeshansky: I've actually embedded the whole Travis notification section in the TODO instead — that way we know exactly what needs to be replicated…

.github/workflows/ci.yml
qingling128: Got it.

.github/workflows/ci.yml
qingling128: Sounds great

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+          BUNDLE_GEMFILE: ${{ github.workspace }}/Gemfile
+        with:
+          ruby-version: ${{ matrix.rvm }}
+          bundler-cache: true
+          # Run bundle install outside the workspace to keep gems out of the
+          # repo checkout.
+          working-directory: ${{ runner.temp }}
+      - run: |
+          bundle exec rake
+      # TODO: notifications:
```
qingling128, line 26:
```suggestion
      # TODO: notifications similar to the travis settings below:
```

qingling128: Clarified that the commented out section are not straight forward GitHub action settings

qingling128 (approved):
LGTM with a suggestion.

.github/workflows/ci.yml
igorpeshansky: The `notifications:` header was actually part of the settings to migrate. Applied in a slightly rephrased way.

PR #465, opened 2021-06-23
Upgrade google-protobuf to v3.17.3

stanhu:
google-protobuf v3.15.x and v3.16.x can seg fault if repeated fields in
a Hash are garbage collected:
https://github.com/protocolbuffers/protobuf/issues/8559

PR #466, opened 2021-07-27
Revert json gem version from 2.4.1 to 2.2.0 for memory regression.

qingling128:
b/188285355

PR #475, opened 2021-09-13
chore(gemspec): update fluentd >= 1.14.x

Cryptophobia:
- update fluentd to 1.14.0 gem
- update json to 2.4.1 for [CVE-2020-10663](https://www.ruby-lang.org/en/news/2020/03/19/json-dos-cve-2020-10663/)
- upgrade google-protobuf gem to 3.17.3  (same as PR https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/465) 

Fluentd changelog for [v1.14.0](https://github.com/fluent/fluentd/blob/master/CHANGELOG.md#v1140)

Signed-off-by: Anton Ouzounov <aouzounov@vmware.com>


qingling128:
We'd need to carefully measure the performance of these gem version upgrades.

The [last attempt](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/447) to upgrade json gem caused a memory regression and we had to [roll it back](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/466).

We do know that json 2.4.1 is not safe to upgrade to. They do have a newer version 2.5.1 out though.

Cryptophobia:
Hi @qingling128 , what version of ruby are you using in your testing environment? Also what version of jemalloc are you using? There are multiple issues with using later than jemalloc v3.6.0 with libmalloc and ruby will cause memory to continue on fluentd. This may not be due to actual json gem upgrade but your environment, configuration.

https://github.com/fluent-plugins-nursery/td-agent-builder/issues/305

Cryptophobia:
I pushed json 2.5.1 here for testing.

Cryptophobia:
This has been rebased with latest from upstream/master and works with ruby 2.7.4 now. Please test. Test seem to fail with ruby 2.4 now.

PR #483, opened 2022-02-09
chore: bump grpc gem to 1.43.1

perek:
Due to a bug in grpc gem you cannot compile this plugin after 10.0 on ARM architecture https://github.com/grpc/grpc/issues/23889

We need to bump this dep up to support native arm extensions

### DIFF
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -28,7 +28,7 @@ eos
   gem.add_runtime_dependency 'google-api-client', '0.30.8'
   gem.add_runtime_dependency 'google-cloud-logging', '1.6.6'
   gem.add_runtime_dependency 'google-protobuf', '3.15.8'
-  gem.add_runtime_dependency 'grpc', '1.31.1'
+  gem.add_runtime_dependency 'grpc', '1.43.1'
   gem.add_runtime_dependency 'json', '2.2.0'
   gem.add_runtime_dependency 'opencensus', '0.5.0'
   gem.add_runtime_dependency 'opencensus-stackdriver', '0.4.1'
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -27,7 +27,7 @@ eos
   gem.add_runtime_dependency 'googleauth', '0.9.0'
   gem.add_runtime_dependency 'google-api-client', '0.30.8'
   gem.add_runtime_dependency 'google-cloud-logging', '1.6.6'
-  gem.add_runtime_dependency 'google-protobuf', '3.15.8'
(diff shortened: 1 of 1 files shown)


fluent-plugin-google-cloud.gemspec
```diff
@@ -27,8 +27,8 @@ eos
   gem.add_runtime_dependency 'googleauth', '0.9.0'
   gem.add_runtime_dependency 'google-api-client', '0.30.8'
   gem.add_runtime_dependency 'google-cloud-logging', '1.6.6'
-  gem.add_runtime_dependency 'google-protobuf', '3.15.8'
-  gem.add_runtime_dependency 'grpc', '1.31.1'
+  gem.add_runtime_dependency 'google-protobuf', '3.18.2'
```
perek, line 30: needed to bump this as well

qingling128 (approved):
Per evaluation in http://b/219963999, these versions look good. LGTM

qingling128:
Looks like this also requires upgrading to >= Ruby 2.5.0

qingling128:
In fact, I just checked the package. We are already using Ruby `2.6.9` there: https://github.com/GoogleCloudPlatform/google-fluentd/blob/250f1343be7c90f35101c4dae708090d257e6ae6/config/software/ruby.rb#L31. So it should be fine to just remove `2.4` from the testing grid.

qingling128:
Sent two PRs:

* https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/485
* https://github.com/GoogleCloudPlatform/google-fluentd/pull/374

qingling128:
@perek - Some updates: We have a custom build of `google-fluentd` with this new gem version now. However, the new gRPC version breaks support on a version of Amazon AMI we support (based on CentOS 6), so we might not be able to release a formal version before we can claim end of support for that.

Are you interested in trying out a custom build instead?

perek:
i am building fluentd from scratch, so I think just a new release would work for me too so I can swap the ruby gems

igorpeshansky:
See also https://github.com/GoogleCloudPlatform/google-fluentd/pull/376

qingling128:
@perek - We might not be able to do a release soon due to the breakage of CentOS 6. If you are building from scratch, would a custom gem push work for you? something like `0.12.5.rc.1`?

igorpeshansky:
I'm going to try to get our Amazon AMI story under control, so we can actually release.

qingling128:
Here we go: https://rubygems.org/gems/fluent-plugin-google-cloud/versions/0.12.5.rc.1

jefferbrecht:
FYI, we're seeing about 15-20 MB increased memory usage overall after upgrading:

image

Purportedly there are some memory fixes in 1.45.0, so I will open a PR to upgrade to it. Not sure if it will make a difference here but it would be good to pick it up anyway on its own merits.

PR #500, opened 2023-03-17
Bump rubocop version to 1.48.1.

igorpeshansky:
Also fix all of the new warnings.

[b/272577695](http://b/272577695).

See also https://github.com/GoogleCloudPlatform/fluent-plugin-detect-exceptions/pull/101.

### DIFF
--- a/fluent-plugin-google-cloud.gemspec
+++ b/fluent-plugin-google-cloud.gemspec
@@ -13,7 +13,7 @@ eos
   gem.version       = '0.12.10'
   gem.authors       = ['Stackdriver Agents Team']
   gem.email         = ['stackdriver-agents@google.com']
-  gem.required_ruby_version = Gem::Requirement.new('>= 2.2')
+  gem.required_ruby_version = Gem::Requirement.new('>= 2.3')
 
   gem.files         = Dir['**/*'].keep_if { |file| File.file?(file) }
   gem.test_files    = gem.files.grep(/^(test)/)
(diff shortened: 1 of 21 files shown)


fluent-plugin-google-cloud.gemspec
```diff
@@ -1,35 +1,35 @@
...
   # https://github.com/GoogleCloudPlatform/google-fluentd/blob/master/config/software/fluentd.rb.
   gem.add_runtime_dependency 'fluentd', '1.13.3'
+  gem.add_runtime_dependency 'google-api-client', '0.30.8'
   gem.add_runtime_dependency 'googleapis-common-protos', '1.3.10'
   gem.add_runtime_dependency 'googleauth', '0.9.0'
-  gem.add_runtime_dependency 'google-api-client', '0.30.8'
   gem.add_runtime_dependency 'google-cloud-logging', '1.6.6'
   gem.add_runtime_dependency 'google-protobuf', '3.20.0'
   gem.add_runtime_dependency 'grpc', '1.45.0'
-  gem.add_runtime_dependency 'json', '2.2.0'
```
LujieDuan, line 32: Just want to check: is this intended upgrade?

LujieDuan (approved):
LGTM! 

Thanks for having each fix in its own commit - that makes verifying each enforcement very easy. 

The presubmit is failed as `rubocop=1.48.1` requires ruby >= 2.6. Should we modify the test matrix in this PR?

fluent-plugin-google-cloud.gemspec
igorpeshansky: Yes, I believe this was required by the newer version of Rubocop…

LujieDuan (approved):
LGTM!

